PR #156, opened 2016-08-03
Authorized Network Management for Cloud SQL

iantalarico:
Add support to manage all the authorized networks for a Cloud SQL instance.


GoogleCloudExtension/GoogleCloudExtension.DataSources/CloudSQLDataSource.cs
```diff
@@ -67,5 +63,60 @@ public Task<IList<DatabaseInstance>> GetInstanceListAsync()
...
+        /// </summary>
+        /// <returns>The Cloud SQL instance.</returns>
+        public Task<DatabaseInstance> GetInstanceAsync(string name)
+        {
+            try
+            {
+                InstancesResource.GetRequest request = Service.Instances.Get(ProjectId, name);
+                return request.ExecuteAsync();
+            }
+            catch (GoogleApiException ex)
```
ivannaranjo: You are not going to get the exception because you are not awaiting the result of `ExecuteAsync`. Change the method to be `async` and await the result, that way you will be able to catch the exception.

GoogleCloudExtension/GoogleCloudExtension.DataSources/DatabaseInstanceExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+        public const string UnknownState = "UNKNOWN_STATE";
+
+        /// <summary>
+        /// Update the authorized networks of a instance.  This completely replaces the current 
+        /// list of networks with the new list.
+        /// </summary>
+        public static void UpdateAuthorizedNetworks(DatabaseInstance instance, IList<AclEntry> acls)
+        {
+            EnsureAuthorizedNetworksInitialized(instance);
+            instance.Settings.IpConfiguration.AuthorizedNetworks = acls;
```
ivannaranjo: Shouldn't this method also do the call to the service to update the instance?

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,132 @@
...
+        public static readonly TimeSpan DefaultPollInterval = TimeSpan.FromSeconds(1);
+
+        /// <summary>
+        /// The default polling timeout.  This value is the total time spent in delays before a timeout 
+        /// is considerd to have occured.  This does not take into account any other time spent.
+        /// Default is 5 minutes.
+        /// </summary>
+        public static readonly TimeSpan DefaultPollTimeout = TimeSpan.FromMinutes(5);
+
+
```
ivannaranjo: Extra white line.

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// Poll for a resource.
+        /// </summary>
+        /// <param name="task">The origional task to start from.</param>
+        /// <param name="fetch">A delegate to fetch an updated resource.</param>
+        /// <param name="stopPolling">A delegate to determine if polling should stop.</param>
+        /// <param name="config">Optional, A polling configuration or the default if unset or null.</param>
+        /// <param name="token">Optional, A cancelation token used to stop polling manually.</param>
+        /// <exception cref="TimeoutException">If the polling passes the timeout threshold.</exception>>
+        /// <returns></returns>
+        public static async Task<T> Poll(Task<T> task, Fetch fetch, StopPolling stopPolling,
```
ivannaranjo: I'm not sure what you are doing here with the polling of a task, is this for operations?

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,132 @@
...
+            PollingConfiguration config = null, CancellationToken? token = null)
+        {
+            // If no configuration is given use the default configuration.
+            config = config ?? new PollingConfiguration();
+
+            // Wait for the initial task to complete before polling.
+            T result = await task;
```
ivannaranjo: This can wait indefinitely, you should probably await the `task` and a task using your default timeout, otherwise this can hung.

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// </summary>
+        /// <param name="task">The origional task to start from.</param>
+        /// <param name="fetch">A delegate to fetch an updated resource.</param>
+        /// <param name="stopPolling">A delegate to determine if polling should stop.</param>
+        /// <param name="config">Optional, A polling configuration or the default if unset or null.</param>
+        /// <param name="token">Optional, A cancelation token used to stop polling manually.</param>
+        /// <exception cref="TimeoutException">If the polling passes the timeout threshold.</exception>>
+        /// <returns></returns>
+        public static async Task<T> Poll(Task<T> task, Fetch fetch, StopPolling stopPolling,
+            PollingConfiguration config = null, CancellationToken? token = null)
```
ivannaranjo: You should use the default value for a cancellation token, not a nullable one, that is the standard practice.

GoogleCloudExtension/GoogleCloudExtension.DataSources/DatabaseInstanceExtensions.cs
iantalarico: I prefer to keep the instance manipulation and the updating separate.  The instance extensions will just handle the object manipulations and the data source will do the updates.

GoogleCloudExtension/GoogleCloudExtension.DataSources/CloudSQLDataSource.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done

iantalarico:
Thanks Ivan, PTAL

GoogleCloudExtension/GoogleCloudExtension.DataSources/DatabaseInstanceExtensions.cs
ivannaranjo: But you could have the extension call into the data source. It is just that the method name, `UpdateAuthorizedNetworks` seem to me that they perform the update.

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,153 @@
...
+                    throw new TimeoutException();
+                }
+
+                // Wait and then poll.
+                await Task.Delay(config.Interval);
+                elapsed += config.Interval;
+
+                // Fetch the next resource.
+                Task<T> task = fetch.Invoke(resource);
+                if (await Task.WhenAny(task, Task.Delay(config.FetchTimeout)) == task)
```
ivannaranjo: You should cap the `Task.Delay` here by the by the remaining time out, `config.Timeout - elapsed`. Otherwise you could wait past the `config.Timeout` as you are awaiting the task for a full `config.FetchTimeout` period.

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// <param name="item">The current resource.</param>
+        /// <returns>The updated resource.</returns>
+        public delegate Task<T> Fetch(T item);
+
+        /// <summary>
+        /// Delegate to determine if polling should stop.
+        /// </summary>
+        /// <param name="item">The current resource.</param>
+        /// <returns>True if the polling should stop.</returns>
+        public delegate bool StopPolling(T item);
```
ivannaranjo: Sorry I didn't see this before, this is just a `Predicate<T>`, I would suggest to use the built in type instead of defining your own version.

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,153 @@
...
+    /// </summary>
+    public class Polling<T>
+    {
+
+        /// <summary>
+        /// Delegate to fetch an updated resource.
+        /// </summary>
+        /// <param name="item">The current resource.</param>
+        /// <returns>The updated resource.</returns>
+        public delegate Task<T> Fetch(T item);
```
ivannaranjo: This could be just a a Func<T, Task<T>> actually, which means you don't need to define this delegate at all.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
```diff
@@ -0,0 +1,168 @@
...
+        public bool HasChanges;
+
+        public AuthorizedNetworkChange(IList<AuthorizedNetworkModel> authorizedNetworks)
+        {
+            // Check if any authorized networks are new or were updated.
+            HasChanges = authorizedNetworks.Any(x => x.NewOrUpdated);
+
+            // Get a list of all of the networks that weren't deleted.
+            AuthorizedNetworks = authorizedNetworks.Where(x => !x.Deleted)
+                .Select(x => new AclEntry() { Name = x.Name, Value = x.Value }).ToList();
```
ivannaranjo: Nit, you can remove the (), just `new AclEntry{ ... }` will be clearer.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
```diff
@@ -0,0 +1,168 @@
...
+    /// This class is the view model for the <seealso cref="AuthorizedNetworksWindow"/> dialog.
+    /// </summary>
+    internal class AuthorizedNetworksViewModel : ViewModelBase
+    {
+        private readonly AuthorizedNetworksWindow _owner;
+
+        /// <summary>
+        /// The list of authorized networks.
+        /// </summary>
+        public ObservableCollection<AuthorizedNetworkModel> Networks
```
ivannaranjo: Not sure you need this property to be mutable, as you expose the collection and then mutate the collection itself.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        /// <summary>
+        /// Convert a list of <seealso cref="AclEntry"/>s to <seealso cref="AuthorizedNetworkModel"/>s.  Used
+        /// to more easily display and manipulate the entries in the UI.
+        /// </summary>
+        /// <param name="instance"></param>
+        /// <returns></returns>
+        private ObservableCollection<AuthorizedNetworkModel> GetAuthorizedNetworks(DatabaseInstance instance)
+        {
+            IList<AclEntry> acls = instance?.Settings?.IpConfiguration?.AuthorizedNetworks ?? new List<AclEntry>();
```
ivannaranjo: If you make this `IEnumerable<AclEntry>` you could just use `Enumerable.Empty<AclEntry>()` as the default value.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        /// <summary>
+        /// Called when the user clicks the Save button.  This populates the results field and closes the dialog.
+        /// </summary>
+        private void OnSaveCommand()
+        {
+            Result = new AuthorizedNetworkChange(Networks);
+            _owner.Close();
+
```
ivannaranjo: Extra white space.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
```diff
@@ -0,0 +1,102 @@
...
+                        <Grid.RowDefinitions>
+                            <RowDefinition Height="*" />
+                            <RowDefinition Height="*"/>
+                        </Grid.RowDefinitions>
+                        <Grid.ColumnDefinitions>
+                            <ColumnDefinition Width="*" />
+                            <ColumnDefinition Width="*" />
+                        </Grid.ColumnDefinitions>
+
+                        <TextBlock Grid.Row="0" Grid.Column="0" Margin="5px, 2px"
```
ivannaranjo: You don't need to specify the `Grid.Row` and `Grid.Column` if they are 0, that's the default value.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
```diff
@@ -0,0 +1,102 @@
...
+                    <Grid Grid.Row="1">
+                        <Button Margin="5px, 2px" Width="Auto" Command="{Binding AddNetwork}"
+                                Content="{x:Static ext:Resources.AuthorizedNetworksWindowAddNetworkButton}"
+                                HorizontalAlignment="Center" VerticalAlignment="Center"/>
+                    </Grid>
+                </Grid>
+            </Border>
+
+            <Grid Grid.Row="2" Margin="3px, 5px">
+                <ScrollViewer VerticalScrollBarVisibility="Auto">
```
ivannaranjo: Why not use a ListView? You shouldn't need to have to specify the `ScrollViewer` yourself.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
```diff
@@ -43,32 +45,78 @@ internal class InstanceViewModel : TreeHierarchy, ICloudExplorerItemSource
...
+
+            try
             {
-                new MenuItem { Header = Resources.CloudExplorerSqlOpenAddDataConnectionMenuHeader, Command = _openAddDataConnectionDialog },
-                new MenuItem { Header = Resources.UiOpenOnCloudConsoleMenuHeader, Command = new WeakCommand(OnOpenOnCloudConsoleCommand) },
-                new MenuItem { Header = Resources.UiPropertiesMenuHeader, Command = new WeakCommand(OnPropertiesCommand) },
-            };
-            ContextMenu = new ContextMenu { ItemsSource = menuItems };
+                // Poll until the update to completes.
+                Polling<Operation>.Fetch fetch = (o) => dataSource.GetOperationAsync(o.Name);
```
ivannaranjo: Shouldn't this just be an await on the operation, or a method that takes an operation and just returns a Task, without having to specify why or how to poll? It seems that this is exposing a bit too much of the implementation details of the polling operation.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
```diff
@@ -82,6 +130,38 @@ private void OnPropertiesCommand()
...
+            }
+
+            ExtensionAnalytics.ReportCommand(
+                CommandName.UpdateCloudSqlAuthorizedNetworks, CommandInvocationSource.Button);
+            IList<AclEntry> updatedNetworks = networkChange.AuthorizedNetworks;
+            DatabaseInstanceExtensions.UpdateAuthorizedNetworks(Instance, updatedNetworks);
+            Task<Operation> operation = _owner.DataSource.Value.UpdateInstanceAsync(Instance);
+            string action = Resources.CloudExplorerSqlUpdatedAthorizedNetworksCaption;
+            string errorMessage = Resources.CloudExplorerSqlUpdateAthorizedNetworksErrorMessage;
+            PollOperation(operation, action, errorMessage);
```
ivannaranjo: Should await the result of the `PollOperation` method.

ivannaranjo:
Took another look, some comments.

Not sure why appveyor is failing for you, perhaps is an older appveyor.yml?

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done.  This makes things cleaner, thanks!

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <summary>
+        /// The list of authorized networks.
+        /// </summary>
+        public ObservableCollection<AuthorizedNetworkModel> Networks
+        {
+            get { return _networks; }
+            set { SetValueAndRaise(ref _networks, value); }
+        }
+        private ObservableCollection<AuthorizedNetworkModel> _networks;
```
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
iantalarico: I prefer to keep them, especially in larger files I think it makes things more explicit.   If you feel strongly thought I can remove them

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
iantalarico: The listview does not seem to properly handle the scrolling.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
iantalarico: That would work, but the operation doesn't know how to poll.  If we did it this way it would make this very specific, this will allow me to reuse this for other operations

iantalarico:
Thanks PTAL.

I'm a few commits behind head, that may be it.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
iantalarico: All of the error handling is happening in the PollOperation function.  Should I move it here? 

I think this will allow me to more generally handle polling SQL operations when I implement more.

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
```diff
@@ -0,0 +1,124 @@
...
+            while (true)
+            {
+                // Check if the current result is in a stopable state.
+                if (stopPolling.Invoke(resource))
+                {
+                    break;
+                }
+
+                // If a cancellation token is present and a cancelation has been requested stop polling.
+                if (token.IsCancellationRequested)
```
ivannaranjo: Actually, prefer using `token.ThrowIfCancellationRequested()` here. That way you can unwind the stack. The caller, who will use `await` to get the result, will get notified of the cancellation that way and can perform cleanup. As you have it right now there's no way to know that there was a cancellation.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworkModel.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// </summary>
+        public ICommand UndoDeleteCommand { get; }
+
+        /// <summary>
+        /// True if this instance is to be deleted.
+        /// </summary>
+        public bool Deleted
+        {
+            get { return _toBeDeleted; }
+            set { SetValueAndRaise(ref _toBeDeleted, value); }
```
ivannaranjo: Should also use `RaisePropertyChanged(nameof(NotDeleted))` so the `NotDeleted` property is updated in the UI if bound.

ivannaranjo: Should probably do the same for any other property that depends on it, like `NewOrUpdated`.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworkModel.cs
```diff
@@ -0,0 +1,110 @@
...
+            Deleted = false;
+        } 
+
+        public override string ToString()
+        {
+            if (String.IsNullOrWhiteSpace(Name))
+            {
+                return Value;
+            }
+            return $"{Name} ({Value})";
```
ivannaranjo: Maybe collapse into `String.IsNullOrWhiteSpace(name) ? Value : $"{Name} ({Value})"`

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
ivannaranjo: It is more idiomatic actually to not specify properties with the default value. Only when you need to change said default value is when you need to do so.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
ivannaranjo: Then we should be looking at why, but having your own `ScrollViewer` and `ItemsControl` is too low level for what you want. Perhaps you need to look into the scroll properties for the `ListView` to see if you can enable the behavior. If you need to dig deeper then look into customizing the template for the control. The thing is that the `ListView` will also bring more behaviors such as selection, etc...

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
```diff
@@ -43,32 +45,77 @@ internal class InstanceViewModel : TreeHierarchy, ICloudExplorerItemSource
...
+            UpdateIcon();
+        }
+
+        /// <summary>
+        /// Poll the status of a pending operation until it is complete.
+        /// </summary>
+        /// <param name="task">The current operation to watch and poll</param>
+        /// <param name="action">A string representation of what the operation is doing for user display</param>
+        /// <param name="errorMessage">A user friendly error message to show to users if a failure occurs</param>
+        private async void PollOperation(Task<Operation> task, string action, string errorMessage)
```
ivannaranjo: Should probably return `Task` so you can await the result of the `PollOperation()` method.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
ivannaranjo: I think that you need to probably handle the exceptions in the caller, and show the right error message. At this point you are using the same error message for everything, which is a bit confusing.

I am still confusing a bit, I think that you should have something that converts an operation into a task that you can await, and hide all of the details in the data source if necessary. There's a bit too much logic in the callers.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksWindowContent.xaml
iantalarico: Done, not sure what I was doing last time.  It was a simple fix

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
iantalarico: I refactored this a bit. Done.

GoogleCloudExtension/GoogleCloudExtension.Utils/Polling.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworkModel.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworkModel.cs
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/CloudSQL/InstanceViewModel.cs
iantalarico: Done

iantalarico:
Fixed.  PTAL

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
ivannaranjo: The property still has a public setter, it should not have one since you are not going to change the collection itself.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// <summary>
+    /// This class is the view model for the <seealso cref="AuthorizedNetworksWindow"/> dialog.
+    /// </summary>
+    internal class AuthorizedNetworksViewModel : ViewModelBase
+    {
+        private readonly AuthorizedNetworksWindow _owner;
+
+        private ObservableCollection<AuthorizedNetworkModel> _networks;
```
ivannaranjo: Mark `_networks` as readonly since it will never  change after initialization. In fact you could get rid of this field all together and just use:
```C#
ObservableCollection<AuthorizedNetworkModel> Networds { get; } = new ObservableCollection<AuthorizedNetworkModel>();
```

ivannaranjo:
Minor thing, otherwise it looks good. Merge once you have the green.

ivannaranjo:
And sorry for the delay.

GoogleCloudExtension/GoogleCloudExtension/AuthorizedNetworkManagement/AuthorizedNetworksViewModel.cs
iantalarico: Done

iantalarico:
Looks like I just needed to merge in q3_master.  All green now.

PR #160, opened 2016-08-10
Fixing application name

### LINKED ISSUE #159: Fix the value passed in as the environment for server side metrics.
It seems that the current value "Google Cloud Tools for Visual Studio" is completely wrong, it should be a single token. Investigating.

ivannaranjo:
This fixes #159 and hopefully allows the server side parsing to succeed.


iantalarico:
LGTM

PR #164, opened 2016-08-17
List machines in a flat list

ivannaranjo:
This PR changes the way the GCE instances are listed by default to show a plain list of instances. The user can still choose to show per zone by using a new context menu option to do so. For completeness the option to filter to only Windows instances is also added to this context menu. 

The options in the context menu are dynamically changed depending on the properties values, instead of using checked options. It feels easier to understand this way.

See the new options in the context menu here:
screen shot 2016-08-17 at 4 30 19 pm

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSource.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSource.cs
@@ -29,7 +29,6 @@ namespace GoogleCloudExtension.CloudExplorerSources.Gce
         private static readonly Lazy<ImageSource> s_instancesOnlyButtonIcon = new Lazy<ImageSource>(() => ResourceUtils.LoadImage(InstancesOnlyButtonIconPath));
 
         private readonly ButtonDefinition _windowsOnlyButton;
-        private readonly ButtonDefinition _instancesOnlyButton;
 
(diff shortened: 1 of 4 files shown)


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSource.cs
```diff
@@ -23,8 +23,10 @@ namespace GoogleCloudExtension.CloudExplorerSources.Gce
     internal class GceSource : CloudExplorerSourceBase<GceSourceRootViewModel>
     {
         private const string WindowsOnlyButtonIconPath = "CloudExplorerSources/Gce/Resources/filter.png";
+        private const string InstancesOnlyButtonIconPath = "CloudExplorerSources/Gce/Resources/filter.png";
```
iantalarico: This is the same image.  Should it be? Do you need to redefine the var?

iantalarico:
Just one small comment otherwise LGTM

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSource.cs
ivannaranjo: That's dead code actually, I'll remove it.

PR #165, opened 2016-08-18
Adding support to show buckets organized per location

ivannaranjo:
This PR adds support to show the buckets organized per location, mirroring the GCE data source. Also the `Location` property is added to the bucket's item so the user can see it in the properties window.

screen shot 2016-08-18 at 2 00 55 pm


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gcs/BucketLocationViewModel.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Media;
+
+namespace GoogleCloudExtension.CloudExplorerSources.Gcs
+{
+    internal class BucketLocationViewModel : TreeHierarchy
```
iantalarico: Class comment?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gcs/BucketLocationViewModel.cs
```diff
@@ -0,0 +1,35 @@
...
+        public BucketLocationViewModel(GcsSourceRootViewModel owner, string name, IEnumerable<BucketViewModel> buckets)
+        {
+            _owner = owner;
+            _buckets = buckets;
+
+            Caption = $"{name} ({buckets.Count()})";
+            Icon = s_zoneIcon.Value;
+            foreach (var bucket in _buckets)
+            {
+                Children.Add(bucket);
```
iantalarico: I remember talks about limiting files, should we also be careful about buckets? I have some projects with large number of buckets.  I'm not familiar with VS performance and large lists so maybe 200 buckets will be okay.

ivannaranjo: Should be fine as the UI is virtualized, but the loading can take quite a bit of time. I think that you bring a very good point and perhaps we have to have a way to "page in" buckets. This also applies to VMs and any other resource for which we can have 100s of items.

ivannaranjo: I have open #167 to track the feature of listing 100s of instances, buckets, Cloud SQL databases, etc... To have a common UI paradigm for this case.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gcs/BucketLocationViewModel.cs
ivannaranjo: Typically not being doing this for `internal` classes. Not a bad thing to start doing.

PR #166, opened 2016-08-18
Hide empty zones

ivannaranjo:
This PR hides the empty zones from the GCE listings, including those zones that become empty when the user filters out the non-Windows machines.

The PR accomplishes this but doing a bit of refactoring on how the zone view models are created, and how the queries that process the data to refresh the UI.

There's also a small cosmetic change, in which the GCE instances, or the zones, are sorted alphabetically.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSourceRootViewModel.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSourceRootViewModel.cs
@@ -176,6 +176,11 @@ namespace GoogleCloudExtension.CloudExplorerSources.Gce
             Process.Start(url);
         }
 
+        private void OnStatusCommand()
+        {
+            Process.Start("https://status.cloud.google.com/");
+        }
+
         public override void InvalidateProjectOrAccount()
         {
             Debug.WriteLine("New credentials, invalidating data source for GCE");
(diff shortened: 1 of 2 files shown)


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSourceRootViewModel.cs
```diff
@@ -176,6 +176,11 @@ private void OnNewInstanceCommand()
             Process.Start(url);
         }
 
+        private void OnStatusCommand()
```
iantalarico: No need to do this now, but we should make this a util somewhere.  I know we do this in many places.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSourceRootViewModel.cs
```diff
@@ -226,73 +231,43 @@ private GceDataSource CreateDataSource()
 
         private void PresentViewModels()
         {
+            IEnumerable<TreeNode> viewModels;
             if (_showZones)
             {
-                PresentZoneViewModels();
+                viewModels = _instancesPerZone
```
iantalarico: This is a little hard to follow as it is such a large query, I would consider breaking it up into parts but up to you

iantalarico:
LGTM, nice refactor

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSourceRootViewModel.cs
ivannaranjo: I'll add a comment to it. I tried to make it a step by step query but clearly failed :). Will see what I can do.

PR #169, opened 2016-08-19
Stored windows credentials

ivannaranjo:
This PR adds functionality to save the Windows credentials created through the "Create or Reset" functionality so the user can retrieve past created accounts. These credentials are stored in an encrypted form using the [`System.Security.Cryptography.ProtectedData`](https://msdn.microsoft.com/en-us/library/system.security.cryptography.protecteddata.aspx) class.

The passwords will only be accessible for the user that created them.

This feature does not solve the issue of two users creating an account with the same name in two different machines. The last account will win, rendering the previous account useless.

The feature is surfaced to the user in the context menu for the VM and it replaces the "Create or Reset" command in beta:
screen shot 2016-08-19 at 1 22 17 pm

The dialog to manage the accounts looks like this:
screen shot 2016-08-19 at 1 23 11 pm

Features that need Windows credentials to function can, in theory, access the central repository. I am looking into how to integrate this into the Terminal Server functionality for example, generating an .rdp file automatically with the username, password and IP address already prepopulated.


GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// In memory cache of the credentials for the current credentials (account and project pair).
+        /// </summary>
+        private readonly Dictionary<string, IEnumerable<WindowsInstanceCredentials>> _credentialsForInstance = new Dictionary<string, IEnumerable<WindowsInstanceCredentials>>();
+
+        public static WindowsCredentialsStore Default => s_defaultStore.Value;
+
+        /// <summary>
+        /// Loads the list of Windows credentials associated with <paramref name="instance"/>.
+        /// </summary>
+        /// <param name="instance">The instance</param>
```
iantalarico: maybe the GCE instance?

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// </summary>
+        /// <param name="instance">The instance.</param>
+        /// <param name="credentials">The credentials to store.</param>
+        public void AddCredentialsToInstance(Instance instance, WindowsInstanceCredentials credentials)
+        {
+            var instancePath = GetInstancePath(instance);
+            var fullInstancePath = Path.Combine(s_credentialsStoreRoot, instancePath);
+
+            SaveEncryptedCredentials(fullInstancePath, credentials);
+            _credentialsForInstance.Remove(instancePath);
```
iantalarico: Why not update the cache instead of invalidating it?

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -0,0 +1,153 @@
...
+            var credentialsPath = Path.Combine(fullInstancePath, GetFileName(credentials));
+
+            if (File.Exists(credentialsPath))
+            {
+                File.Delete(credentialsPath);
+                _credentialsForInstance.Remove(instancePath);
+            }
+        }
+
+        private WindowsInstanceCredentials LoadEncryptedCredentials(string path)
```
iantalarico: I like comments on private functions but not required

GoogleCloudExtension/GoogleCloudExtension/ManageWindowsCredentials/ManageWindowsCredentialsViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+                userName: SelectedCredentials.User,
+                password: SelectedCredentials.Password,
+                instanceName: _instance.Name);
+        }
+
+        private void OnDeleteCredentialsCommand()
+        {
+            if (!UserPromptUtils.YesNoPrompt(
+                String.Format(Resources.ManageWindowsCredentialsDeleteCredentialsPromptMessage, SelectedCredentials.User),
+                Resources.ManageWindowsCredentialsDeleteCredentialsPromptTitle))
```
iantalarico: nit: tab in?

iantalarico:
Very nice, just a few small nits and one question about the caching.  Sorry for the slow turn around!

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ivannaranjo: It is actually cleaner this way, it ensures that the cache contents are only read in one place, and only during one method `GetCredentialsForInstance()`. At the end is the same cost.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ivannaranjo: Good point, will do.

ivannaranjo:
No worries about the delay, I appreciate that you are busy and are really thankful that you can look through the PR :).

GoogleCloudExtension/GoogleCloudExtension/ManageWindowsCredentials/ManageWindowsCredentialsViewModel.cs
ivannaranjo: Ug, yes.

iantalarico:
LGTM :)

PR #170, opened 2016-08-22
User useragent

ivannaranjo:
Specifying the user agent to use when reporting analytics to match the overall user agent for the extension.


iantalarico:
LGTM

PR #171, opened 2016-08-24
Use stored credentials to launch Terminal Server session

ivannaranjo:
This PR adds support for using stored Windows VM credentials to launch a Terminal Server session. This is accomplished by creating a new .rdp file, a Terminal Server connection file, which contains the address of the machine to connect to, the user name and the encrypted password.

The password is encrypted by using the [System.Security.Cryptography.ProtectedData](https://msdn.microsoft.com/en-us/library/system.security.cryptography.protecteddata%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) class. The Terminal Server client (`mstsc.exe`) expects the password in the .rdp file encrypted in this manner.

With this PR the user will no longer need to copy the password from the show password dialog and pasting it in the Terminal Server window. 

There's a new dialog that allows the user to choose a set of credentials:
screen shot 2016-08-24 at 12 03 42 pm

By clicking on the "..." button the user will get access to the manage credentials dialog, upon creating a new set of credentials, the user will be able to choose it and use it to start the Terminal Server session:
screen shot 2016-08-24 at 12 04 17 pm


GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -112,6 +112,12 @@ public void DeleteCredentialsForInstance(Instance instance, WindowsInstanceCrede
             }
         }
 
+        public string GetStoragePathForInstance(Instance instance)
```
iantalarico: Replace the above calls with this?

var instancePath = GetInstancePath(instance);
var fullInstancePath = Path.Combine(s_credentialsStoreRoot, instancePath);

GoogleCloudExtension/GoogleCloudExtension/TerminalServer/TerminalServerManager.cs
```diff
@@ -0,0 +1,100 @@
...
+                // The user name to use.
+                writer.WriteLine($"username:s:{credentials.User}");
+
+                // The encrypted password to use.
+                writer.WriteLine($"password 51:b:{EncryptPassword(credentials.Password)}");
+            }
+        }
+
+        /// <summary>
+        /// Encrypts the given password a form that mstsc.exe will accept.
```
iantalarico: nit: in a form

iantalarico:
Just a few small comments. But LGTM.

What is the flow if a password is reset from another location? This flow will just fail, right?  Yann any input?

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ivannaranjo: Centralizing the logic to create the full path does makes sense. Done.

PR #172, opened 2016-08-24
Save .publishsettings file with credentials from the Windows VM credentials store

ivannaranjo:
This PR adds functionality to save .publishsettings files to publish ASP.NET applications to GCE VMs with the credentials already populated from the Windows VM credentials store.

The user is able to choose the credentials using the choose credentials dialog like so:
screen shot 2016-08-24 at 2 24 59 pm

Then the user will choose where to store the file:
screen shot 2016-08-24 at 2 26 09 pm

The file name created combines the name of the machine and the name of the user selected to produce a unique name. This file can then later be imported with the "Publish..." dialog as before. The user name and password will be pre-populated.

Notice that the .publishsettings file produced contains the password in plain text because this is required by the import process. This means that the user is then responsible for deleting these files once they are no longer needed. The password is stored encrypted in the publishing profile, with a key associated with the user and machine.

This PR build on top of #171.


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceInstanceViewModel.cs
```diff
@@ -231,6 +231,37 @@ private void UpdateContextMenu()
...
+                    Title = Resources.CloudExplorerGceSavePubSettingsCredentialsTitle,
+                    Message = Resources.CloudExplorerGceSavePubSettingsCredentialsMessage,
+                });
+            if (credentials == null)
+            {
+                Debug.WriteLine("User canceled when selecting credentials.");
+                return;
+            }
+
+            var storePath = PromptForPublishSettingsPath($"{Instance.Name}-{credentials.User}");
```
iantalarico: Shouldn't we use "Project-Instance-User"?  Otherwise user and machine names could conflict between projects?

iantalarico:
"Notice that the .publishsettings file produced contains the password in plain text because this is required by the import process"

This makes me uncomfortable.  I think this should be explicitly called out in some place.  If someone doesn't know this or isn't paying attention all the machine creds could be stored in plain text on their dev machine.

Again sorry for the slow turn around!

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceInstanceViewModel.cs
ivannaranjo: Actually that is a really good point, will do that.

ivannaranjo:
I have opened #173 to track showing the user that the password is not encrypted.

ivannaranjo:
Before merging this I want to verify that I can generate a .publishsettings with an encrypted password. It seems that there _could_ be a way.

ivannaranjo:
Verified, the password cannot be encrypted. The user will have to be careful with these files.

PR #186, opened 2016-09-13
Pub sub

ILMTitan:
Add Pubsub to the base of the Google Cloud Tools window. Add a list of topics under that node. Add a list of subscriptions under each topic.


ivannaranjo:
Please change the base tree, do this against q3_master instead, thats' where we will ship from when we're ready.

ILMTitan:
PTAL

.gitignore
```diff
@@ -9,3 +9,4 @@
 **/project.lock.json
 .DS_Store
 **/packages/*
+/GoogleCloudExtension/GoogleCloudExtension.sln.DotSettings.user
```
ivannaranjo: Thanks for adding this, maybe add the generic version? **.DotSettings.user?

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,98 @@
...
+using Google.Apis.Pubsub.v1;
+using Google.Apis.Pubsub.v1.Data;
+using GoogleCloudExtension.DataSources;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    public class PubSubDataSource : DataSourceBase<PubsubService>
```
ivannaranjo: Add a small class comment.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// Initializes a new data source that connects to Google Cloud Pub/Sub.
+        /// </summary>
+        /// <param name="projectId">The id of the project to connect to.</param>
+        /// <param name="credential">The Google Credential to connect to.</param>
+        /// <param name="applicationName">The name of the application. </param>
+        public PubSubDataSource(string projectId, GoogleCredential credential, string applicationName) :
+            base(projectId, credential, init => new PubsubService(init), applicationName)
+        { }
+
+        public Task<IList<Topic>> GetTopicListAsync()
```
ivannaranjo: Add a method comment.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,98 @@
...
+                token =>
+                {
+                    request.PageToken = token;
+                    return request.ExecuteAsync();
+                },
+                response => response.Topics,
+                response => response.NextPageToken);
+        }
+
+        public Task<IList<Subscription>> GetSubscriptionListAsync()
```
ivannaranjo: Add comments and all the rest of public methods.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            ProjectsResource.SubscriptionsResource.ListRequest request =
+                Service.Projects.Subscriptions.List($"projects/{ProjectId}");
```
ivannaranjo: Consider having a method that returns the resource name for the project from the project ID.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public async Task<IList<ReceivedMessage>> PullSubscriptionMessagesAsync(Subscription subscription)
+        {
+            ProjectsResource.SubscriptionsResource.PullRequest request =
+                Service.Projects.Subscriptions.Pull(new PullRequest(), subscription.Name);
+            PullResponse response = await request.ExecuteAsync();
+            return response.ReceivedMessages;
+        }
+
+        public Task<Empty> RemoveMessagesAsync(
```
ivannaranjo: This is the wrong name of the method it should probably be `AcknowledgeMessagesAsync()` since that is what it is doing.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,98 @@
+﻿// // Copyright 2016 Google Inc. All Rights Reserved.
```
ivannaranjo: Double comment on the licenses? Here and other files? Do you mind running the tools/ensure_licenses.py script to ensure the licenses in your files?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+            Caption = Resources.CloudExplorerPubSubNoTopicsFoundCaption,
+            IsWarning = true
+        };
+        private static readonly TreeLeaf s_errorPlaceholder = new TreeLeaf
+        {
+            Caption = Resources.CloudExplorerPubSubListTopicsErrorCaption,
+            IsError = true
+        };
+
+        public override string RootCaption { get; } = Resources.CloudExplorerPubSubRootCaption;
```
ivannaranjo: You can use the => notation to avoid the storage, since these properties are really just aliases for the values.

ivannaranjo: as in
```C#
public override string RootCaption => Resources.CloudExplorerPubSubRootCaption;
```

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            else
+            {
+                return null;
+            }
+        }
+
+        public override void Initialize(ICloudSourceContext context)
+        {
+            base.Initialize(context);
```
ivannaranjo: This method doesn't do much, you can remove it.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public override void InvalidateProjectOrAccount()
+        {
+            Debug.WriteLine("New credentials, invalidating the GCS source.");
```
ivannaranjo: Should say "invalidating the PubSub source."

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+            IsError = true
+        };
+
+        public object Item => _item.Value;
+        public event EventHandler ItemChanged;
+
+        public TopicViewModel(PubSubSourceRootViewModel owner, Topic topic)
+        {
+            _owner = owner;
+            _item = new Lazy<TopicItem>(() => new TopicItem(topic));
```
ivannaranjo: This is actually overkill, you can create it on the spot, it is a very cheap object.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private async Task<IList<SubscriptionViewModel>> ListSubscriptionViewModelsAsync()
+        {
+            Task<IList<string>> topicSubscriptionNamesTask =
+                _owner.DataSource.GetTopicSubscriptionListAsync(_item.Value.Name);
+            IList<Subscription> allSubscriptions = await _owner.Subscriptions;
+            IList<string> topicSubscriptionNamesList = await topicSubscriptionNamesTask;
+            ISet<string> topicSubscriptionNames = new HashSet<string>(topicSubscriptionNamesList);
+            var topicSubscriptions = new List<SubscriptionViewModel>();
```
ivannaranjo: This could be simplified to a LINQ query.
```C#
allSubscriptions.Where(x => topicSubscriptionNames.Contains(x.Name)).Select(x => new SubscriptionMode....)
```

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+            _dataSource = new Lazy<PubSubDataSource>(CreateDataSource);
+        }
+
+        protected override async Task LoadDataOverride()
+        {
+            try
+            {
+                _subscriptions = new Lazy<Task<IList<Subscription>>>(GetSubscriptionsAsync);
```
ivannaranjo: You don't need the Lazy<T> here, you can store the Task<T> in the field then await it when you need the value. In fact starting the load process as soon as possible will be good so you have the data by the time you need it. You can await the same task multiple times if necessary.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+using GoogleCloudExtension.CloudExplorer;
+using GoogleCloudExtension.Utils;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Windows.Media;
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    internal class TopicViewModel : SourceRootViewModelBase
```
ivannaranjo: `SourceRootViewModel` is the wrong base class here, you should look at the `BucketViewModel` class as a model. You should derive from `TreeHierarchy` and `ICloudExplorerItemSource`.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                if (topicSubscriptionNames.Contains(subscription.Name))
+                {
+                    topicSubscriptions.Add(new SubscriptionViewModel(this, subscription));
+                }
+            }
+            return topicSubscriptions;
+        }
+
+        protected virtual void OnItemChanged()
```
ivannaranjo: Do you intend for your item to change at all? If so the `Item` property setter should be the one that raises this event.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using Google.Apis.Pubsub.v1.Data;
+using GoogleCloudExtension.CloudExplorer;
+using GoogleCloudExtension.Utils;
+using System;
+using System.Windows.Media;
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    internal class SubscriptionViewModel : TreeLeaf
```
ivannaranjo: You should implement `ICloudExplorerItemSource` so the system knows that you have an item to display when selected.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionItem.cs
```diff
@@ -0,0 +1,37 @@
...
+// // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// // See the License for the specific language governing permissions and
+// // limitations under the License.
+
+using Google.Apis.Pubsub.v1.Data;
+using GoogleCloudExtension.Utils;
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    internal class SubscriptionItem
```
ivannaranjo: You should derive from `PropertyWindowItemBase` that way you will get the right behaviors in the property window, such as the class name, item name, etc...

ivannaranjo: See the [`BucketItem`](https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/blob/df3aeee5c5daca25687b98ed1d78544bb322f166/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gcs/BucketItem.cs#L20) class for more info.

ivannaranjo:
Left some comments, let's iterate on this as we go.

ILMTitan:
PTAL

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    /// <summary>
+    /// Data source that returns information about Pubsub resources.
+    /// </summary>
+    public class PubsubDataSource : DataSourceBase<PubsubService>
+    {
+        private string ProjectResourceName => $"projects/{ProjectId}";
```
ivannaranjo: This seems to me like this could be a race between initialization of the `ProjectId` property and this field. Why not do this in the constructor? You could also move this formatting to a method since I believe it is shared.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSource.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Source class for the Pubsub tree.
+    /// </summary>
+    internal class PubsubSource : CloudExplorerSourceBase<PubsubSourceRootViewModel>
+    {
+        public PubsubSource(ICloudSourceContext context) : base(context) { }
+
+        /// <summary>
+        /// Gets the last part of the full name.
+        /// </summary>
+        internal static string GetPartialName(string fullName)
```
ivannaranjo: The name of this method seems confusing, why not call it what it is? Get the leaf of the path, or get the topic/get subscription, etc...

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,123 @@
...
+                Task<IList<Subscription>> subscriptionsTask = GetSubscriptionsAsync();
+                IEnumerable<Topic> topics = await DataSource.GetTopicListAsync();
+                Subscriptions = await subscriptionsTask;
+                Children.Clear();
+                foreach (var topic in topics)
+                {
+                    Children.Add(new TopicViewModel(this, topic));
+                }
+            }
+            catch (Exception e)
```
ivannaranjo: Better to catch the actual exception type. I do think that we need to have a better high level error handling but it should not be at this level.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                Children.Add(new TreeLeaf { Caption = e.Message, IsError = true });
```
ivannaranjo: Probably better to use a canned error node instead of one with the actual error message from the exception, at they can be quite cryptic.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionViewModel.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        // TODO(Jimwp): Change to Pubsub specific icon.
+        private const string IconResourcePath = "CloudExplorerSources/Gcs/Resources/bucket_icon.png";
+        private static readonly Lazy<ImageSource> s_subscriptionIcon =
+            new Lazy<ImageSource>(() => ResourceUtils.LoadImage(IconResourcePath));
+        private TopicViewModel _owner;
+        private SubscriptionItem _subscriptionItem;
+        public object Item => _subscriptionItem;
+        public event EventHandler ItemChanged;
+
```
ivannaranjo: Extra lines.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ivannaranjo: Disregard, I just realized that this is a property :). Not a bad way of solving it actually.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ILMTitan: I'm removing the TreeLeaf call altogether. It was cleared by the catch in the parent method anyway.

ILMTitan:
PTAL

ILMTitan:
Done.

PR #210, opened 2016-09-30
Adding tool to convert resx to csv.

ivannaranjo:
Simple script that extracts the strings from the .resx file and produces a .csv file that can be imported into spreadsheets for easy review of the strings.

### DIFF
--- a/tools/resx2csv.py
+++ b/tools/resx2csv.py
@@ -4,6 +4,9 @@
 This script will take any .resx file and extract all of the strings to a .csv file
 easily importable into spread sheets.
 
+Example of usage:
+resx2csv.py --resx GoogleCloudExtension\Resources.resx -o strings.csv
+
 """
 
 from __future__ import print_function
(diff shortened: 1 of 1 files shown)


tools/resx2csv.py
```diff
@@ -0,0 +1,54 @@
...
+        self.name = root.attrib['name']
+        self.value = root.find('value').text
+        self.comment = root.find('comment').text
+
+
+def convert_to_csv(root, output):
+    """Takes the root of the Xml file and converts it into a .csv file."""
+
+    with open(output, 'wt') as out:
+        out.write("name,value,comment\n")
```
chrsmith: All other string literals in the file use `'`, whereas this is using `"`.

tools/resx2csv.py
```diff
@@ -0,0 +1,54 @@
+#! /usr/bin/env python
+"""Script to convert the .resx file into a .csv file.
+
+This script will take any .resx file and extract all of the strings to a .csv file
+easily importable into spread sheets.
+
+"""
```
chrsmith: An example command line for how to invoke the tool would be very useful. e.g.:

python .\tools\resx2csv.py --resx .\src\strings\UiStrings.resx --output ui-strings.txt

ivannaranjo: Done.

tools/resx2csv.py
ivannaranjo: Cleaned up the literals, old habits die hard.

ivannaranjo:
I am using Python at the moment out of convenience. But I will definitely think about writing these, and future scripts, in Powershell. Now it is more portable than ever before.

PR #213, opened 2016-10-03
Analytics mandatory fields

ivannaranjo:
This PR adds missing mandatory fields in the GA hit for events.

### DIFF
--- a/GoogleCloudExtension/GoogleAnalyticsUtils/EventsReporter.cs
+++ b/GoogleCloudExtension/GoogleAnalyticsUtils/EventsReporter.cs
@@ -25,8 +25,28 @@ namespace GoogleAnalyticsUtils
     /// </summary>
     public class EventsReporter : IEventsReporter
     {
+        private const string TrueValue = "true";
+        private const string FalseValue = "false";
+
         // The custom dimension index for the various properties sent to Google Analytics.
-        private const int ProjectIdHashIndex = 11;      // The project hash is sent in the custom dimension c11.
+
+        // IsUserSignedIn (cd16): true if a user is signed on, false otherwise.
+        private const int IsUserSignedInIndex = 16;
+
+        // IsInternalUser (cd17): true if the suer is internal to Google, false otherwise.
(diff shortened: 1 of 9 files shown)


GoogleCloudExtension/GoogleAnalyticsUtils/EventsReporter.cs
```diff
@@ -25,9 +25,30 @@ namespace GoogleAnalyticsUtils
     /// </summary>
     public class EventsReporter : IEventsReporter
     {
+        private const string TrueValue = "true";
```
iantalarico: Why not Boolean.TrueString?

GoogleCloudExtension/GoogleAnalyticsUtils/EventsReporter.cs
```diff
@@ -25,9 +25,30 @@ namespace GoogleAnalyticsUtils
...
+        private const string FalseValue = "false";
+
         // The custom dimension index for the various properties sent to Google Analytics.
-        private const int ProjectIdHashIndex = 11;      // The project hash is sent in the custom dimension c11.
 
+        // IsUserSignedIn (cd16): true if a user is signed on, false otherwise.
+        private const int IsUserSignedInIndex = 16;
+
+        // IsInternalUser (cd17): true if the suer is internal to Google, false otherwise.
```
iantalarico: nit: s/suer/user

GoogleCloudExtension/GoogleAnalyticsUtils/EventsReporter.cs
ivannaranjo: Because that's the wrong value. The server expects "true" in lowercase while the value of `Boolean.TrueString` is "True".

GoogleCloudExtension/GoogleAnalyticsUtils/EventsReporter.cs
ivannaranjo: I think suer is more appropriate 😛 but I kid, of course is user.

PR #214, opened 2016-10-04
Adding progress ui

ivannaranjo:
Adding progress UI to be shown while the password is being set of a user.


ivannaranjo:
Here's how the dialog now looks like in the progress state.

progressstate

PR #224, opened 2016-10-07
Add the ability to manage a GAE service's traffic splitting

iantalarico:
Add the ability to manage a GAE service's traffic splitting.

traffic split


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/ServiceViewModel.cs
```diff
@@ -313,6 +359,61 @@ private void OnOpenService()
...
 
+        /// <summary>
+        /// Update a service's traffic split.
+        /// </summary>
+        private async void UpdateTrafficSplit(TrafficSplit split)
+        {
+            IsLoading = true;
+            // Ensure there are no children and they are not expanded so a
+            // reload of children will occur after the split update happens.
+            IsExpanded = false;
```
ivannaranjo: I think just removing the children will do, that way when they are reloaded they can be shown again.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/ServiceViewModel.cs
```diff
@@ -313,6 +359,61 @@ private void OnOpenService()
...
+            Children.Clear();
+            UpdateContextMenu();
+            Caption = Resources.CloudExplorerGaeUpdateTrafficSplitMessage;
```
ivannaranjo: I think the name of this string should have the `LoadingMessage` suffix.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/ServiceViewModel.cs
ivannaranjo: Also, I think we should follow the model of the other hierarchies where we substitute the children for a placeholder loading message, instead of changing the node itself.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/ServiceViewModel.cs
```diff
@@ -64,7 +65,7 @@ class ServiceViewModel : TreeHierarchy, ICloudExplorerItemSource
 
         public readonly GaeSourceRootViewModel root;
 
-        public readonly Service service;
+        public Service service { get; private set; }
```
ivannaranjo: Uppercase `Service` for the property name actually.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/ServiceViewModel.cs
```diff
@@ -313,6 +359,61 @@ private void OnOpenService()
...
+                if (operation.Error != null)
+                {
+                    throw new DataSourceException(operation.Error.Message);
+                }
+                service = await datasource.GetServiceAsync(service.Id);
+            }
+            catch (DataSourceException ex)
+            {
+                IsError = true;
+                UserPromptUtils.ErrorPrompt(ex.Message,
```
ivannaranjo: Instead of prompting update the node state with an error message or placeholder node. The error should be shown inplace. Otherwise a dialog will show up out of nowere.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficModel.cs
```diff
@@ -0,0 +1,58 @@
...
+        private string _versionId;
+        private int _trafficAllocation;
+
+        /// <summary>
+        /// The id of the GAE version.
+        /// </summary>
+        public string VersionId
+        {
+            get { return _versionId;  }
+            set
```
ivannaranjo: You can put the whole setter in a single line, looks better.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficModel.cs
```diff
@@ -0,0 +1,58 @@
...
+        public int TrafficAllocation
+        {
+            get { return _trafficAllocation; }
+            set
+            {
+                SetValueAndRaise(ref _trafficAllocation, value);
+            }
+        }
+
+        public SplitTrafficModel(string versionId, int trafficAllocation)
```
ivannaranjo: Since you have public setters you don't really need to do this in the model constructor. You should also add a default constructor so the model can be created empty.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+{
+    /// <summary>
+    /// The changes made by the user in the <seealso cref="SplitTrafficWindow"/> dialog.
+    /// </summary>
+    public class SplitTrafficChange
+    {
+        /// <summary>
+        /// The map of version Ids to traffic percentages.
+        /// </summary>
+        public IDictionary<string, double?> Allocations;
```
ivannaranjo: This is a public field, I think you mean a public read only property, which would be:
```C#
public IDictionary<string, double?> Allocations { get; }
```

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+
+        /// <summary>
+        /// The mechanism to shard the traffic by.  Either "COOKIE" or "IP"
+        /// </summary>
+        public string ShardBy;
```
ivannaranjo: Same thing, you probably want a readonly property, not a public field, never a public field.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+        }
+    }
+
+    /// <summary>
+    /// This class is the view model for the <seealso cref="SplitTrafficWindow"/> dialog.
+    /// </summary>
+    public class SplitTrafficViewModel : ViewModelBase
+    {
+        private readonly SplitTrafficWindow _owner;
+
```
ivannaranjo: Group all of the fields together.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+            foreach (var allocation in GetAllocations(service))
+            {
+                Allocations.Add(allocation);
+            }
+            foreach (var version in GetAvailableVersions(service, versions))
+            {
+                AvailableVersions.Add(version);
+            }
+
+            Result = null;
```
ivannaranjo: Null is the default value, not a bad thing but is redundant.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+        {
+            _owner = owner;
+            _service = service;
+
+            // Set up the radio buttons properly.  Default to "IP" splitting if none is set.
+            IsCookieChecked = GaeServiceExtensions.ShardByCookie.Equals(service?.Split?.ShardBy);
+            IsIpChecked = !IsCookieChecked;
+
+            // Set up the current allocations and avaible versions.
+            foreach (var allocation in GetAllocations(service))
```
ivannaranjo: `GetAllocations()` and `GetAvailableVersions()` already return an `ObservableCollection<T>`, why not just set the value of the property to their return value directly?

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+            SplitTrafficModel split = new SplitTrafficModel(VersionId, TrafficAllocation);
+            Allocations.Add(split);
+            AvailableVersions.Remove(VersionId);
+
+            // Reset the version and traffic
+            VersionId = null;
+            TrafficAllocation = 0;
+
+            // Ensure the 'HasAvailableVersions' is up to date so the view state is up to date.
+            RaisePropertyChanged(nameof(HasAvailableVersions));
```
ivannaranjo: I would suggest you add an event handler for the `CollectionChanged` event in the `AvailableVersions` collection and raise the the `HasAvailableVersions` property changed event then. This way you will never forget to do so.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+        /// <seealso cref="Allocations"/> and adds the version to the the list
+        /// of <seealso cref="AvailableVersions"/>.
+        /// </summary>
+        private void OnDeleteCommand(SplitTrafficModel sender)
+        {
+            Allocations.Remove(sender);
+            AvailableVersions.Add(sender.VersionId);
+
+            // Ensure the 'HasAvailableVersions' is up to date so the view state is up to date.
+            RaisePropertyChanged(nameof(HasAvailableVersions));
```
ivannaranjo: Same as above.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
```diff
@@ -0,0 +1,147 @@
...
+            </Border>
+
+            <!-- Add version to traffic split section. -->
+            <Border Grid.Row="1" BorderBrush="Black" BorderThickness="1px" Margin="3px, 2px">
+                <Grid Background="#FFECECEC">
+                    <Grid.RowDefinitions>
+                        <RowDefinition Height="50px" />
+                        <RowDefinition Height="25px"/>
+                    </Grid.RowDefinitions>
+                    <Grid>
```
ivannaranjo: The `IsEnabled` property is actually inherited from container to child, the container being `Grid` in this case. So you could just bind `IsEnabled` at the grid level and let all of the controls under it be enabled or disabled. This makes sense since all of the controls are enabled, or disabled, together.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
```diff
@@ -0,0 +1,147 @@
...
+
+                        <TextBlock Grid.Row="1" Margin="5px, 2px" 
+                                   TextWrapping="Wrap"  VerticalAlignment="Center"  
+                                   Text="{x:Static ext:Resources.SplitTrafficWindowTrafficAllocationLabel}"/>
+                        <TextBox Grid.Row="1" Grid.Column="1" Margin="5px, 2px"
+                                 IsEnabled="{Binding HasAvailableVersions}"
+                                 Text="{Binding TrafficAllocation, Mode=TwoWay}"  Height="20px" />
+                    </Grid>
+
+                    <Grid Grid.Row="1">
```
ivannaranjo: Why a `Grid` container for a single button? Set the `Grid.Row` property on the `Button`. `Grid.Row` is an attached property and can be set on anything.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
```diff
@@ -0,0 +1,147 @@
...
+                            <Grid Height="25px">
+                                <Grid.ColumnDefinitions>
+                                    <ColumnDefinition Width="*" />
+                                    <ColumnDefinition Width="*" />
+                                    <ColumnDefinition Width="*" />
+                                </Grid.ColumnDefinitions>
+
+                                <!-- Delete version from traffic split button. -->
+                                <Button Width="75px" Margin="1px" x:Name="deleteButton" VerticalAlignment="Center" HorizontalAlignment="Center"
+                                        Command="{Binding Path=DataContext.DeleteCommand,
```
ivannaranjo: This binding expression seems over complicated to me, but sadly it makes sense since you want to share the `DeleteCommand`. I was thinking more along the lines of having a ViewModel for each element in the list but this works as well.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
```diff
@@ -0,0 +1,147 @@
...
+                                    <ColumnDefinition Width="*" />
+                                    <ColumnDefinition Width="*" />
+                                    <ColumnDefinition Width="*" />
+                                </Grid.ColumnDefinitions>
+
+                                <!-- Delete version from traffic split button. -->
+                                <Button Width="75px" Margin="1px" x:Name="deleteButton" VerticalAlignment="Center" HorizontalAlignment="Center"
+                                        Command="{Binding Path=DataContext.DeleteCommand, 
+                                            RelativeSource={RelativeSource Mode=FindAncestor, AncestorType={x:Type ListView}}}"
+                                            CommandParameter="{Binding ElementName=deleteButton, Path=DataContext}"
```
ivannaranjo: Just use `{Binding}`. This will bind the `CommandParameter` property to the current `DataContext`, which is the item. This also means that you don't need to name the button.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
```diff
@@ -0,0 +1,147 @@
...
+                                        Content="{x:Static ext:Resources.UiDeleteButtonCaption}"/>
+
+                                <!-- Version id. -->
+                                <Grid Grid.Column="1" VerticalAlignment="Center" HorizontalAlignment="Center">
```
ivannaranjo: It seems like you are simulating a horizontal `StackPanel` with this grid (and the one below). Would simplify the Xaml to just use `StackPanel`?

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
```diff
@@ -0,0 +1,147 @@
...
+        <Grid x:Key="content">
+            <Grid.RowDefinitions>
+                <RowDefinition Height="30px" />
+                <RowDefinition Height="85px"/>
+                <RowDefinition Height="*"/>
+            </Grid.RowDefinitions>
+
+            <!-- Shard traffic by selector. -->
+            <Border  BorderBrush="Black" BorderThickness="1px" Margin="3px, 2px">
+                <Grid Background="#FFECECEC">
```
ivannaranjo: This looks like a horizontal `StackPanel` as well.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
iantalarico: I did not realize that, that makes things easy! Done.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
iantalarico: Done

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficWindowContent.xaml
iantalarico: Done.

GoogleCloudExtension/GoogleCloudExtension/SplitTrafficManagement/SplitTrafficModel.cs
ivannaranjo: Nit: the extended building syntax is better for mutable objects, it makes it much more readable as in:
```C#
new SplitTrafficModel
{
  VersionId = ...,
  ...
}
```
But is not critical for this change.

iantalarico: Okay.  I'm going to leave this as is for now as it make the linq easier to read.  I can come back and change this if you feel strongly though.

Issue #198, opened 2016-09-21
Create an unhandled exception handler for the VS extension


ivannaranjo:
Since all of the entry points into the exception are known (all menu handlers and command handlers) we need to ensure that we wrap them with a special try/catch handler to log and diagnose all of the unhandled exceptions. Ensure that we don't crash and keep enough information for diagnostics (at least on the debug version).

ivannaranjo:
This is fixed as of  b10bdfd

PR #226, opened 2016-10-10
Fixing issue 198

ivannaranjo:
This PR adds a wrapper that can be used to harden `Action` instances to ensure that all exceptions that are safe to be handled are handled by showing the user an error message. This ensures that VS will not be brought down by the plugin not catching an exception.

If a critical exception escapes the `Action` then the handler will let it go through as that means that the process cannot continue safely.

This PR deletes the `WeakAction`, `WeakDelegate` and `WeakCommand` classes and replaces them with the `ProtectedAction` and `ProtectedCommand` counterparts. The `Weak*` classes were not actually needed and the `WeakDelegate` as not even used. The `ProtectedCommand` is specially interesting since it ensures that all command handlers are protected by default. Since we're using the MVVM pattern this should cover most of the code.

Event handlers should be hardened by using the `ErrorHandlerUtils.HandleExceptions()` method as shown in the PR.


ivannaranjo:
This fixed #198 and will ensure that the exception does not crashes the IDE unnecesarily.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSource.cs
```diff
@@ -32,22 +32,33 @@ public GceSource(ICloudSourceContext context) : base(context)
...
         }
 
+        #region Event handlers
+
         private void OnShowOnlyWindowsInstancesChanged(object sender, EventArgs e)
         {
-            _windowsOnlyButton.IsChecked = ActualRoot.ShowOnlyWindowsInstances;
+            ErrorHandlerUtils.HandleExceptions(() =>
+            {
+                _windowsOnlyButton.IsChecked = ActualRoot.ShowOnlyWindowsInstances;
```
iantalarico: Does this need to be wrapped in the handler?

GoogleCloudExtension/GoogleCloudExtension/ResetPassword/ResetPasswordWindowContent.xaml.cs
```diff
@@ -38,7 +39,10 @@ public ResetPasswordWindowContent()
 
         private void OnPasswordChanged(object sender, RoutedEventArgs e)
         {
-            ViewModel.Password = _password.Password;
+            ErrorHandlerUtils.HandleExceptions(() =>
+            {
+                ViewModel.Password = _password.Password;
```
iantalarico: And here? Does this need to be wrapped in the handler?

GoogleCloudExtension/GoogleCloudExtension/Utils/ErrorHandlerUtils.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// </summary>
+    public static class ErrorHandlerUtils
+    {
+        /// <summary>
+        /// Runs the given <seealso cref="Action"/> and handles all non-critical exceptions by showing an
+        /// error dialog to the user. If the exception is critical, as determiend by <seealso cref="ErrorHandler.IsCriticalException(Exception)"/>
+        /// then it is re-thrown as this could be that the process is not in a good state to continue executing.
+        /// </summary>
+        /// <param name="action"></param>
+        public static void HandleExceptions(Action action)
```
iantalarico: Should we also report this in the metrics? It would be nice to have an idea of how often this is happening,

GoogleCloudExtension/GoogleCloudExtension/Utils/ProtectedAction.cs
```diff
@@ -0,0 +1,86 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace GoogleCloudExtension.Utils
+{
+    /// <summary>
+    /// This class wraps an <seealso cref="Action"/> with the <seealso cref="ErrorHandlerUtils.HandleExceptions(Action)"/>
+    /// method to handle all exceptions that might escape correctly.
```
iantalarico: super nit: "to handle all exceptions that might escape correctly" This was hard for me to parse.  Maybe "to correctly handle all exceptions that might escape."?

And twice below

GoogleCloudExtension/GoogleCloudExtension/Utils/ProtectedAction.cs
```diff
@@ -0,0 +1,86 @@
+﻿// Copyright 2016 Google Inc. All Rights Reserved.
```
iantalarico: Can this file be branched like weakcommand was?  It seems like that's better for tracking/file history

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gce/GceSource.cs
ivannaranjo: I'm going for full coverage, or at least trying to. So yes, the property can have side effects, maybe not now but in the future. The `HandleExceptions()` method is a defensive move so we need to err on the side of precaution.

GoogleCloudExtension/GoogleCloudExtension/Utils/ProtectedAction.cs
ivannaranjo: Well, the class is completely different though even if it serves a similar purpose. The weak aspect of it is gone because it wasn't actually fixing anything.

ivannaranjo:
PTAL

Issue #234, opened 2016-10-12
Using the project name instead of the project id on the URLs for the new ASP.NET instance


ivannaranjo:
This is making it so the URL will never work.

PR #235, opened 2016-10-12
Using the project id instead of the project name.

ivannaranjo:
In all the cases we should be using the `ProjectId` instead of the `Name` property. The `Name` is the human friendly (and therefore not unique) string but all APIs and pages expect the `ProjectId` which is much more restricted and actually unique.


ivannaranjo:
This PR fixed #234

PR #241, opened 2016-10-13
Adding the external tools to the path.

ivannaranjo:
This PR adds the external tools (bower, grung, node, etc...) that come pre-installed with the [.NET Core tools](https://go.microsoft.com/fwlink/?LinkID=827546) for Visual Studio. By adding this to the path before the "dotnet publish" invocation we allow the publish process to use these tools instead of requiring the user to download and install node.js and bower separately. These are the same tools that are used from Visual Studio when the user published the app that way.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Deployment/NetCoreDeployment.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Deployment/NetCoreDeployment.cs
@@ -15,6 +15,7 @@
 using GoogleCloudExtension.GCloud;
 using GoogleCloudExtension.Utils;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


GoogleCloudExtension/GoogleCloudExtension.Deployment/NetCoreDeployment.cs
```diff
@@ -249,5 +256,11 @@ private static string GetProjectName(string projectPath)
             var directory = Path.GetDirectoryName(projectPath);
             return Path.GetFileName(directory);
         }
+
+        private static string GetExternalToolsPath()
+        {
+            var programFilesPath = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86);
+            return Path.Combine(programFilesPath, @"Microsoft Visual Studio 14.0\Web\External");
```
iantalarico: Will this always be "Microsoft Visual Studio 14.0"? Is there a constant we can use instead?

PR #256, opened 2016-11-01
Misc dialogs

ivannaranjo:
This PR styles some of the common prompt dialogs to use the new common text styles. This PR also adds a new dialog `UserPromptDialogWindow` to prompt the user for Yes/No, Ok and Error prompts. The extension now uses this new dialog instead of the VS built-in one. This way we can style this dialog to match the look and feel of all of our windows.

The dialog looks like this:
image


GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
```diff
@@ -24,27 +23,30 @@ namespace GoogleCloudExtension.Utils
...
-        /// These values come from a comment on https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.shell.vsshellutilities.showmessagebox(v=vs.100).aspx.
-        /// </summary>
-        private const int IDYES = 6;
-        private const int IDNO = 7;
-
-        /// <summary>
         /// Show a message dialog with a Yes and No button to the user.
         /// </summary>
-        /// <param name="message">The message for the dialog.</param>
+        /// <param name="prompt">The message for the dialog.</param>
```
iantalarico: nit: s/message/prompt

ivannaranjo: Ah, thank you.

PR #265, opened 2016-11-07
Update existing dialogs

ivannaranjo:
This PR applies the styles to all of the existing dialogs. There is a bit of re-structuring of the styles to share more between dialogs (such as the common button styles). There are also some fixes in the code behind of some dialogs for some functionality that was not working correctly.

As part of this change all of the dialogs have been changed from hardcoded sizes to use a common set of styles that harmonizes the sizes of the dialogs for the extension. Most of these dialogs are dynamic in size and will grow vertically as needed.

The dialogs affected by this change are:

The Cloud SQL add authorized network dialog, which now looks like this:
screen shot 2016-11-07 at 3 21 20 pm

The firewall options dialog, which now looks like this:
screen shot 2016-11-07 at 3 22 29 pm

The manage accounts dialog, which now looks like this:
image

The MySQL installer window, which now looks like this:
screen shot 2016-11-07 at 3 30 22 pm

The OAUTH cancellation window, which now looks like this:
screen shot 2016-11-07 at 3 31 15 pm

PR #267, opened 2016-11-07
Fixing focus issues for dialogs

### LINKED ISSUE #263: Ensure focus order and keyboard shortcuts are set
Ensure that the focus order makes sense, and that the right elements are focusable.

Also ensure that we enable keyboard shortcuts.

ivannaranjo:
This PR ensures that the right elements in the tree are focusable and adds some missing shortcut keys to the caption of buttons in the main dialogs.

This fixes #263 and the PR is based on PR #265 and contains part of those changes. Will be merged after all other PRs are in.


iantalarico:
This looks very similar to the #265, what are the new changes here?

ivannaranjo:
Merged #265 so the change makes more sense now, PTAL

ivannaranjo:
Added a couple of smaller fixes to this PR as well, since it will likely be that last (or one of the last) ones.

PR #271, opened 2016-11-09
Updating icons strings

ivannaranjo:
Adding the new blue copy icon for the hover state. Updating console name.


GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -138,11 +138,11 @@
     <comment>Menu header for the Enable API command.</comment>
   </data>
   <data name="CloudExploreNoProjectMessage" xml:space="preserve">
-    <value>No projects found, to create a project navigate to the Google Cloud Console.</value>
+    <value>No projects found, to create a project navigate to the Cloud Platform Console.</value>
```
iantalarico: The wording is a little mixed here and below.  In some places it says "... to the Cloud.." and others is just "... to Cloud..."

ivannaranjo: Ok, will remove the "to the" in all instances and merge.

iantalarico: I think it should be "to the" given that Cloud Platform is a proper noun.  But I'm not the best with this sort of thing.  May be useful to talk to a tech writer.

ivannaranjo: For now I am going to change it to _"No projects found, to create a project open the Cloud Platform Console"_ and will consult with the writer that is helping us.

PR #290, opened 2016-11-16
Ga missing analytics

ivannaranjo:
Adding more analytics events to be able to get more insights on how the extension is used.


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gcs/BucketViewModel.cs
```diff
@@ -73,6 +75,8 @@ private void OnPropertiesCommand()
 
         private void OnOpenConCloudConsoleCommand()
```
iantalarico: Feel free to punt to another CL but I just saw this "OnOpenConCloudConsoleCommand".  Should be "OnOpenOn"...

PR #292, opened 2016-11-17
Browse log

Deren-Liao:
Hi Ivan,
I will add a Stackdriver Logging DataSource. It will use Google.Logging.V2. The latest version is 1.19 . With this new version of Logging API, the dependencies must be upgraded too.
Please take a look to make sure it is Okay to (1) use Google.Logging.V2 (2) Upgrade all related packages to version 1.19.

Also, I created another intermediate branch stackdriver_log before merging it into ga_master. Is this good protection or unnecessary overhead?

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.DataSources/packages.config
+++ b/GoogleCloudExtension/GoogleCloudExtension.DataSources/packages.config
@@ -1,14 +1,15 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="BouncyCastle" version="1.7.0" targetFramework="net452" />
-  <package id="Google.Apis" version="1.11.1" targetFramework="net452" />
-  <package id="Google.Apis.Auth" version="1.11.1" targetFramework="net452" />
-  <package id="Google.Apis.CloudResourceManager.v1" version="1.12.0.440" targetFramework="net452" />
-  <package id="Google.Apis.Compute.v1" version="1.12.0.462" targetFramework="net452" />
-  <package id="Google.Apis.Core" version="1.11.1" targetFramework="net452" />
-  <package id="Google.Apis.Plus.v1" version="1.12.0.463" targetFramework="net452" />
-  <package id="Google.Apis.SQLAdmin.v1beta4" version="1.11.1.417" targetFramework="net452" />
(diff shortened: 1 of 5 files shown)


GoogleCloudExtension/GoogleCloudExtension.DataSources/packages.config
```diff
@@ -1,14 +1,15 @@
...
   <package id="BouncyCastle" version="1.7.0" targetFramework="net452" />
-  <package id="Google.Apis" version="1.11.1" targetFramework="net452" />
-  <package id="Google.Apis.Auth" version="1.11.1" targetFramework="net452" />
-  <package id="Google.Apis.CloudResourceManager.v1" version="1.12.0.440" targetFramework="net452" />
-  <package id="Google.Apis.Compute.v1" version="1.12.0.462" targetFramework="net452" />
-  <package id="Google.Apis.Core" version="1.11.1" targetFramework="net452" />
-  <package id="Google.Apis.Plus.v1" version="1.12.0.463" targetFramework="net452" />
-  <package id="Google.Apis.SQLAdmin.v1beta4" version="1.11.1.417" targetFramework="net452" />
-  <package id="Google.Apis.Storage.v1" version="1.12.0.454" targetFramework="net452" />
+  <package id="Google.Apis" version="1.19.0" targetFramework="net452" />
```
ivannaranjo: Updating the version of needed packages is usually not a problem, as long as you don't update Newtonsoft.Json (JSON.NET) as that is still problematic.

ivannaranjo:
As discussed on email you should plan on using the lower level API package for logging instead of the new higher level one as that is still in flux.

ivannaranjo:
In general we should only PR things that are going into the shared branch. Just plan on merging into the shared branch often enough that the PRs are not _too_ big.

GoogleCloudExtension/GoogleCloudExtension.DataSources/packages.config
Deren-Liao: The odd thing was that, when I imported the new Logging package, VS automatically upgraded Google.Apis.dll but not the rest of them.  
But when I tried to run the extension, it gave me error about loading failure.  So I had to manually upgrade the rest of the related package.

PR #302, opened 2016-11-21
Show stackdriver log

Deren-Liao:
Add blank Stackdriver Logging Viewer blank tool window,  LoggingDataSource.cs that encapsulates call to Google.Apis.Logging.V2.  
To mention, I changed the feature branch name to q1_logging. I'll create a different branch for q1_error_reporting when needed.


GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+
+using Google.Apis.Auth.OAuth2;
+using Google.Apis.Logging.v2;
+using Google.Apis.Logging.v2.Data;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.DataSources
+{
+
```
ivannaranjo: Nit: Extra white line.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+        public LoggingDataSource(string projectId, GoogleCredential credential, string appName)
+            : base(projectId, credential, init => new LoggingService(init), appName)
+        { }
+
+        /// <summary>
+        /// The ListLogEntriesRequest page size.
+        /// </summary>
+        public int? PageSize { get; set; }
+
+        private ListLogEntriesRequest CreateRequestFromParams(LogEntryRequestParams requestParams)
```
ivannaranjo: Nit: Try to put the public methods _before_ the private ones in the class. It is a convention that is used throughout the code.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        private class LogEntryRequestParams
+        {
+            public string PageToken { get; set; }
+            public string Filter { get; set; }
+            public string ProjectId { get; set; }
+            public int? PageSize { get; set; }
+        }
+
+        private LogEntryRequestParams _lastSuccessfulRequestParams;
```
ivannaranjo: Having this state in the data source is not a good idea. If you want to enable explicit paging have the caller pass this state back to you when they want to read the next page. The data sources should be a stateless as possible, only containing the project ID and credentials.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <param name="projectId">The Google Cloud Platform project id of the current user account .</param>
+        /// <param name="credential">The credentials to use for the call.</param>
+        public LoggingDataSource(string projectId, GoogleCredential credential, string appName)
+            : base(projectId, credential, init => new LoggingService(init), appName)
+        { }
+
+        /// <summary>
+        /// The ListLogEntriesRequest page size.
+        /// </summary>
+        public int? PageSize { get; set; }
```
ivannaranjo: This should probably be a parameter to the public method so the caller can specify this per call. This would allow you to share the data source.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        /// <summary>
+        /// Returns the next page of all LogEntry items of the project id.
+        /// Please note: The values of other method parameters should be identical to those in the previous call.
+        /// </summary>
+        /// <returns></returns>
+        public async Task<IList<LogEntry>> GetNextPageLogEntryListAsync()
+        {
+            if (_lastSuccessfulRequestParams.ProjectId != ProjectId)
```
ivannaranjo: This cannot happen, if the project changes a new data source will be created.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+            _lastSuccessfulRequestParams.PageToken = response.NextPageToken;
+            return response.Entries;
+        }
+
+        /// <summary>
+        /// Returns the next page of all LogEntry items of the project id.
+        /// Please note: The values of other method parameters should be identical to those in the previous call.
+        /// </summary>
+        /// <returns></returns>
+        public async Task<IList<LogEntry>> GetNextPageLogEntryListAsync()
```
ivannaranjo: This is the method where you should ask the caller to explicitly pass in the enumeration state.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (_lastSuccessfulRequestParams.ProjectId != ProjectId)
+            {
+                throw new DataSourceException("The project id has changed. Fail to fetch next page. Do refresh.");
+            }
+
+            return await MakeAsyncRequest(_lastSuccessfulRequestParams);
+        }
+
+        /// <summary>
+        /// Returns the list of all log entries of the project id.
```
ivannaranjo: Not _"all log entries"_ just the first page. Since you are going towards the direction of explicit paging you should make it clear in all docs.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// This function returns first page of log entries if one page size does not fit.
+        /// </summary>
+        /// <param name="filter">The Google.Apis.Logging.v2.Data.ListLogEntriesRequest Filter</param>
+        public async Task<IList<LogEntry>> GetLogEntryListAsync(string filter)
```
ivannaranjo: This is where you should return the list and the enumeration state so the user can pass it back and fetch the next page.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// Returns the list of MonitoredResourceDescriptor.
+        /// </summary>
+        public async Task<IList<MonitoredResourceDescriptor>> GetResourceDescriptorAsync()
+        {
+            List<MonitoredResourceDescriptor> results = new List<MonitoredResourceDescriptor>();
+            var request = Service.MonitoredResourceDescriptors.List();
+            ListMonitoredResourceDescriptorsResponse response = null;
+            do
```
ivannaranjo: You should use `LoadPagedListAsync()` here to do this page enumeration for you. It will take care of managing the page tokens and all that.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowCommand.cs
```diff
@@ -0,0 +1,114 @@
...
+            // is actually the only one.
+            // The last flag is set to true so that if the tool window does not exists it will be created.
+            ToolWindowPane window = this.package.FindToolWindow(typeof(LogsViewerToolWindow), 0, true);
+            if ((null == window) || (null == window.Frame))
+            {
+                throw new NotSupportedException("Cannot create tool window");
+            }
+
+            IVsWindowFrame windowFrame = (IVsWindowFrame)window.Frame;
+            Microsoft.VisualStudio.ErrorHandler.ThrowOnFailure(windowFrame.Show());
```
ivannaranjo: Remove this part and use the same patterns as the other commands. We have a special class to ensure that exceptions never leave any of our event handlers. That way we cannot crash the extension if we have an unforeseen bug.

GoogleCloudExtension/GoogleCloudExtension/GoogleCloudExtension.vsct
```diff
@@ -68,6 +68,14 @@
         </Strings>
       </Button>
 
+      <!-- This command shows the Stackdriver Logs Viewer tool window. -->
+      <Button guid="guidGoogleCloudExtensionPackageCmdSet" id="cmdidLogsViewerToolWindowCommand" priority="0x0100" type="Button">
+        <Parent guid="guidGoogleCloudExtensionPackageCmdSet" id="GcpToolsGroup" />
```
ivannaranjo: This command will probably need to be disabled if there is no current project selected, which can happen if the user is not logged in or if the user just doesn't have any projects. Think about the zero state as well.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Changed.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: My memory somehow remembers opposite which is wrong. Corrected.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: I'll follow this pattern.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done.

Deren-Liao:
Please take a look again.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,153 @@
...
+            return new ListLogEntriesRequest()
+            {
+                ResourceNames = new List<string>(new string[] { projectsFilter }),
+                Filter = requestParams.Filter,
+                PageSize = requestParams.PageSize,
+                PageToken = requestParams.PageToken
+            };
+        }
+
+        private async Task<Tuple<IList<LogEntry>, string>> MakeAsyncRequest(LogEntryRequestParams requestParams)
```
ivannaranjo: Name should be `MakeRequestAsync`. Every method that returns a `Task<T>` should have a `Sync` suffix.

ivannaranjo: Not sure why you need the `LogEntryRequestParams` class at all, why not just pass the parameters to the method directly?

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    var request = Service.MonitoredResourceDescriptors.List();
+                    request.PageToken = token;
+                    return request.ExecuteAsync();
+                },
+                x => x.ResourceDescriptors,
+                x => x.NextPageToken);
+        }
+
+        private ListLogEntriesRequest CreateRequestFromParams(LogEntryRequestParams requestParams)
```
ivannaranjo: Ditto with the `LogEntryRequestParams`, why not pass the params directly?

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// </summary>
+        /// <param name="filter">Optional. The Google.Apis.Logging.v2.Data.ListLogEntriesRequest Filter</param>
+        /// <param name="pageSize">
+        ///     Optional. specify the page size. 
+        ///     A default value is used on severside if not specified.
+        /// </param>
+        /// <returns>
+        ///     A tuple of :   List of log entries,  optional next page token.
+        /// </returns>
+        public async Task<Tuple<IList<LogEntry>, string>> GetLogEntryListAsync(
```
ivannaranjo: It will probably make sense to create a type for the return type that contains the page token and the first page of results rather than the tuple.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,153 @@
...
+                PageSize = pageSize
+            });
+        }
+
+        /// <summary>
+        /// Returns the list of MonitoredResourceDescriptor.
+        /// The size of entire set of MonitoredResourceDescriptor is small. 
+        /// Batch all in one request in case it spans multiple pages.
+        /// </summary>
+        public async Task<IList<MonitoredResourceDescriptor>> GetResourceDescriptorAsync()
```
ivannaranjo: Name should be plural, as in `GetResourceDescriptorsAsync`.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,153 @@
...
+                PageSize = requestParams.PageSize,
+                PageToken = requestParams.PageToken
+            };
+        }
+
+        private async Task<Tuple<IList<LogEntry>, string>> MakeAsyncRequest(LogEntryRequestParams requestParams)
+        {
+            try
+            {
+                ListLogEntriesRequest requestData = CreateRequestFromParams(requestParams);
```
ivannaranjo: Can't you just do:
``` C#
var response = Service.Entry.List(requestData).ExecuteAsync();
```
Will simplify the code a bit.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,153 @@
...
+                var request = Service.Entries.List(requestData);
+                ListLogEntriesResponse response = await request.ExecuteAsync();
+                return new Tuple<IList<LogEntry>, string>(response?.Entries, response?.NextPageToken);
```
ivannaranjo: Response should never be `null` actually, although the value of `response.Entries` can be `null`. So you don't really need to use the `?.` operator here.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+                var menuCommandID = new CommandID(CommandSet, CommandId);
+                var menuItem = new OleMenuCommand(this.ShowToolWindow, menuCommandID);
+                commandService.AddCommand(menuItem);
+                Action setEnabled = () =>
+                {
+                    menuItem.Enabled = (CredentialsStore.Default?.CurrentAccount != null &&
+                        !string.IsNullOrWhiteSpace(CredentialsStore.Default?.CurrentProjectId));
+                };
+
+                // Not using BeforeQueryStatus for the event may not be fired in some cases.
```
ivannaranjo: You must use `BeforeQueryStatus` actually and then you need to invalidate the state of the commands to force this callback to happen. We will also have to do this when the state of the `CurrentProjectId` property changes to update all menus that depend on this.

We can punt on this for this PR but keep it in mind for future ones.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: When there are several parameter inputs, I agree with what this post proposes "Avoid long parameter -list .   What do you think?
https://big.corp.google.com/~jmcmaster/testing/2016/08/episode-441-avoid-long-parameter-list.html

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: I renamed it to ListLogEntriesAsync.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Looks good to me too. I made the change.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowCommand.cs
Deren-Liao: I saw you already added the CurrentProjectId change event callback that invalidate all commands.  I am changing my code back to responds to BeforeQueryStatus.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
ivannaranjo: That is not idiomatic C#. In C# 6 now we have parameters with default values and you can use the name of the parameter in the call site to remove ambiguities, which makes this pattern of parameter classes completely unnecessary. Please remove them.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <summary>
+        /// Optional "timestamp desc" or "timestamp asc"
+        /// </summary>
+        public string OrderBy;
+    }
+
+    /// <summary>
+    /// Wrap up Log Entries list and the next page token as get log entry list methods result.
+    /// </summary>
+    public class LogEntryRequestResult
```
ivannaranjo: Consider moving this class definition to its own file.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <param name="nextPageToken">
+        /// The page token from last list request response.
+        /// </param>
+        /// <returns>
+        ///     LogEntryRequestResult object that contains log entries and next page token.
+        /// </returns>
+        public async Task<LogEntryRequestResult> GetNextPageLogEntryListAsync(
+            LogEntryRequestParams requestParams, string nextPageToken)
+        {
+            Debug.Assert(requestParams != null);
```
ivannaranjo: The asserts really don't add much since you are throwing.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+            return new ListLogEntriesRequest()
+            {
+                ResourceNames = new List<string>(new string[] { projectsFilter }),
+                Filter = requestParams.Filter,
+                PageSize = requestParams.PageSize,
+                OrderBy = requestParams.OrderBy
+            };
+        }
+
+        private async Task<LogEntryRequestResult> ListLogEntriesAsync(
```
ivannaranjo: Why not make this method public? It seems that other methods are just trivial wrappers around this one.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+                {
+                    var request = Service.MonitoredResourceDescriptors.List();
+                    request.PageToken = token;
+                    return request.ExecuteAsync();
+                },
+                x => x.ResourceDescriptors,
+                x => x.NextPageToken);
+        }
+
+        private ListLogEntriesRequest CreateRequestFromParams(LogEntryRequestParams requestParams)
```
ivannaranjo: You can also just inline this method into `ListLogEntriesAsync()` and simplify the code a bit.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <summary>
+        /// Returns the next page of all LogEntry items of the project id.
+        /// Please note, when calling this method, the filter, pageSize must be same as prior call.
+        /// </summary>
+        /// <param name="requestParams">The request parameters.</param>
+        /// <param name="nextPageToken">
+        /// The page token from last list request response.
+        /// </param>
+        /// <returns>
+        ///     LogEntryRequestResult object that contains log entries and next page token.
```
ivannaranjo: In the annotation you should use the `<seealso>` XML element to refer to other types. So in your return comment it should be:
```XML
<seealso ref="LogEntryRequestResult" /> object that contains...
```

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// Returns the first page of log entries of the project id.
+        /// </summary>
+        /// <param name="requestParams">
+        /// The request parameters.
+        /// NextPageToken i
+        /// </param>
+        /// <returns>
+        ///     LogEntryRequestResult object that contains log entries and next page token.
+        /// </returns>
+        public async Task<LogEntryRequestResult> GetLogEntryListAsync(LogEntryRequestParams requestParams)
```
ivannaranjo: See below, this method is just a trivial wrapper around `ListLogEntriesAsync()` it can be removed.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+
+    /// <summary>
+    /// Wrap up Log Entries list and the next page token as get log entry list methods result.
+    /// </summary>
+    public class LogEntryRequestResult
+    {
+        /// <summary>
+        /// The returned log entries.  It could be null if no logs found by the filter condition.
+        /// </summary>
+        public IList<LogEntry> LogEntries;
```
ivannaranjo: This should be properties, as defined here these are actually fields, you should use:
```C#
public IList<LogEntry> LogEntries { get; }
```

To get a read only property, set the values in the constructor so the instances are immutable.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
```diff
@@ -0,0 +1,182 @@
...
+                },
+                x => x.ResourceDescriptors,
+                x => x.NextPageToken);
+        }
+
+        private ListLogEntriesRequest CreateRequestFromParams(LogEntryRequestParams requestParams)
+        {
+            string projectsFilter = $"projects/{ProjectId}";
+
+            return new ListLogEntriesRequest()
```
ivannaranjo: No need to use the empty () when using the initializers so it can look like this:
```C#
return new ListLogEntriesRequest
{
    ResourceNames = new List<string>(new string[] { projectsFilter }),
    Filter = requestParams.Filter,
    PageSize = requestParams.PageSize,
    OrderBy = requestParams.OrderBy
};
```

Deren-Liao: Didn't notice I had this ()

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LoggingDataSource.cs
Deren-Liao: I'd like to stick to this parameter object.  When I call the two methods, the get next page follows the get first page call.  It is hard requirement for the second call to use exactly same parameters + a next page token.  It is helpful for the caller to use parameter object to save the calling state. If separating them, the caller keeps 3 individual variables.  The code is not concise and easy to get wrong.

Deren-Liao:
Made the change according to your feedback.  Please take a look again.

GoogleCloudExtension/GoogleCloudExtension.DataSources/LogEntryRequestResult.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// Wrap up Log Entries list and the next page token as get log entry list methods result.
+    /// </summary>
+    public class LogEntryRequestResult
+    {
+        /// <summary>
+        /// Initializes an instance of LogEntryRequestResult.
+        /// </summary>
+        /// <param name="logEntries">A list of LogEntry objects. Null is valid input.</param>
+        /// <param name="pageToken">Next page token to retrieve more pages.</param>
+        public LogEntryRequestResult(IList<LogEntry> logEntries, string nextPageToken)
```
ivannaranjo: Nit: Put the properties first. In general follow the order:
* Constants.
* Static fields.
* Instance fields.
* Properties.
* Events.
* Constructors.
* Public methods.
* Private methods.

PR #325, opened 2016-12-08
Ensure that all strings are extracted.

ivannaranjo:
This PR adds scripts that will scan the main code on the extension looking for patterns of strings that are not extracted out to `Resources.resx`. This script will be invoked from the `appveryor.yaml` build script and will make the build fail, ensuring that all strings are extracted before a PR can make it through.

The script found already some strings that were not extracted, this PR includes the fixes for those so the build stays green.

### DIFF
--- a/tools/find_files.py
+++ b/tools/find_files.py
@@ -1,0 +1,43 @@
+#! /usr/bin/env python
+"""Script that lists all files that match the given extension
+
+This script takes a directory and will list all of the files
+recursively filtering by the given extension.
+
+"""
+
+from __future__ import print_function
+
+import argparse
+import os
+import sys
+import msvcrt
+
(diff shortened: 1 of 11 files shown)


tools/find_files.py
```diff
@@ -0,0 +1,43 @@
...
+parser.add_argument('-d', '--directory',
+                    help='The root directory',
+                    required=True)
+parser.add_argument('-e', '--extension',
+                    help='The extension to filter by.',
+                    required=True)
+
+
+def print_all_files(dir):
+    for root, dirnames, filenames in os.walk(dir):
```
Deren-Liao: nit, dirnames not used,  you can do 
     for root, _, filenames in os.walk(dir):

PR #326, opened 2016-12-09
Check for unused strings

ivannaranjo:
This PR adds scripts that will check all of the strings referenced in the code against the strings defined in the resources, flagging those that are not used. Because unused strings are a waste having them will cause the build to fail in appveyor so we know when submitting PRs if the code is clean.

To keep the code building this PR also cleans up all of the unused strings that are found.

### DIFF
--- a/tools/find_unused_strings.py
+++ b/tools/find_unused_strings.py
@@ -50,7 +50,7 @@ def find_string(line, idx):
     return line[begin:end]
 
 
-def find_strings(file):
+def get_used_strings(file):
     result = set()
     with open(file, 'r') as src:
         for line in src.readlines():
@@ -75,7 +75,7 @@ def main(params):
     strings = load_strings(params.strings)
     files = list_all_files(params.directory)
     for file in files:
-        used = find_strings(file)
+        used = get_used_strings(file)
         strings = strings - used
     if len(strings) > 0:
         print("Strings that are not used:")
--- a/tools/find_unused_strings.py
+++ b/tools/find_unused_strings.py
@@ -1,0 +1,74 @@
+#! /usr/bin/env python
+"""Script that lists all files that match the given extension
+
+This script takes a directory and will list all of the files
+recursively filtering by the given extension.
+
+"""
+
+import argparse
+import os
+import sys
+
+# Parse for the program.
+parser = argparse.ArgumentParser()
+parser.add_argument('-d', '--directory',
+                    help='The root directory',
+                    required=True)
+parser.add_argument('-s', '--strings',
+                    help='The strings (.resx) file to check.',
+                    required=True)
+
+string_prefix="Resources."
+
+def list_all_files(dir):
+    result = []
+    for root, _, filenames in os.walk(dir):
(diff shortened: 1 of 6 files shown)


tools/find_unused_strings.py
```diff
@@ -0,0 +1,112 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument('-d', '--directory',
+                    help='The root directory',
+                    required=True)
+parser.add_argument('-s', '--strings',
+                    help='The strings (.resx) file to check.',
+                    required=True)
+
+
+string_prefix="Resources."
```
Deren-Liao: Since this is a hard coded string prefix, using name like resource_namespace  makes it easier to read. 
Plus,  Use  single quote to make it consistent cross the entire file.

tools/find_unused_strings.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def list_all_files(dir):
+    """Lists all of the source files (.cs and .xaml) under the given directory"""
+
+    result = []
+    for root, _, filenames in os.walk(dir):
+        for name in filenames:
+            filename, ext = os.path.splitext(name)
+            if ext == ".cs" or ext == ".xaml":
```
Deren-Liao: I did not notice..   Please use single quote for the entire file to make it consistent

tools/find_unused_strings.py
```diff
@@ -0,0 +1,112 @@
...
+                result.append(os.path.join(root, name))
+    return result
+
+
+def is_valid_char(src):
```
Deren-Liao: can it be isalnum() ?

tools/find_unused_strings.py
```diff
@@ -0,0 +1,112 @@
...
+    return result
+
+
+def load_strings(src):
+    """Loads the .resx file and extracts all of the string names."""
+
+    tree = ET.parse(src)
+    root = tree.getroot()
+    result = set()
+    for child in root.findall("data"):
```
Deren-Liao: Use single quote

tools/find_unused_strings.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def main(params):
+    strings = load_strings(params.strings)
+    files = list_all_files(params.directory)
+    for file in files:
+        used = get_used_strings(file)
+        strings = strings - used
+    if len(strings) > 0:
+        print("Strings that are not used:")
```
Deren-Liao: Use single quote

tools/find_unused_strings.py
ivannaranjo: Done.

tools/find_unused_strings.py
ivannaranjo: Done.

tools/find_unused_strings.py
ivannaranjo: Done.

tools/find_unused_strings.py
ivannaranjo: Done.

PR #330, opened 2016-12-13
Updating the major version for the branch.

ivannaranjo:
This PR updates the major version component of the extension in preparation for merging q1_master into master to continue working on the extension after the GA release. The next public release of the extension will have the 1.1.x.0 version format.

If we have a hotfix to be released between now and when we release 1.1.x.0 then that hotfix will be created from the v1.0.3.0 tag and will be versioned 1.0.x.0 (where x > 3).


ivannaranjo:
Thank you for the speedy review :).

PR #340, opened 2016-12-15
Show log entry detail as tree view 

Deren-Liao:


GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -42,6 +43,7 @@
 
             <Style x:Key="rowStyle" TargetType="DataGridRow">
                 <Setter Property="TextBlock.TextWrapping" Value="NoWrap" />
+                <Setter Property="TextBlock.FontSize" Value="12" />
```
ivannaranjo: Nit, should try to use the common styles

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -57,6 +59,34 @@
...
+            <vct:LogItemConverter x:Key="LogItemConverter"/>
+
+            <Style x:Key="TreeNodeText" 
+                   TargetType="TextBlock" 
+                   BasedOn="{StaticResource CommonTextStyle}" >
+                <Setter Property="Padding" Value="3, 2, 3, 2"/>
+                <Setter Property="Foreground" Value="DarkSlateGray"/>
+            </Style>
+
+            <Style x:Key="TreeNodeLabel"
```
ivannaranjo: Nit, you could move the styles into a standalone `ResourceDictionary` and load it in here to simplify the xaml file a bit. I want to do the same thing in the cloud explorer.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -57,6 +59,34 @@
                 </Style.Triggers>
             </Style>
 
+            <!-- Region TreeView styles and converters. -->
+            <vct:LogItemConverter x:Key="LogItemConverter"/>
+
+            <Style x:Key="TreeNodeText"
```
ivannaranjo: Nit, you should call all of the styles with the `Style` suffix, makes the code easier to read.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -57,6 +59,34 @@
...
+
+            <Style x:Key="TreeNodeLabel" 
+                   TargetType="TextBlock" 
+                   BasedOn="{StaticResource TreeNodeText}">
+                <Setter Property="Margin" Value="0, 0, 5, 0" />
+            </Style>
+
+            <HierarchicalDataTemplate DataType="{x:Type vct:ObjectNodeTree}" 
+                                      ItemsSource="{Binding Children}">
+                <StackPanel Orientation="Horizontal" >
```
ivannaranjo: Should consider doing this formatting on the model instead since this might need to be localized at some point. Also it is better to do it there so you can create the string you need instead of using textblocks for this.

Alternatively you can use a `TextBlock` element with children `TextBlock` for formatting. No need to use a `StackPanel`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml.cs
```diff
@@ -37,6 +39,22 @@ public LogsViewerToolWindowControl()
...
+        /// <param name="dependencyObj">A <seealso cref="DependencyObject"/> element. </param>
+        /// <returns>null or TControl object.</returns>
+        private TControl FindAncestorControl<TControl>(DependencyObject dependencyObj) where TControl : Control
+        {
+            while ((dependencyObj != null) && !(dependencyObj is TControl))
+            {
+                dependencyObj = VisualTreeHelper.GetParent(dependencyObj);
+            }
+
+            return dependencyObj as TControl;  // Note, null as Class is val
```
ivannaranjo: Fix the comment.

ivannaranjo: Also this code seems like it could be a utils class instead of here.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/LogItemConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace GoogleCloudExtension.StackdriverLogsViewer.TreeViewConverters
+{
+    /// <summary>
+    /// Convert LogItem object to a collection of <seealso cref="ObjectNodeTree"/> that treeview can display.
+    /// </summary>
+    internal sealed class LogItemConverter : IValueConverter
```
ivannaranjo: This is not really the right approach, the conversion should happen in the view model itself. Value converters are to be used to actually convert values, like converting a string to integer, boolean to visibility that sort of thing. What you are doing here is actually what the view model does, adapt the model to what the view needs. And it is better expressed that way. Please move this code to the view model and bind directly to the property.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/LogItemConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <param name="targetType">Ignored. The type of the binding target property.</param>
+        /// <param name="parameter">Ignored. The converter parameter to use. .</param>
+        /// <param name="culture">Ignored. The culture to use in the converter.</param>
+        /// <returns></returns>
+
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (!(value is LogItem))
+            {
+                throw new NotSupportedException("GetType().Name can only converte value type of LogItem");
```
ivannaranjo: Don't use a converter, but just for completeness converters are not supposed to throw, if you can't convert the value you are supposed to return `DependencyProperty.UnsetValue`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// Gets the DisplayValue
+        /// Tree node displays label in format of Name : DisplayValue.  
+        /// </summary>
+        public string NodeValue { get; private set; }
+
+        /// <summary>
+        /// Gets the obj visibility. 
+        /// Do not display ":" if the NodeValue is empty
+        /// </summary>
+        public Visibility ValueVisibility =>
```
ivannaranjo: Instead of this you can just return the string fully formatted and return the right string depending on whether there's a value or not. No need to use binding to show/hide the ":".

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// <summary>
+        /// Gets the object name.
+        /// </summary>
+        public string Name { get; private set; }
+
+        /// <summary>
+        /// Gets tree node children.
+        /// It contains all properties of the node object.
+        /// </summary>
+        public object Children { get; private set; }
```
ivannaranjo: Shouldn't this be a collection for some sort?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+            var collection = new ObservableCollection<object>();
+            Children = collection;
+            return collection;
+        }
+
+        private void ParseEnumerable(object obj)
+        {
+            var collection = CreateCollectionChildren();
+            IEnumerable arr = obj as IEnumerable;
+            Debug.Assert(arr != null);
```
ivannaranjo: This assert is not helping, if the data coming down from the server is bad then it is bad, it is not a bug in the VS code and there's nothing we can do to fix it. Asserts should only be used for situations that are truly bugs in the code.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+            if (arr == null)
+            {
+                // Don't expect null. Be protective
+                NodeValue = obj.ToString();
```
ivannaranjo: Shouldn't we just use a hardcoded string in this case then? We don't know what obj is and the `ToString()` call can return something that is not very useful to the user.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+                return;
+            }
+
+            int i = 0;
+            foreach (var ele in arr)
+            {
+                collection.Add(new ObjectNodeTree($"[{i}]", ele));
```
ivannaranjo: This formatting string is something the user sees, it should be in the resources.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+                ++i;
+            }
+
+            NodeValue = $"[{i}]";
```
ivannaranjo: Same...

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        private void ParseDictionary(object obj)
+        {
+            var collection = CreateCollectionChildren();
+            var dict = obj as IDictionary;
+            Debug.Assert(dict != null);
```
ivannaranjo: Same thing as above, if `obj` is not what you expect you should just show an error message and bail out. Don't try to fix it, the data coming down for the server is bad.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                // The best effort.
+                NodeValue = obj.ToString();
+            }
+        }
+
+        #region parser
+        private ObservableCollection<object> CreateCollectionChildren()
+        {
+            var collection = new ObservableCollection<object>();
```
ivannaranjo: Instead of this method, you should use a `Lazy<ObservableCollection<object>>` member, that way the collection will be created the very first time you access it. This way you don't need to implement the lazy pattern yourself.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+                    {
+                        collection.Add(new ObjectNodeTree(p.Name, v));
+                    }
+                }
+                catch (Exception ex)
+                {
+                    if (ex is ArgumentException || ex is TargetException || ex is TargetParameterCountException
+                        || ex is MethodAccessException || ex is TargetInvocationException)
+                    {
+                        // Value convertion error, display for awarness.
```
ivannaranjo: Typo in comment, not sure that this actually helps the user in any way. You should just use a generic error message that describes the fact that the server is sending bad data or it is down.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TypeUtil.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            return type.IsGenericType && type.GetGenericTypeDefinition().IsAssignableFrom(typeof(Dictionary<,>));
+        }
+
+        /// <summary>
+        /// Check if the object is a IList type
+        /// </summary>
+        public static bool IsListObject(this object obj)
+        {
+            return obj != null && obj is IList && obj.GetType().IsListType();
```
ivannaranjo: This seems redundant, the `obj is IList` bit will already return if the object is a list, no need to check the `Type`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
ivannaranjo: You can type this as `IEnumerable` at the very least.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// Create an instance of the <seealso cref="ObjectNodeTree"/> class.
+        /// </summary>
+        /// <param name="name">object name</param>
+        /// <param name="obj">An object</param>
+        private ObjectNodeTree(string name, object obj)
+        {
+            Name = name;
+            if (obj == null)
+            {
+                NodeValue = null;
```
ivannaranjo: `NodeValue` will be `null` by default, you only need to set the value if `obj` is not `null`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// Gets the obj visibility. 
+        /// Do not display ":" if the NodeValue is empty
+        /// </summary>
+        public Visibility ValueVisibility => 
+            String.IsNullOrWhiteSpace(NodeValue) ? Visibility.Hidden : Visibility.Visible;
+
+        /// <summary>
+        /// Gets the object name.
+        /// </summary>
+        public string Name { get; private set; }
```
ivannaranjo: Since these properties are not changed after the object is initialized you don't even need to declare the setter, just declare them with a getter as:
```C#
public string Name { get; }
```
This way the properties will not be mutable, which is what you want I think.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+                ParseDictionary(obj);
+            }
+            else if (s_supportedTypes.Contains(type))
+            {
+                ParseClassProperties(obj, type);
+            }
+            else
+            {
+                // The best effort.
+                NodeValue = obj.ToString();
```
ivannaranjo: I think that you should hide the node if the type is unknown, don't bother showing data that you don't know what it is.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TypeUtil.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// Utility methods primarily used by ObjectNodeTree 
+    /// </summary>
+    public static class TypeUtil
+    {
+        /// <summary>
+        /// Check if the type is a IList generic type.
+        /// </summary>
+        public static bool IsListType(this Type type)
+        {
+            return type.IsGenericType && type.GetGenericTypeDefinition().IsAssignableFrom(typeof(List<>));
```
ivannaranjo: `List<T>` will implement `IList<T>` and `IList` so you really only need to check if the object implements `IList`. Same with `IDictionary`, so these methods are redundant.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
ivannaranjo: Looking at the code it seems that you don't actually need an `ObservableCollection<T>` since the contents of the collection never actually change. A regular `List<object>` would do fine as well.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        /// <summary>
+        /// Parsing the object properties recursively as a tree.
+        /// </summary>
+        private void ParseObjectTree(object obj)
+        {
+            Type type = obj.GetType();
+
+            if (obj.IsNumericType() || obj is string || obj is DateTime)
```
ivannaranjo: For `DateTime` and numbers you might need to format using a localized string of some sort so the strings and numbers show correctly in the various locales. This means that `ToString()` is not going to be sufficient.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                NodeValue = obj.ToString();
+            }
+            else if (type.IsArray)
```
ivannaranjo: You could just do `obj is IEnumerable` instead.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        #region parser
+        private ObservableCollection<object> CreateCollectionChildren()
+        {
+            var collection = new ObservableCollection<object>();
+            Children = collection;
+            return collection;
+        }
+
+        private void ParseEnumerable(object obj)
```
ivannaranjo: You should accept an `IEnumerable` as the parameter here. The caller would do the right thing.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+            foreach (var ele in arr)
+            {
+                collection.Add(new ObjectNodeTree($"[{i}]", ele));
+                ++i;
+            }
+
+            NodeValue = $"[{i}]";
+        }
+
+        private void ParseDictionary(object obj)
```
ivannaranjo: You should accept an `IDictionary` here, in general try to be as typed as possible, let the compiler help.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            foreach (var key in dict.Keys)
+            {
+                string name = key.ToString();
+                collection.Add(new ObjectNodeTree(name, dict[key]));
+            }
+        }
+
+        private void ParseClassProperties(object obj, Type type)
```
ivannaranjo: While this makes sense for debugging you should probably use localized names for the properties, not all properties might be shown, etc... This means that you should probably have a sub-class for each of the supported types and create the right sub-tree for each supported type. This is what we do in the cloud explorer, you can look in there for patterns to use.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -263,20 +293,24 @@
...
-                                    <Setter Property="Foreground" Value="DarkSlateGray" />
-                                </Style>
-                            </TreeView.ItemContainerStyle>
-                        </TreeView>
+                        <!-- Use a border to make the background right. -->
+                        <Border Background="White" Padding="30, 0, 0 , 2">
+                            
+                            <!-- Override the template to disable inner scrollbar -->
+                            <TreeView BorderThickness="0" 
+                                ItemsSource="{Binding Converter={StaticResource LogItemConverter}}">
```
ivannaranjo: This is where your view model should have a property that returns the appropriate `ItemsSource`, don't use a converter for this.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: I removed the font setting.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: Renamed.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: This is a better way.  Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: The NodeValue will need to be formatted differently.  I'll combine the Name and ":" .

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/LogItemConverter.cs
Deren-Liao: Removed.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: As I explained, some spaces are needed.  Don't want to use space in formatting string.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: The code used to do private set, with the last change it is set only in constructor.  
Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: Changed to IList

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: Makes sense.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: The purpose was to leave Children as null if not needed.  After I change Children type to IList  I don't need this helper function.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: The purpose was not to protect unknown data from server. It is that supposedly, arr should not be null after this conversion. However, if it turned out to be null, it is a code bug. The Assert here is to help find the bug if any.

Deren-Liao: I removed it if this confuses people.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: This should not be null.  I removed the section.  An array is always Enumerable.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: Added to resources.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: After IsDictionaryObject() test,  it should not be null. I remove the null test section.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: using ParseClassProperties can be flexible. If in the future, more fields are added, we don't have to come back to modify the code to display them.  
Also, there is no spec or hard requirement about displaying which property or not to.  I compared the current results to Pantheon results with many different log entries,  they do match. 
Plus, creating sub-class for each type would take a lot of time. I created issue #350 so that I'll revisit it later.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TreeObjectNode.cs
Deren-Liao: I added a generic error.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TypeUtil.cs
Deren-Liao: Yes, I don't know why I did this in first place.  I removed them.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/TypeUtil.cs
Deren-Liao: Removed.

Deren-Liao:
Please take a look again.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1371,4 +1371,12 @@
     <value>Loading . . .</value>
     <comment>List log entries can take seconds to complete, show a status message.</comment>
   </data>
+  <data name="LogsViewerDataConversionGenericError" xml:space="preserve">
+    <value>Data conversion error. Log entry may contain invalid properties.</value>
```
ivannaranjo: Nit, I would make this error a bit more generic, something along the lines of "Error displaying log entries", the user doesn't really know anything about what we're doing to display the entry so the "conversion" doesn't really make sense from that perspective.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1371,4 +1371,12 @@
...
+    <comment>Showing generic error for parsing LogEntry property tree.</comment>
+  </data>
+  <data name="LogsViewerDetailTreeViewArrayIndexFormat" xml:space="preserve">
+    <value>[{0}]</value>
+    <comment>The detail tree view array index format.</comment>
```
ivannaranjo: Nit, add an indication of what the `{0}` means for possible future translators.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
```diff
@@ -0,0 +1,64 @@
...
+    </ResourceDictionary.MergedDictionaries>
+
+    <!-- DataGrid cell style. 
+        Note: Add padding to text, padding at row style does not work. -->
+    <Style x:Key="cellStyle" TargetType="DataGridCell">
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="DataGridCell">
+                    <Border  BorderThickness="0">
+                        <Border x:Name="border" Padding="0, 4, 0, 4" >
```
ivannaranjo: Nit, no need for spaces in the `Padding`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
```diff
@@ -0,0 +1,64 @@
...
+                            <ContentPresenter />
+                        </Border>
+                    </Border>
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>
+    </Style>
+
+    <Style x:Key="rowStyle" TargetType="DataGridRow">
+        <Setter Property="Margin" Value="0, 0, 0, 1"/>
```
ivannaranjo: Nit, same with the formatting of the margins, etc... no spaces after ",".

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
```diff
@@ -0,0 +1,64 @@
...
+            </Trigger>
+            <Trigger Property="IsMouseOver" Value="True">
+                <Setter Property="Background" Value="#FFFFD6" />
+            </Trigger>
+        </Style.Triggers>
+    </Style>
+
+    <!--Region Log entry detail treeView styles. -->
+    <Style x:Key="TreeNodeTextStyle" 
+                   TargetType="TextBlock"
```
ivannaranjo: Nit, alignment of the attributes, they should all align with the "x:Key" attribute.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
```diff
@@ -0,0 +1,64 @@
...
+                <Setter Property="Background" Value="#FFFFD6" />
+            </Trigger>
+        </Style.Triggers>
+    </Style>
+
+    <!--Region Log entry detail treeView styles. -->
+    <Style x:Key="TreeNodeTextStyle" 
+                   TargetType="TextBlock" 
+                   BasedOn="{StaticResource CommonTextStyle}" >
+        <Setter Property="Padding" Value="3, 2, 3, 2"/>
```
ivannaranjo: Shouldn't this be `Margin` though? as you want space around the `TextBlock` not inside of it?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
```diff
@@ -0,0 +1,64 @@
...
+        <Setter Property="Foreground" Value="DarkSlateGray"/>
```
ivannaranjo: Not for this PR but at some point we shoud review the colors and use the full hex name for them. We should run these colors through the UI designers too to make sure they make sense.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -21,78 +16,22 @@
...
                                       ItemsSource="{Binding Children}">
-                <StackPanel Orientation="Horizontal" >
-                    <TextBlock Text="{Binding Name}" Style="{StaticResource TreeNodeLabel}"/>
-                    <TextBlock Text=":" Style="{StaticResource TreeNodeLabel}" 
-                               Visibility="{Binding ValueVisibility}"/>
-                    <TextBlock Text="{Binding NodeValue}" Style="{StaticResource TreeNodeText}"
-                               Visibility="{Binding ValueVisibility}"/>
+                <StackPanel Orientation="Horizontal">
+                    <TextBlock Text="{Binding Name}" Style="{StaticResource TreeNodeTextStyle}" />
+                    <TextBlock Text=":"
```
ivannaranjo: This formatting should be done by the view model, that is, you should produce a string that contains the "{Name} : {NodeValue}" or not. Don't use Xaml for string formatting like this.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -53,22 +55,22 @@ internal class ObjectNodeTree
         public string NodeValue { get; private set; }
 
         /// <summary>
-        /// Gets the obj visibility. 
+        /// Gets the Colon visibility. 
         /// Do not display ":" if the NodeValue is empty
         /// </summary>
-        public Visibility ValueVisibility => 
-            String.IsNullOrWhiteSpace(NodeValue) ? Visibility.Hidden : Visibility.Visible;
+        public Visibility ColonVisibility =>
```
ivannaranjo: As mentioned above, don't show/hide the colon. Have a property that returns the fully formatted string, the value of the property can change depending on whether the ":" must be shown or not. This is a one time thing (it seems) so in practice you will calculate this string only once.

The way you are doing it right now is hiding a formatting string in Xaml, which will make it difficult later to review the strings shown to the user. Using visibility to show/hide UI elements is really only when you are going to show/hide buttons, cells, etc... not parts of a string.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -101,83 +102,60 @@ private void ParseObjectTree(object obj)
...
-            if (arr == null)
-            {
-                // Don't expect null. Be protective
-                NodeValue = obj.ToString();
-                return;
-            }
-
+            Children = new List<ObjectNodeTree>();
             int i = 0;
             foreach (var ele in arr)
```
ivannaranjo: Random nit (mostly my personal taste), prefer to spell the full names in the variables. `obj` is fine but `ele` and `arr` could be fully spelled.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -101,83 +102,60 @@ private void ParseObjectTree(object obj)
...
             {
-                collection.Add(new ObjectNodeTree($"[{i}]", ele));
+                Children.Add(
+                    new ObjectNodeTree(String.Format(Resources.LogsViewerDetailTreeViewArrayIndexFormat, i), ele));
                 ++i;
             }
 
-            NodeValue = $"[{i}]";
+            NodeValue = String.Format(Resources.LogsViewerDetailTreeViewArrayIndexFormat, i);
```
ivannaranjo: Thank you :)

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -101,83 +102,60 @@ private void ParseObjectTree(object obj)
...
                 string name = key.ToString();
-                collection.Add(new ObjectNodeTree(name, dict[key]));
+                Children.Add(new ObjectNodeTree(name, dict[key]));
             }
         }
 
         private void ParseClassProperties(object obj, Type type)
         {
-            var collection = CreateCollectionChildren();
+            Children = new List<ObjectNodeTree>();
```
ivannaranjo: You could use a `Lazy<List<ObjectNodeTree>>` value for Children and then make it so you don't have to actually have the `new List<objectNodeTree>()` all over the parser code, instead just access the value of the Children property and create the list there lazily.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -191,17 +169,17 @@ private void ParseClassProperties(object obj, Type type)
                     object v = p.GetValue(obj);
                     if (v != null)
                     {
-                        collection.Add(new ObjectNodeTree(p.Name, v));
+                        Children.Add(new ObjectNodeTree(p.Name, v));
                     }
                 }
                 catch (Exception ex)
```
ivannaranjo: Nit, you could do this with an exception filter see [this](http://geekswithblogs.net/BlackRabbitCoder/archive/2015/04/09/c.net-little-wonders-exception-filtering-in-c-6.aspx)

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
Deren-Liao: I don't seem to see difference.   Margin is good too.  
Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerResources.xaml
Deren-Liao: #352

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: I can combine {Name} :  
NodeValue will need different formatting to make it looks like a 'link' so as to do advanced filtering recommendation.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: If I use lazy,  when the xaml calls Get,  Children will create an empty list.  My intention is to keep it null instead of creating an empty list since this is the case for most of the instances.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: As mentioned above,  NodeValue needs very different format later. I am combining Name and colon as NameLabel.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: This is very useful enhancement !  Great to learn the new Technics.

Deren-Liao:
I made changes on the feedback. 
The only one comment I kept wast to separate NodeValue with different format.  Later, I'll change it to a button link to do advanced filtering recommendation.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1372,11 +1372,15 @@
     <comment>List log entries can take seconds to complete, show a status message.</comment>
   </data>
   <data name="LogsViewerDataConversionGenericError" xml:space="preserve">
-    <value>Data conversion error. Log entry may contain invalid properties.</value>
+    <value>Error displaying the entry. Log entry may contain invalid properties.</value>
```
ivannaranjo: Nit, leave the "log entry may contain invalid properties." out of the error message as the user cannot do anything about it.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -24,10 +24,7 @@
             <HierarchicalDataTemplate DataType="{x:Type vct:ObjectNodeTree}" 
                                       ItemsSource="{Binding Children}">
                 <StackPanel Orientation="Horizontal">
```
ivannaranjo: I'm convinced that you will be better off putting both `TextBlock` inside of a `TextBlock` (which is legal) to avoid having a `StackPanel` perform the layout of the text, which will be picky to get precisely right. This will be specially important later on if you want to make the `NodeValue` a `Hyperlink`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -49,34 +49,38 @@ internal class ObjectNodeTree
...
-        public Visibility ColonVisibility =>
-            String.IsNullOrWhiteSpace(NodeValue) ? Visibility.Collapsed : Visibility.Visible;
+        public string NodeValue { get; private set; }
 
         /// <summary>
-        /// Gets the object name.
+        /// Gets the object name with optional colon.
         /// </summary>
-        public string Name { get; }
+        public string NameLabel => String.IsNullOrWhiteSpace(NodeValue) ? _name :
```
ivannaranjo: Nit, this could still be called `Name`, this is a view model and the view model will contain calculated properties to adapt the data model to the view.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -129,33 +133,36 @@ private void ParseObjectTree(object obj)
         }
 
         #region parser
-        private void ParseArray(IEnumerable arr)
+        private void ParseArray(IEnumerable arrayObject)
```
ivannaranjo: Nit, name is too specific. The argument is of type `IEnumerable` not array, use a more neutral name for the argument.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -129,33 +133,36 @@ private void ParseObjectTree(object obj)
...
 
             NodeValue = String.Format(Resources.LogsViewerDetailTreeViewArrayIndexFormat, i);
         }
 
-        private void ParseDictionary(IDictionary dict)
+        private void ParseDictionary(IDictionary dictionaryObject)
         {
-            Children = new List<ObjectNodeTree>();
-            foreach (var key in dict.Keys)
+            foreach (var key in dictionaryObject.Keys)
```
ivannaranjo: Nit, not required for this PR but I would recommend exploring the use of the LINQ methods to perform this sort of operations on collections and `IEnumerable` instances. You will find that it simplifies the code a lot.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
```diff
@@ -129,33 +133,36 @@ private void ParseObjectTree(object obj)
...
             {
                 string name = key.ToString();
-                Children.Add(new ObjectNodeTree(name, dict[key]));
+                Children.Add(new ObjectNodeTree(name, dictionaryObject[key]));
             }
         }
 
+        private bool IsPropertyInfoGetValueException(Exception ex)
```
ivannaranjo: Nit, this method could be `static`.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: I changed it to enumerable.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: LINQ is great.  As I did a quick try,  seems that IDictionary does not expose IEnumerable that LINQ does not apply.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs
Deren-Liao: Done

PR #354, opened 2017-01-09
Add Logging and Error Reporting Project Template

jsimonweb:


iantalarico:
I think we probably just want to have one template that contains everything.  This way users get it all from the start.

ivannaranjo:
I'm closing this PR since we're not going to submit it as is.

PR #373, opened 2017-01-23
Adding error handling to the Flex deployment step.

ivannaranjo:
This PR adds better error handling to the Flex and GCE deployment steps. All non-critical exceptions produced during the Flex deployment will be reported to the user as an error message in the output window in Visual Studio.

This  change also changes the way the step UI is enabled in the GCE deployment step when all VMs are loaded by using the new `TrackTask` method in the publish dialog.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/FlexStep/FlexStepViewModel.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/FlexStep/FlexStepViewModel.cs
@@ -79,53 +79,61 @@ namespace GoogleCloudExtension.PublishDialogSteps.FlexStep
 
         public override async void Publish()
         {
-            var context = new GCloudContext
-            {
-                CredentialsPath = CredentialsStore.Default.CurrentAccountPath,
-                ProjectId = CredentialsStore.Default.CurrentProjectId,
-                AppName = GoogleCloudExtensionPackage.ApplicationName,
-                AppVersion = GoogleCloudExtensionPackage.ApplicationVersion,
-            };
-            var options = new AppEngineFlexDeployment.DeploymentOptions
-            {
-                Version = Version,
-                Promote = Promote,
-                Context = context
-            };
             var project = _publishDialog.Project;
-
-            GcpOutputWindow.Activate();
-            GcpOutputWindow.Clear();
-            GcpOutputWindow.OutputLine(String.Format(Resources.GcePublishStepStartMessage, project.Name));
-
-            _publishDialog.FinishFlow();
-
-            AppEngineFlexDeploymentResult result;
-            using (var frozen = StatusbarHelper.Freeze())
-            using (var animationShown = StatusbarHelper.ShowDeployAnimation())
-            using (var progress = StatusbarHelper.ShowProgressBar(Resources.FlexPublishProgressMessage))
-            using (var deployingOperation = ShellUtils.SetShellUIBusy())
+            try
             {
-                result = await AppEngineFlexDeployment.PublishProjectAsync(
-                    project.FullPath,
-                    options,
-                    progress,
-                    GcpOutputWindow.OutputLine);
-            }
-
-            if (result != null)
-            {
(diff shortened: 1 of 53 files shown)


GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/FlexStep/FlexStepViewModel.cs
```diff
@@ -79,53 +79,61 @@ public override void OnPushedToDialog(IPublishDialog dialog)
...
-                    Process.Start(url);
+                    GcpOutputWindow.OutputLine(String.Format(Resources.FlexPublishFailedMessage, project.Name));
+                    StatusbarHelper.SetText(Resources.PublishFailureStatusMessage);
                 }
             }
-            else
+            catch (Exception ex) when (!ErrorHandlerUtils.IsCriticalException(ex))
             {
                 GcpOutputWindow.OutputLine(String.Format(Resources.FlexPublishFailedMessage, project.Name));
                 StatusbarHelper.SetText(Resources.PublishFailureStatusMessage);
```
Deren-Liao: Can  you popup an error message box ?  When I used the publishing for a few times,  without seeing "complete" or "error" message box, I often feel unclear about "where am I now? Is it done or not".

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GceStep/GceStepViewModel.cs
```diff
@@ -136,39 +136,47 @@ private void OnManageCredentialsCommand()
...
+                else
+                {
+                    GcpOutputWindow.OutputLine(String.Format(Resources.GcePublishFailedMessage, project.Name));
+                    StatusbarHelper.SetText(Resources.PublishFailureStatusMessage);
                 }
             }
-            else
+            catch (Exception ex) when (!ErrorHandlerUtils.IsCriticalException(ex))
             {
                 GcpOutputWindow.OutputLine(String.Format(Resources.GcePublishFailedMessage, project.Name));
```
Deren-Liao: Same ask as in FlexStepViewModel.  I am not asking to add now if you agree with the idea.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GceStep/GceStepViewModel.cs
```diff
@@ -178,6 +186,8 @@ private void OnManageCredentialsCommand()
         public override void OnPushedToDialog(IPublishDialog dialog)
         {
             _publishDialog = dialog;
+
+            _publishDialog.TrackTask(Instances.ValueTask);
```
Deren-Liao: It's not obvious to me how you handle the failure of retrieving Instances.   Inside IPublishDialog.TrackTask(Task task),  you simply swallow them.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/FlexStep/FlexStepViewModel.cs
ivannaranjo: The problem with popping an error dialog is that it will show up without context to the user. It is best to show the error in the output window as part of the normal flow of publishing. This is for the case when an error happens after the process has started.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GceStep/GceStepViewModel.cs
ivannaranjo: Same explanation, lack of context for the dialog to show up.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GceStep/GceStepViewModel.cs
ivannaranjo: Nothing. `IPublishDialog.TrackTask()` is only interested in the completion state of the task not on its result. The `AsyncPropertyValue<T>` instance will transition into the error state if the task fails.

That brings a good point though, the UI should show an error state if we fail to load the data. Will work on that.

PR #375, opened 2017-01-24
Adding functionality to specify number of replicas.

ivannaranjo:
This PR adds functionality to specify the number of replicas for the new deployment. If a deployment with the same name already exists then the user is warned about this and if the user decides to proceed the deployment will be scaled up/down as necessary to match the requested number of replicas.

There's a bit of refactoring to allow the deployment code to query to see if the deployment already exists and warn the user.

The deployment step now will look like this:
screen shot 2017-01-24 at 3 55 33 pm


GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlContext.cs
```diff
@@ -12,16 +12,53 @@
...
+    /// This class owns the context on which to run kubectl commands.
     /// </summary>
-    public class KubectlContext
+    public class KubectlContext : IDisposable
     {
         /// <summary>
         /// Path to the config file that identifies the cluster for kubcectl commands.
         /// </summary>
-        public string Config { get; set; }
+        public string Config { get; private set; }
```
Deren-Liao: Nit,  rename to ConfigFilePath etc.

GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlContext.cs
```diff
@@ -12,16 +12,53 @@
...
+                return;
+            }
+
+            try
+            {
+                File.Delete(Config);
+            }
+            catch (IOException ex)
+            {
+                Debug.WriteLine($"Failed to delete {Config}: {ex.Message}");
```
Deren-Liao: What's inside the config file? Does it contain any PII ?

GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlContext.cs
```diff
@@ -12,16 +12,53 @@
...
+        void IDisposable.Dispose()
+        {
+            if (Config == null)
+            {
+                return;
+            }
+
+            try
+            {
+                File.Delete(Config);
```
Deren-Liao: Kind of odd you delete the file that is not created by the itself.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepContent.xaml
```diff
@@ -36,13 +36,38 @@
...
+
+            <Label Content="{x:Static ext:Resources.GkePublishDeploymentVersionMessage}"
+                   Target="{Binding ElementName=_deploymentVersion}"
+                   Style="{StaticResource CommonLabelStyle}" />
+            <TextBox x:Name="_deploymentVersion"
+                     Grid.Row="1"
+                     Text="{Binding DeploymentVersion, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
+                     Style="{StaticResource CommonTextBoxStyle}" />
+
+            <!-- TODO: Use an up/down textbox control. -->
```
Deren-Liao: Nit, Create an issue number and add it to the TODO comment.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepViewModel.cs
```diff
@@ -133,84 +148,130 @@ public override void OnPushedToDialog(IPublishDialog dialog)
...
                 StatusbarHelper.SetText(Resources.PublishFailureStatusMessage);
+                _publishDialog.FinishFlow();
             }
         }
 
         #endregion
 
+        private bool ValidateInput()
+        {
+            int replicas = 0;
```
Deren-Liao: Consider to use TextBox validation?

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepViewModel.cs
```diff
@@ -133,84 +148,130 @@ public override void OnPushedToDialog(IPublishDialog dialog)
...
+                    var options = new GkeDeployment.DeploymentOptions
+                    {
+                        Cluster = SelectedCluster.Name,
+                        Zone = SelectedCluster.Zone,
+                        DeploymentName = DeploymentName,
+                        DeploymentVersion = DeploymentVersion,
+                        ExposeService = ExposeService,
+                        GCloudContext = gcloudContext,
+                        KubectlContext = kubectlContext,
+                        Replicas = int.Parse(Replicas),
```
Deren-Liao: Define Replicas as int type may make this easier.

GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlContext.cs
ivannaranjo: The instance owns the file after the creation. The file is created by the caller, `GCloudWrapper` in this case and handed over to the `KubectlContext` instance for safekeeping.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepContent.xaml
ivannaranjo: It is not a good practice to add bug/issue numbers to comments though. This is issue #377

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepViewModel.cs
ivannaranjo: TextBox validation doesn't quite work well for our scenarios though. We use the same pattern all over the extension, validate when the user ask us to "commit" the action to ensure that all values are correct at that point, show error messages for those values that are not valid.

Validation as the user enters data seems like a good idea, but data being entered will always be wrong as it is being entered. Validation is better performed on commit.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepViewModel.cs
ivannaranjo: Not really, the TextBox value needs to be parsed later on when validating to ensure that it is a valid value. If we define as `int` then all we have is the value 0.

GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlContext.cs
ivannaranjo: Will use `ConfigPath`, I think is cleaner.

PR #378, opened 2017-01-25
Adding link to new cluster page, refresh of clusters.

ivannaranjo:
This PR adds a link to the GKE deployment step to the new cluster page in the Cloud console so the user can create a new cluster if required. A new link to refresh the list of clusters is also added. This PR also adds support for the case when no clusters are found in a project, previously the extension would crash.

The dialog step will now look like this when no clusters are found:
screen shot 2017-01-25 at 3 17 13 pm


GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepContent.xaml
```diff
@@ -20,14 +20,34 @@
...
                   Style="{StaticResource CommonComboBoxStyle}"/>
 
+        <StackPanel Margin="0,0,0,15" Orientation="Horizontal">
+
+            <TextBlock Margin="0,0,5,0"
+                       Style="{StaticResource CommonTextStyle}">
+                <Hyperlink Command="{Binding CreateClusterCommand}"
+                           Style="{StaticResource CommonHyperlinkStyle}">
+                    <TextBlock Text="{x:Static ext:Resources.GkePublishNewClusterCaption}" />
+                </Hyperlink>
```
Deren-Liao: Nit:  Add a tooltip saying something like. "Create cluster at Google Cloud Console"

Issue #367, opened 2017-01-19
Failed to load GoogleCloudExtension.Utils.dll inside xaml


Deren-Liao:
xaml references to GoogleCloudExtension.Utils as following. 
Only when the user control is opened inside VS extension it throws exception of failed to load the dll. 
The hack way to fix is to call a method of the dll before loading the xaml file. 
It has something to do with VSIX installation. 
Possibly it is required to put the dependencies into source.extension.vsixmanifest

--------
<UserControl x:Class="GoogleCloudExtension.StackdriverLogsViewer.LogsViewerToolWindowControl"
             xmlns:utils="clr-namespace:GoogleCloudExtension.Utils;assembly=GoogleCloudExtension.Utils"

            <utils:VisibilityConverter x:Key="visibilityConverter" />

--------

 public LogsViewerToolWindow() : base(null)
        {
            // disposable does nothing.
            // This is a hack to force load GoogleCloudExtension.Utils
            using (var disposable = new Disposable())   
            {
                Caption = Resources.LogViewerToolWindowCaption;
            }

ivannaranjo:
This is working for literally everywhere else where I am using these classes. If this is giving you trouble then the best thing we can do is to move these classes back into the main assembly.

Deren-Liao:
I guess it's because everywhere else, you have c# code that already
referenced the dll that it indirectly loaded the module into app domain,
especially if it has the Cloud Explorer opened as kind first step.  If the
module is not loaded, and an xaml file references it, and it is inside
Visual Studio Extension, it fails.

Logs Viewer can be opened from menu before Cloud Explorer is opened.  As I
mentioned, if Cloud Explorer is opened already, there is no problem.

The fix, I kind feel it could be to add the dll to dependencies in the
vsixmanifest file.  I tired to ad an item, but the format was wrong.   I'll
see if I can get the format correct and have a try later.

Deren-Liao:
Adding it to dependency make the installation fail.

Deren-Liao:
Please note,  this won't happen if it is non-VS extension xaml.

PR #380, opened 2017-01-26
Fix issue #367

Deren-Liao:
https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/issues/367


ivannaranjo:
I'm aproving this to unblock you, but please find the actual solution before merging into master.

Deren-Liao:
Thank you.

Deren-Liao:
Hi Ivan, 

Thanks for approving the pull request. 

Please note,  this won't happen if it is non-VS extension xaml.    

I have a repro at git https://github.com/Deren-Liao/tide.git .  Open the VSIX project.   
Start debug,  in the launched VS, click view -> other windows -> ToolWindow.  

This is a "hack" way, but I'd say this is one  way to fix the VS extension bug.   Unless we set a rule not to have tool window xaml reference third party assembly, we need something alike as workaround.

PR #392, opened 2017-02-02
Initial population of the m2 GCS browser branch.

ivannaranjo:
This PR moves the code from my private repo to the main repo now that the GCS browser code is in good enough shape to review.

My apologies for the rather large review, please do as you can to get through it. I will reply and apply all comments the day after as usual.

After this PR goes in all next work will happen in PRs agains the m2_gcs_browser branch in the main repo.


GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+            }
+        }
+
+        /// <summary>
+        /// Uploads the given <paramref name="stream"/> to the <paramref name="bucket"/> with the given <paramref name="name"/>.
+        /// </summary>
+        /// <param name="bucket">The bucket that will own the file.</param>
+        /// <param name="name">The name to use.</param>
+        /// <param name="stream">The stream with the contents.</param>
+        /// <param name="contentType">The c ontent type to use, optional.</param>
```
iantalarico: nit: extra space "c ontent"

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+            catch (GoogleApiException ex)
+            {
+                throw new DataSourceException(ex.Message, ex);
+            }
+        }
+
+        /// <summary>
+        /// Starts a file upload operation reporting the status and progress to the given <paramref name="operation"/>.
+        /// </summary>
+        /// <param name="sourcePath">The path to the file to open, should a full path.</param>
```
iantalarico: "should be a..."

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+        /// <param name="bucket">The bucket that will own the file.</param>
+        /// <param name="name">The name to use.</param>
+        /// <param name="operation">The operation that will receive the status and progress notifications.</param>
+        /// <param name="token">The cancellation token to cancel the operation.</param>
+        public async void StartFileUploadOperation(
```
iantalarico: This should return a Task

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+                        },
+                        bucket,
+                        stream,
+                        null);
+                    request.ProgressChanged += (p) => OnUploadProgress(p, totalSize, operation);
+                    var response = await request.UploadAsync(token);
+                    operation.Completed();
+                }
+            }
+            catch (GoogleApiException ex)
```
iantalarico: Do you need to catch file error?

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+        /// <param name="name">The name to use.</param>
+        /// <param name="operation">The operation that will receive the status and progress notifications.</param>
+        /// <param name="token">The cancellation token to cancel the operation.</param>
+        public async void StartFileUploadOperation(
+            string sourcePath,
+            string bucket,
+            string name,
+            IGcsFileOperation operation,
+            CancellationToken token)
+        {
```
iantalarico: Should content type be a parameter here too? Or can you pull it from the file?

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+
+        /// <summary>
+        /// Starts a file download operation, reporting the status and progress to the given <paramref name="operation"/>.
+        /// </summary>
+        /// <param name="bucket">The bucket that owns the file.</param>
+        /// <param name="name">The file name.</param>
+        /// <param name="destPath">Where to save the file, this should be a full path.</param>
+        /// <param name="operation">The operation that will receive the status and progress notifications.</param>
+        /// <param name="token">The cancellation token to cancel the operation.</param>
+        public async void StartFileDownloadOperation(
```
iantalarico: This should return a Task

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+                    var obj = await request.ExecuteAsync(token);
+                    ulong totalSize = obj.Size ?? 0;
+
+                    // Hookup the progress indicator.
+                    request.MediaDownloader.ProgressChanged += (p) => OnDownloadProgress(p, totalSize, operation);
+                    var response = await request.DownloadAsync(stream, token);
+                    operation.Completed();
+                }
+            }
+            catch (GoogleApiException ex)
```
iantalarico: Do you need to catch file error?

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -62,5 +67,225 @@ public Task<IList<Bucket>> GetBucketListAsync()
...
+        }
+
+        /// <summary>
+        /// Starts a delete operation, sending the notifications to the given <paramref name="operation"/>.
+        /// </summary>
+        /// <param name="bucket">The bucket that owns the file.</param>
+        /// <param name="name">The name of the file.</param>
+        /// <param name="operation">The operation that will recieve the status.</param>
+        /// <param name="token">The cancellation token to cancel the operation.</param>
+        public async void StartDeleteOperation(string bucket, string name, IGcsFileOperation operation, CancellationToken token)
```
iantalarico: This should return a Task

GoogleCloudExtension/GoogleCloudExtension.DataSources/IGcsFileOperation.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Called if the user cancelled the operation.
+        /// </summary>
+        void Cancelled();
+
+        /// <summary>
+        /// Called when the operation failed.
+        /// </summary>
+        /// <param name="ex"></param>
```
iantalarico: Empty param tag

GoogleCloudExtension/GoogleCloudExtension.Utils/ControlBehaviors.cs
```diff
@@ -0,0 +1,115 @@
...
+            DependencyProperty.RegisterAttached(
+                "DoubleClickCommandParameter",
+                typeof(object),
+                typeof(ControlBehaviors));
+
+        /// <summary>
+        /// The getter for the <seealso cref="DoubleClickCommandParameterProperty"/>.
+        /// </summary>
+        /// <param name="self">The <seealso cref="Control"/>.</param>
+        /// <returns></returns>
```
iantalarico: nit: empty returns

GoogleCloudExtension/GoogleCloudExtension.Utils/DataGridBehaviors.cs
```diff
@@ -0,0 +1,135 @@
...
+                self.Sorting += OnDataGridSorting;
+            }
+        }
+
+        #endregion
+
+        #region CustomSort property.
+
+        /// <summary>
+        /// This attached property is meant to applied to <seealso cref="DataGridColumn"/> instances, and contains
```
iantalarico: "meant to be applied to"

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gcs/BucketViewModel.cs
```diff
@@ -83,5 +88,10 @@ private void OnOpenConCloudConsoleCommand()
         }
 
         private BucketItem GetItem() => new BucketItem(_bucket);
+
+        public void OnDoubleClick()
+        {
+            GcsFileBrowserWindow.BrowseBucket(_bucket);
```
iantalarico: This is awesome :-)

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserState.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        /// <summary>
+        /// The list of individual items (or blobs) at this part on the path.
+        /// </summary>
+        public IList<GcsRow> Items { get; }
+
+        /// <summary>
+        /// The pathsteps to get this state.
+        /// </summary>
+        public IEnumerable<PathStep> PathSteps { get; }
```
iantalarico: It may be useful to make this an IList to be explicit that it has order.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+            NavigateToRootCommand = new ProtectedCommand(OnNavigateToRootCommand);
+            NavigateToCommand = new ProtectedCommand<PathStep>(OnNavigateToCommand);
+            RefreshCommand = new ProtectedCommand(OnRefreshCommand);
+            DoubleClickCommand = new ProtectedCommand<GcsRow>(OnDoubleClickCommand);
+        }
+
+        /// <summary>
+        /// Method called to upload the given set of files.
+        /// </summary>
+        /// <param name="files"></param>
```
iantalarico: nit: empty param

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+        public void StartDroppedFilesUpload(IEnumerable<string> files)
```
iantalarico: This is also awesome :-)

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+
+            UpdateContextMenu();
+        }
+
+        /// <summary>
+        /// This method creates the <seealso cref="GcsFileOperation"/> instances that represent the upload
+        /// operations for the given paths of files. If the <paramref name="sources"/> entry represents a directory
+        /// then it will recurse into the directories to create the upload operations for those files as well.
+        /// </summary>
+        private IEnumerable<GcsFileOperation> CreateUploadOperations(IEnumerable<string> sources)
```
iantalarico: A param comment for sources would be helpful. Are they absolute paths?

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+                                bucket: Bucket.Name,
+                                destination: $"{CurrentState.CurrentPath}{info.Name}"),
+                          };
+                   }
+               })
+               .SelectMany(x => x);
+
+        /// <summary>
+        /// Creates the <seealso cref="GcsFileOperation"/> instances for all of the files in the given directory. The
+        /// target directory will be ased on <paramref name="basePath"/>.
```
iantalarico: based?

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+
+        /// <summary>
+        /// The download command is implemented with the following steps:
+        /// 1) The user is prompted for the directory that will serve as the root of all of the downloads.
+        /// 2) The list of files to be downloaded, including subdirectories is collected.
+        /// 3) The subdirectories that will receive downloads are created under the download root.
+        /// 4) The file operations are started and the progress dialog is shown to the user. The same <seealso cref="CancellationToken"/>
+        ///    is used for all operations so the progress dialog can be used to cancel all operations.
+        /// </summary>
+        private async void OnDownloadCommand()
```
iantalarico: This should return a Task

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+                tokenSource: tokenSource);
+        }
+
+        /// <summary>
+        /// The delete command is implemented with the following steps:
+        /// 1) The user is asked to confirm the operation.
+        /// 2) The files to be deleted are collected, this involves listing requests sent to the server for
+        ///    the subdirectories being deleted. For each file to be deleted a <seealso cref="GcsFileOperation"/> instance
+        ///    is created to track the progress of the operation. Each delete operation is independent.
+        /// 3) The operations are started and the progress dialog that tracks the progress of all operations is opened. All of
```
iantalarico: "all of the" or "all"

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+            // 1) The user is asked to confirm the deletion operation.
+            if (!UserPromptUtils.ActionPrompt(
+                prompt: Resources.GcsFileBrowserDeletePromptMessage,
+                title: Resources.UiDeleteButtonCaption,
+                actionCaption: Resources.UiDeleteButtonCaption,
+                cancelCaption: Resources.UiCancelButtonCaption))
+            {
+                return;
+            }
+
```
iantalarico: super nit: two empty line

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+        /// 1) The user is asked to confirm the operation.
+        /// 2) The files to be deleted are collected, this involves listing requests sent to the server for
+        ///    the subdirectories being deleted. For each file to be deleted a <seealso cref="GcsFileOperation"/> instance
+        ///    is created to track the progress of the operation. Each delete operation is independent.
+        /// 3) The operations are started and the progress dialog that tracks the progress of all operations is opened. All of
+        ///    operations use the same <seealso cref="CancellationToken"/> and the <seealso cref="CancellationTokenSource"/> is
+        ///    passed to the progress dialog so the user can cancel the operations.
+        /// 4) The listing of objects is invalidated and the window refreshed.  
+        /// </summary>
+        private async void OnDeleteCommand()
```
iantalarico: Should return Task

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -0,0 +1,556 @@
...
+        }
+
+        #endregion
+
+        private void UpdateCurrentState()
+        {
+            UpdateCurrentState(CurrentState.CurrentPath);
+        }
+
+        private async void UpdateCurrentState(string newPath)
```
iantalarico: Should return Task

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
ivannaranjo: No, the idea is that the operation is started and you get all notifications throu the `IGcsFileOperation` interface. Methods marked as `async` don't need to return a `Task` unless you intend for the user to be able to `await` until the method is logically done. In this case I intend for the caller to only get notifications through the provided `IGcsFileOperation`, which is why I am returning `void`.

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
ivannaranjo: In a more general API yes, for us the server content identification is good enough. If we need to augment this with local code it would be hidden away in this method anyway. If we ever need to specify the content type later one we can add it.

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
ivannaranjo: Really good point, will catch `IOException`.

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
ivannaranjo: Same explanation as before, these are all fire and forget methods.

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
ivannaranjo: Same.

GoogleCloudExtension/GoogleCloudExtension.DataSources/IGcsFileOperation.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.Utils/ControlBehaviors.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.Utils/DataGridBehaviors.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gcs/BucketViewModel.cs
ivannaranjo: Thank you :)

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserState.cs
ivannaranjo: `IEnumerable<T>` have an order, the difference is whether you get random access to the elements in the collection or not. In this case You don't need the random access to the elements so I went with the lower version of the interface. If we ever need to do that we can change it.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Thank you :).

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Indeed.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: No it can't. Event handlers are not supposed to return anything.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Not really :).

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Fixed.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Not really, the caller doesn't need to wait for the method to finish.

ivannaranjo:
Thank you for the ongoing review, fixed most of the comments. PTAL.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserState.cs
iantalarico: I didn't think IEnumerable guaranteed order?

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsColumnSorter.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (lhs_value is string)
+            {
+                var lhs_str = (string)lhs_value;
+                var rhs_str = (string)rhs_value;
+
+                var comparison = lhs_str.CompareTo(rhs_str);
+                return descending ? -1 * comparison : comparison;
+            }
+
+            if (lhs_value is ulong)
```
iantalarico: Can columns have other types? Datetime? int?

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsDataSourceExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+        public GcsFileReference(string bucket, string name, string prefix)
+        {
+            Bucket = bucket;
+            Name = name;
+            RelativeName = GetRelativeName(name, prefix);
+        }
+
+        private static string GetRelativeName(string fullName, string prefix)
+        {
+            if (fullName.StartsWith(prefix))
```
iantalarico: Suggestion:

private static string GetRelativeName(string fullName, string prefix) =>
    fullName.StartsWith(prefix) ? fullName.Substring(prefix.Length) : fullName;

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsFileBrowserWindowControl.xaml.cs
```diff
@@ -0,0 +1,65 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Linq;
+using System.Windows;
+using System.Windows.Controls;
+//------------------------------------------------------------------------------
+// <copyright file="GcsFileBrowserWindowControl.xaml.cs" company="Company">
```
iantalarico: This seems out of place?

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsFileBrowserWindowControl.xaml
```diff
@@ -0,0 +1,144 @@
...
+                                <!-- Icon for the name. -->
+                                <Grid Margin="0,0,5,0">
+                                    <Image Source="{mb:ImageResource GcsFileBrowser/Resources/file.png}"
+                                           Visibility="{Binding IsFile, Converter={StaticResource commonVisibilityConverter}}"/>
+                                    <Image Source="{mb:ImageResource GcsFileBrowser/Resources/folder.png}"
+                                           Visibility="{Binding IsDirectory, Converter={StaticResource commonVisibilityConverter}}"/>
+                                </Grid>
+                                
+                                <!-- The name of the object. -->
+                                <TextBlock Text="{Binding FileName}" Style="{StaticResource CommonTextStyle}" />
```
iantalarico: Should this be Name instead of FileName?  Is the comment wrong?  I also may be getting my files mixed up :)

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsRow.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// </summary>
+        public string GcsPath => $"gs://{Bucket}/{Name}";
+
+        public GcsRow() { }
+
+        /// <summary>
+        /// Creates the row given the directory name.
+        /// </summary>
+        /// <param name="bucket">The bucket that owns the directory.</param>
+        /// <param name="name">The name (path) to the directory.</param>
```
iantalarico: This says name (path) but assigns to Name which is not the path.  I think this is part of the confusion in an above comment

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsRow.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <returns>The newly created <seealso cref="GcsRow"/>.</returns>
+        public static GcsRow CreateDirectoryRow(string bucket, string name) =>
+            new GcsRow
+            {
+                Bucket = bucket,
+                Name = name,
+                FileName = GetLeafName(name),
```
iantalarico: Should this be:
Name = GetLeafName(name), 
FileName = name,

GoogleCloudExtension/GoogleCloudExtension/NamePrompt/NamePromptViewModel.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using GoogleCloudExtension.Utils;
+using System;
+using System.Linq;
+using System.Windows.Input;
+
+namespace GoogleCloudExtension.NamePrompt
+{
+    /// <summary>
+    /// The view model for the name prompt window.
```
iantalarico: This is to name a file or dir? Can you elaborate here what this is for?

GoogleCloudExtension/GoogleCloudExtension/NamePrompt/NamePromptWindow.cs
```diff
@@ -0,0 +1,46 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GoogleCloudExtension.Theming;
+
+namespace GoogleCloudExtension.NamePrompt
+{
+    /// <summary>
+    /// This class represents the dialog used to prompt for a name.
```
iantalarico: Same here

GoogleCloudExtension/GoogleCloudExtension/NamePrompt/NamePromptWindow.cs
```diff
@@ -0,0 +1,46 @@
...
+            Content = new NamePromptContent
+            {
+                DataContext = ViewModel
+            };
+        }
+
+        /// <summary>
+        /// Prompts the user for a name.
+        /// </summary>
+        /// <returns></returns>
```
iantalarico: nit: empty returns

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserState.cs
ivannaranjo: It does, using `IEnumerable` in a list will yield the items in the order of the list. You can also force a particular order using LINQ for example. In this case I am using the minimum interface that will work. I am not very consistent on this however. Another nice property of `IEnumerable` is that it is immutable, as you cannot use it to modify the underlying collection.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsColumnSorter.cs
```diff
@@ -0,0 +1,131 @@
+﻿// Copyright 2016 Google Inc. All Rights Reserved.
```
ivannaranjo: This was created last year, have been working on this PR for a while. I'll modify the new files in this PR to match this year.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsColumnSorter.cs
```diff
@@ -0,0 +1,131 @@
...
+using GoogleCloudExtension.Utils;
+using System;
+using System.Diagnostics;
+using System.Reflection;
+using System.Windows.Markup;
+
+namespace GoogleCloudExtension.GcsFileBrowser
+{
+    /// <summary>
+    /// This class sorts the GcsItem instances with directories always first. Because this class is
```
ivannaranjo: Clarified the comment a bit, hopefully.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsColumnSorter.cs
```diff
@@ -0,0 +1,131 @@
...
+        #region MarkupExtension implementation.
+
+        public override object ProvideValue(IServiceProvider serviceProvider) => this;
+
+        #endregion
+
+        #region IColumnSorter implementation.
+
+        /// <summary>
+        /// This method comapres to <seealso cref="GcsRow"/> instances returns the right ordering
```
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsColumnSorter.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// depending <paramref name="descending"/> and <seealso cref="SortDirectories"/>.
+        /// </summary>
+        /// <param name="x">The first object to compare.</param>
+        /// <param name="y">The second object to compare.</param>
+        /// <param name="descending">The direction of comparison.</param>
+        /// <returns>See <seealso cref="IColumnSorter"/> for details.</returns>
+        public int Compare(object x, object y, bool descending)
```
ivannaranjo: Indeed, I'll add them in a subsequent PR. Opened #398 to track this work.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsColumnSorter.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                Debug.WriteLine($"Unknown types passed to the {nameof(GcsColumnSorter)}");
+                return 0;
+            }
+
+            // If the items being compared is a directory vs. a file then the directory will
+            // always be first.
+            if (lhs.IsDirectory != rhs.IsDirectory)
+            {
+                return lhs.IsDirectory ? -1 : 1;
```
ivannaranjo: No, the comment I wrote for the `SortDirectories` was wrong. The property means whether the directories order is to follow the `descending` order or be fixed. In the final order directories will always have to be above the files, the question is the relative order of the directories.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsColumnSorter.cs
ivannaranjo: They can, but the type of columns that we're going to be sorting in this grid is known and limited to the current set. If we were to also add new types then we will also modify the sorter.

The list of columns is determined in the Xaml, not just the columns coming from the source object.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsDataSourceExtensions.cs
ivannaranjo: Accepted.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsFileBrowserWindowControl.xaml.cs
ivannaranjo: Indeed, gone.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsFileBrowserWindowControl.xaml
ivannaranjo: No, it is FileName, it is coming from `GcsRow`. This is actually the name of the leaf in the path. Fixed the comment in the source class as well.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsRow.cs
ivannaranjo: Well, GCS calls those names not paths, because GCS doesn't actually have paths. Each blob as a "name" and belongs to a bucket. This is trying to follow that same convention.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsRow.cs
ivannaranjo: No, see above. `Name` is the name of the blob, `FileName` is the final part. Maybe I have to rename to `BlobName` and `LeafName`.

GoogleCloudExtension/GoogleCloudExtension/NamePrompt/NamePromptViewModel.cs
ivannaranjo: It is a general "name" dialog, which is why I am vague about it. It is used to prompt the user to enter a name for something. Now in this case is used for directories.

GoogleCloudExtension/GoogleCloudExtension/NamePrompt/NamePromptWindow.cs
ivannaranjo: Because that's what it does, prompt the user for a name for something.

GoogleCloudExtension/GoogleCloudExtension/NamePrompt/NamePromptWindow.cs
ivannaranjo: Fixed.

ivannaranjo:
Addressed the comments, PTAL.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsRow.cs
iantalarico: Ah okay I see now.  It may help to rename but I don't think it's required.

Deren-Liao:
Good.  Sorry for not having time to look into this.

PR #393, opened 2017-02-02
Cleanup the name to ensure no invalid chars.

### LINKED ISSUE #381: Can't add an account with an existing password via Manage Windows Credentials
This is what happens when I try:
image

If I try again after this, I get VS crashing.

ivannaranjo:
The cause of this bug was the fact that the project ID had a character that is not a valid character for a path. We were taking the project ID verbatim and used it in the path for the credentials file. With this fix we will cleanup invalid chars from the project ID and replace them with the "_" character.

This PR fixes #381.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
@@ -156,13 +156,10 @@ namespace GoogleCloudExtension.Accounts
         private static string GetInstancePath(Instance instance)
         {
             var credentials = CredentialsStore.Default;
-            return $@"{GetValidName(credentials.CurrentProjectId)}\{GetValidName(instance.GetZoneName())}\{GetValidName(instance.Name)}";
+            return $@"{ToValidPathName(credentials.CurrentProjectId)}\{ToValidPathName(instance.GetZoneName())}\{ToValidPathName(instance.Name)}";
         }
 
-        private static string GetValidName(string name)
-        {
-            return s_invalidNameCharPattern.Replace(name, "_");
-        }
+        private static string ToValidPathName(string name) => s_invalidNameCharPattern.Replace(name, "_");
 
         private static string GetFileName(WindowsInstanceCredentials credentials) => $"{credentials.User}{PasswordFileExtension}";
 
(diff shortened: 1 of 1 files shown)


GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -40,6 +41,7 @@ internal class WindowsCredentialsStore
 
         private static readonly Lazy<WindowsCredentialsStore> s_defaultStore = new Lazy<WindowsCredentialsStore>();
         private static readonly string s_credentialsStoreRoot = GetCredentialsStoreRoot();
+        private static Regex s_invalidNameCharPattern = new Regex("[;:\\?\\\\]");
```
iantalarico: Regexs are hard (at least for me) Can you add a comment about what it does exactly?

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -154,7 +156,12 @@ private static string GetCredentialsStoreRoot()
         private static string GetInstancePath(Instance instance)
         {
             var credentials = CredentialsStore.Default;
-            return $@"{credentials.CurrentProjectId}\{instance.GetZoneName()}\{instance.Name}";
+            return $@"{GetValidName(credentials.CurrentProjectId)}\{GetValidName(instance.GetZoneName())}\{GetValidName(instance.Name)}";
+        }
+
+        private static string GetValidName(string name)
```
iantalarico: Suggestion:
private static string GetValidName(string name) => return s_invalidNameCharPattern.Replace(name, "_");

Deren-Liao: Nit: GetValidFilePathName or ToValidFileName etc

ivannaranjo: True, true... I still resist this style somehow...

ivannaranjo: Settled on `ToValidPathName` since it is going to be part of a path not necessarily the file part.

Issue #371, opened 2017-01-21
Show app engine instance id in the explorer


Deren-Liao:
I am using the version merged from q1_master.  If you believe I am using a wrong version, just close it.

As shown in the following picture, the  app engine instance  does not show any label.  If this is not a bug, I'd ask you to consider to add instance id to it.

![gae_instance_no_label](https://cloud.githubusercontent.com/assets/22829476/22171342/3bf37986-df40-11e6-9102-6d4c8232ed6e.PNG)

ivannaranjo:
This is referring to the name of the GCE instance that runs the backend for the app. Yes, this is a bug, the instance name should be shown correctly.

Issue #370, opened 2017-01-21
Google App Engine tree does not show flexible versions when I choose to show only flexible versions.  


Deren-Liao:
I am using the version merged from q1_master.

I refreshed the Google Cloud Explorer for many times. It successfully show the version once only.

![gae_instance_no_versions_found_error](https://cloud.githubusercontent.com/assets/22829476/22171190/c6fb6316-df3d-11e6-9fab-5a25ccee8681.PNG)

![gae_instance_no_versions_found_pantheon_page](https://cloud.githubusercontent.com/assets/22829476/22171194/c994f330-df3d-11e6-8041-90a564aaeb57.PNG)

Deren-Liao:
The version node does show a spinning progress icon, but it ends up with no results.

Deren-Liao:
Now I figured out, the reason was I chose Flexible versions only.  
Then if In open the context menu, it shows me "Show Flexible and Standard versions".  This is good. The problem is I weren't aware it was showing only flexible versions.

Deren-Liao:
Sorry for this keep updating comment.  This really confused me as this is my first time using app engine. 
(1) I used VS to publish a .NET core app that it automatically created the default service for me. 
(2) I now realize the instances are created on "Flexible" environment.
(3) When I chose "Show flexible versions only" in the Google Explorer,  it shows me no results.  This is a bug too.

Issue #348, opened 2017-01-05
Cannot see App Engine versions deployed to non-default services.


ivannaranjo:
The service is listed under the "Google App Engine" node but no versions are listed under the service.

Issue #303, opened 2016-11-22
Need to expand a service in the App Engine node before the "Split" menu item can be used


ivannaranjo:
Currently when expanding the "Google App Engine" in the cloud explorer the "Split Traffic" option in the context menu for the service node is disabled until the user also expands that service. This is because the loading of the list of versions is done lazily.

This is quite confusing because at first it seems that the split operation cannot be performed. Then when the user drills down this operation is available. The extension should pre-load all of the versions when loading the services and then determine what operations are available or not.

PR #401, opened 2017-02-07
Master app engine load upfront

ivannaranjo:
This PR changes the model of how the GAE source works in the cloud explorer. Instead of loading parts of the information on demand we are going to be loading everything upfront in parallel. This fixes issues with whether we should enable the traffic split options or not, which was only enabled after the user expanded the version node.

This PR also fixes some minor bugs, the complete list is:
* Issue #371, using the id of the instance instead of the `VmName` property which inexplicably is empty.
* Issue #370, which came from how we detect whether a version is running in Flex or not. Also changed the defaults to show everything by default, the user can then filter down if needed.
* Issue #348, which I am now thinking was my own interpretation of #370.
* Issue #303, which required the user to expand the version node before the "Split Traffic" option was enabled. This was very counter intuitive. This bug more than anything else motivated to load everything upfront.
* Issue #243, the user could not navigate to a previous, and still serving, version on a service. This has been fixed by allowing opening the version as long as it is serving.

This PR also brings the question of whether we should keep showing the instances for a particular version. There's no much information that we can offer up to the user and it makes the loading process longer. Maybe this would be better served in the GCE source instead? Or perhaps loading the images can be done on demand still. Food for thought.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.DataSources/GaeVersionExtensions.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.DataSources/GaeVersionExtensions.cs
@@ -22,7 +22,7 @@ namespace GoogleCloudExtension.DataSources
         /// <summary>
         /// The value of a .Net runtime for a GAE version.
         /// </summary>
-        public const string DotNetRuntime = "aspnetcore";
+        public const string AspNetCoreRuntime = "aspnetcore";
 
         /// <summary>
         /// The serving status for a GAE version serving.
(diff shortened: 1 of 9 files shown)


GoogleCloudExtension/GoogleCloudExtension.DataSources/GaeVersionExtensions.cs
```diff
@@ -22,7 +22,7 @@ public static class GaeVersionExtensions
         /// <summary>
         /// The value of a .Net runtime for a GAE version.
         /// </summary>
-        public const string DotNetRuntime = "dotnet";
+        public const string DotNetRuntime = "aspnetcore";
```
iantalarico: Nit: rename this to be aspnetcore specific?

ivannaranjo: Done.

PR #407, opened 2017-02-08
Setting the working directory to the project's root for ASP.NET Core apps when deploying.

### LINKED ISSUE #342: Calls to dotnet CLI should set the working directory
dotnet determines which SDK version to use by walking up from the CWD until it finds a global.json.

It looks like this isn't happening when calling "dotnet publish", as preview4 was being used for me, instead of preview2 as specified in global.json. Renaming my preview4 directory fixed this as a workaround.

ivannaranjo:
This PR fixed issue #342 by adding functionality to the `ProcessUtils` class to set the working directory when launching a process. This new functionality is then used to set the working directory to the project's directory. With this now the "dotnet" tool should be able to find the right "global.json" file.

The way the "dotnet" tool is launched for publishing the app is now also simplified. Since we're setting the current directory to the project's root directory we don't need to explicitly point to the "project.json" file, the tool will find the right one automatically.

This fixes #342

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Utils/ProcessUtils.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Utils/ProcessUtils.cs
@@ -186,7 +186,8 @@ namespace GoogleCloudExtension.Utils
             IDictionary<string, string> environment)
         {
             // If the caller provides a working directory use it otherwise default to the user's profile directory
-            // random intereference from Visual Studio changing the working directory often.
+            // so we have a stable working directory instead of a random working directory as Visual Studio changes the
+            // current working directory often.
             ProcessStartInfo startInfo = new ProcessStartInfo
             {
                 FileName = file,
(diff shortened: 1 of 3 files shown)


GoogleCloudExtension/GoogleCloudExtension.Utils/ProcessUtils.cs
```diff
@@ -177,32 +179,30 @@ public static Task<int> LaunchCommandAsync(string file, string args, IDictionary
...
+            string file,
+            string args,
+            string workingDir,
+            IDictionary<string, string> environment)
         {
-            // Always start the tool in the user's home directory, avoid random directories
-            // coming from Visual Studio.
-            ProcessStartInfo result = new ProcessStartInfo
+            // If the caller provides a working directory use it otherwise default to the user's profile directory
+            // random intereference from Visual Studio changing the working directory often.
```
iantalarico: interference?

iantalarico: I'm not sure what the second part of the comment is trying to say

ivannaranjo: Clarified the comment, I hope. Done.

PR #409, opened 2017-02-09
Ensuring all GAE Property window items have descriptions and display names, removing instances.

### LINKED ISSUE #284: Need to have descriptions in the GAE InstanceView for the properties window
The properties in the GAE's `InstanceItem` class do not have descriptions, when selecting the properties in the properties window only the name is shown.

We should have descriptions assigned to them so the user get's a hint of the meaning of the properties.

ivannaranjo:
This PR primary purpose is to add display name and descriptions for all properties in GAE items. Because we have determined that the GAE items are not useful enough this PR also removes the instance classes instead of adding the annotations to them.

Fixes #284.


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionItem.cs
```diff
@@ -30,49 +30,45 @@ public VersionItem(Version version) : base(className: Resources.CloudExplorerGae
...
+        public string ServingStatus => _version.ServingStatus;
 
         [LocalizedCategory(nameof(Resources.CloudExplorerGaeVersionCategory))]
-        public string Deployer => _version.CreatedBy;
+        [LocalizedDisplayName(nameof(Resources.CloudExplorerGaeVersionCreatedByDisplayName))]
+        [LocalizedDescription(nameof(Resources.CloudExplorerGaeVersionCreatedByDescription))]
+        public string CreatedBy => _version.CreatedBy;
+
+        [LocalizedDisplayName(nameof(Resources.CloudExplorerGaeVersionCreationTimeDisplayName))]
+        [LocalizedCategory(nameof(Resources.CloudExplorerGaeVersionCategory))]
```
iantalarico: super nit: move LocalizedCategory to the top to match others

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1419,6 +1355,50 @@
     <value>Invalid deployment version {0}. The version must only use lowercase letters, digits and the - character.</value>
     <comment>Message shown for an invalid deployment version. {0} is the deployment version given by the user.</comment>
   </data>
+  <data name="CloudExplorerGaeServiceIdDescription" xml:space="preserve">
+    <value>Relative name of the service within the application.</value>
+    <comment>The description of the Id property.</comment>
+  </data>
+  <data name="CloudExplorerGaeServiceNameDescription" xml:space="preserve">
+    <value>Full path to the Service resource in the API.</value>
```
iantalarico: Nit: you capitalize service and version in some descriptions and not others.  I'm not sure which is correct though

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionItem.cs
ivannaranjo: Done. I also like the regularity.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
ivannaranjo: Not sure, I am going to go with all lowercase to start. Will go through all the strings anyway.

PR #420, opened 2017-02-14
Adding support for parsing the 'service: name' portion of an app.yaml.

### LINKED ISSUE #419: Need to parse the app.yaml if it exists looking for the service: tag.
When the user ticks the "open website" checkbox in the Flex deployment publish step we need to parse the yaml file to get the name of the service, if it is not the default one, so we can create the right URL for the just deployed app.

ivannaranjo:
This PR adds a simple parser that looks for the "service: name" portion of an app.yaml to determine the name of the service to navigate when the user requests that the extension opens the website after a successful deployment.

Fixes #419.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
@@ -179,7 +179,7 @@ namespace GoogleCloudExtension.Deployment
         /// <summary>
         /// This methods looks for lines of the form "service: name" in the app.yaml file provided.
         /// </summary>
-        /// <param name="appYamlPath">The path to the app.yaml to parse.</param>
+        /// <param name="projectPath">The path to the project.json for the project, the app.yaml should be next to it.</param>
         /// <returns>The service name if found, <seealso cref="DefaultServiceName"/> if not found.</returns>
         private static string GetAppEngineService(string projectPath)
         {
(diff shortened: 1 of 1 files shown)


GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
```diff
@@ -175,19 +176,38 @@ public static ProjectConfigurationStatus CheckProjectConfiguration(string projec
             return new ProjectConfigurationStatus(hasAppYaml: hasAppYaml, hasDockerfile: hasDockefile);
         }
 
+        /// <summary>
+        /// This methods looks for lines of the form "service: name" in the app.yaml file provided.
+        /// </summary>
+        /// <param name="appYamlPath">The path to the app.yaml to parse.</param>
```
iantalarico: nit: update param name here or below

GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
```diff
@@ -175,19 +176,38 @@ public static ProjectConfigurationStatus CheckProjectConfiguration(string projec
...
+            if (File.Exists(appYaml))
             {
-                // TODO: Load the app yaml and look for the service key.
-                return DefaultServiceName;
+                try
+                {
+                    var lines = File.ReadLines(appYaml);
+                    foreach (var line in lines)
+                    {
+                        if (line.StartsWith(ServiceStatement))
```
iantalarico: do you need to trim the line? could white space be before the service statement?

ivannaranjo: Yaml is whitespace sensitive, there shouldn't be any white space before "service: " since it is at the top level. If I trim it I could get the wrong setting.

GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
ivannaranjo: Done, fixed the comment.

PR #422, opened 2017-02-14
Showing long error details on optional TextBox.

### LINKED ISSUE #397: Error message for a failure to set a pw is long and confusing
Can we format this message better? Perhaps provide a summary and details? And a scrollbar?

vs-error-msg-long

ivannaranjo:
The error messages coming from gcloud are specially bad because tend to be quite long and they make the error dialog unreadable. This PR adds a readonly `TextBox` to the error dialog to display the error details in a useable way.

The error dialog with error details will now look like this:
screen shot 2017-02-14 at 5 08 29 pm

Fixes #397


ivannaranjo:
Sadly that's a gcloud integration issue. The error message really is the stderr output from gcloud. I don't want to get in the business of parsing the output because then it will make us more fragile. This is an area to improve with gcloud.

PR #433, opened 2017-02-16
Add setting the GOOGLE_CLOUD_ENVIRONMENT variable.

### LINKED ISSUE #432: Add support for environments for Flex and Gke for ASP.NET Core apps.
ASP.NET Core has a notion of application settings stored in appsettings.json file. These application settings can be made environment specific, so different settings are loaded depending on what environment the app is running in.

We plan on creating default app.yaml (for Flex) and deployment specification (for GKE) that set value of the `ASPNETCORE_ENVIRONMENT` to a value that depends the environment on which the app is running:
* For App Engine Flex the value of the variable will be `flex`.
* For GKE/Container Engine the value of the variable will be `gke`.

The user can then set various configuration options in the appsettings.json depending on where the app is running.

ivannaranjo:
This PR modifies the default app.yaml and GKE configuration to set the GOOGLE_CLOUD_ENVIRONMENT to either "flex" for App Engine Flex or "gke" for when running on GKE.

With this environment variable the app can select the right configuration to use, etc... 

Fixes #432

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
@@ -32,7 +32,12 @@ namespace GoogleCloudExtension.Deployment
 
         private const string AppYamlDefaultContent =
             "runtime: custom\n" +
-            "env: flex\n";
+            "env: flex\n" +
+            "\n" +
+            "# Setting the environment vairable to flex so the app can use that\n" +
+            "# for extra configuration.\n" +
+            "env_variables:\n" +
+            "  ASPNETCORE_ENVIRONMENT: flex\n";
 
         private const string DefaultServiceName = "default";
         private const string ServiceStatement = "service:";
--- a/GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
@@ -37,7 +37,7 @@ namespace GoogleCloudExtension.Deployment
             "# Setting the environment vairable to flex so the app can use that\n" +
             "# for extra configuration.\n" +
             "env_variables:\n" +
-            "  ASPNETCORE_ENVIRONMENT: flex\n";
+            "  GOOGLE_CLOUD_ENVIRONMENT: flex\n";
 
         private const string DefaultServiceName = "default";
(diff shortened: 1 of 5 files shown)


ivannaranjo:
// cc: @jskeet FYI

jskeet:
Let's hold off on actually merging until we've tried it and potentially consulted the ASP.NET team to see whether this feels idiomatic to them - but it seems reasonable to me.

GoogleCloudExtension/GoogleCloudExtension.Deployment/CommonUtils.cs
```diff
@@ -26,6 +26,11 @@ namespace GoogleCloudExtension.Deployment
     internal static class CommonUtils
     {
         /// <summary>
+        /// Environment variable to be set with the current environment in which the app is running.
+        /// </summary>
+        public const string GoogleCloudEnvironmentVar = "GOOGLE_CLOUD_ENVIRONMENT";
```
iantalarico: This isn't a previously used var right?  It might be useful to add a note here about this being the only place used.

ivannaranjo: Yep, will update the comment.

GoogleCloudExtension/GoogleCloudExtension.Deployment/AppEngineFlexDeployment.cs
```diff
@@ -32,7 +32,12 @@ public static class AppEngineFlexDeployment
 
         private const string AppYamlDefaultContent =
             "runtime: custom\n" +
-            "env: flex\n";
+            "env: flex\n" +
+            "\n" +
+            "# Setting the environment vairable to flex so the app can use that\n" +
+            "# for extra configuration.\n" +
+            "env_variables:\n" +
+            "  GOOGLE_CLOUD_ENVIRONMENT: flex\n";
```
Deren-Liao: How do you come up this "flex" name?   If this is intended to be used by apps, do we have document that help developers know of it?

ivannaranjo:
Agreed, let's consult and see what happens. This environment variable really is to help the user with configuration and it is, as much in ASP.NET Core, a pattern or good practice to follow. It is not a feature of the base runtime, only a suggestion of a way of doing things for the user. Eventually if we go this way the templates for GCP projects will incorporate code to take this variable into account as well.

ivannaranjo:
This is not the right solution at this point, but it helps move into the right direction. Closing for now.

PR #434, opened 2017-02-16
Ensuring right credentials for kubectl calls.

### LINKED ISSUE #424: Need to ensure that gcloud container does not use the default credentials.
By setting the env variable  `CLOUDSDK_CONTAINER_USE_APPLICATION_DEFAULT_CREDENTIALS` to `false` we can ensure that the container will use gcloud's credentials. We want this behavior because the VS extension provides the credentials directly to gcloud to control what environment is running in.

ivannaranjo:
There's an issue in which `kubectl` will use the application default credentials by default to obtain the credentials for calls into the cluster. If a user has more than one account registered with gcloud the default credentials will be one of the one registered with gcloud. This is also independent of the accounts registered with Visual Studio.

This could be workaround by setting `CLOUDSDK_CONTAINER_USE_APPLICATION_DEFAULT_CREDENTIALS` to `false` except that exposes a bug in how `kubectl` quotes the path to gcloud to invoke it to fetch credentials.

In order to workaround these two issues this PR changes the extension to set `CLOUDSDK_CONTAINER_USE_APPLICATION_DEFAULT_CREDENTIALS` to `true` to force `kubectl` to use the application default credentials and then the extension will also set the value of the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to the credentials file that corresponds to the user selected in the extension. This way if `kubectl` will always use the same credentials as the extension no matter what.

This also as it turns out is the right behavior for wrapping `kubectl` and ensuring credentials, as we don't want `kubectl` to use `gcloud` to fetch credentials, we want to supply the credentials to it.

Fixes #424

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
@@ -1,4 +1,18 @@
-﻿namespace GoogleCloudExtension.GCloud
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace GoogleCloudExtension.GCloud
 {
     /// <summary>
     /// This class contains defintions for the environment variables used the various wrappers.
--- a/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
@@ -15,7 +15,7 @@
 namespace GoogleCloudExtension.GCloud
 {
     /// <summary>
-    /// This class contains defintions for the environment variables used the various wrappers.
+    /// This class contains defintions for the environment variables used by the various wrappers.
     /// </summary>
     internal static class CommonEnvironmentVariables
     {
--- a/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
@@ -1,10 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
-namespace GoogleCloudExtension.GCloud
+﻿namespace GoogleCloudExtension.GCloud
 {
     /// <summary>
     /// This class contains defintions for the environment variables used the various wrappers.
--- a/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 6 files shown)


GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlWrapper.cs
```diff
@@ -175,5 +182,20 @@ private static string FormatCommand(string command, KubectlContext context, bool
 
             return $"{command} --kubeconfig=\"{context.ConfigPath}\" {format}";
         }
+
+        /// <summary>
+        /// Returns the environment variables to use to invoke kubectl safely. This environemnt is necessary
+        /// to ensure that the right credentials are used should the access token need to be refreshed.
+        /// </summary>
+        /// <param name="context"></param>
```
iantalarico: nit: empty params and returns

GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
```diff
@@ -0,0 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace GoogleCloudExtension.GCloud
+{
+    /// <summary>
+    /// This class contains defintions for the environment variables used the various wrappers.
```
Deren-Liao: used by ?

GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// </summary>
+    internal static class CommonEnvironmentVariables
+    {
+        // This variables is used to force gcloud to use application default credentials when generating
+        // the cluster information for the cluster.
+        public const string GCloudContainerUseApplicationDefaultCredentialsVariable = "CLOUDSDK_CONTAINER_USE_APPLICATION_DEFAULT_CREDENTIALS";
```
Deren-Liao: nit:  Use a shorter name? When it's too long, it's not convenient to use it in other places.

GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlContext.cs
```diff
@@ -28,9 +29,15 @@ public class KubectlContext : IDisposable
         /// </summary>
         public string ConfigPath { get; private set; }
 
-        public KubectlContext(string configPath)
+        /// <summary>
+        /// Path to the application credentials to use while calling into kubectl.
+        /// </summary>
+        public string CredentialsPath { get; private set; }
+
+        public KubectlContext(string configPath, string credentialsPath)
```
Deren-Liao: Add comment.

GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlWrapper.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.GCloud/CommonEnvironmentVariables.cs
ivannaranjo: Well, that is the unfortunate name of the environment variable, shortening will not really make the code much more readable.

GoogleCloudExtension/GoogleCloudExtension.GCloud/KubectlContext.cs
ivannaranjo: It is a constructor, there's really nothing to comment :).

ivannaranjo:
Thank you for the review. Setting the `GOOGLE_APPLICATION_CREDENTIALS` will only happen during the calls to `kubectl` or the `gcloud container cluster` tools. We actually do need to set the env var because the extension needs to always be in control of what credentials are being used.

It turns out it was good that `kubectl` uses the default credentials because this gives us the chance to override them with the ones for the current user in the extension. If `kubectl` uses the `gcloud` helper then it would be probably be getting the wrong credentials to use.

Issue #437, opened 2017-02-17
Using the official aspnetcore image on the generated Dockerfile for ASP.NET Core apps.


ivannaranjo:
This issue is to track the trivial change to use the official image in the generated Dockerfile for ASP.NET Core apps. The image is at `gcr.io/google-appengine/aspnetcore:1.0.3`.

PR #439, opened 2017-02-17
Using the official runtime tag.

ivannaranjo:
This PR updates the default Dockerfile to use the official image.


ivannaranjo:
Fixes #437

PR #441, opened 2017-02-17
Use the "runtime: aspnetcore" in the generated app.yaml.

ivannaranjo:
This change configures `gcloud` via environment variables so the App Engine Flex deployment uses the runtime builders. This uses the dogfood version of the runtime builder so this change is temporary until the final version is out. This will allow us to verify the end to end deployment.


ivannaranjo:
// cc @chrsmith FYI, this build contains the dogfood version of the end to end story for deploying ASP.NET Core apps to Flex.

ivannaranjo:
Thank you for the review.

ivannaranjo:
Fixes #441

Issue #436, opened 2017-02-17
Update all aspects of a GKE deployment when deploying an app.


ivannaranjo:
If a user is deploying a GKE app and the deployment, and service, already exist we need to ensure that all aspects of the deployment are updated. If the user unchecks the "publish service" checkbox we need to ensure that the service is deleted so the final step is what is expected.

PR #447, opened 2017-02-20
Adding checkbox to choose whether a service is public.

### LINKED ISSUE #438: Separate the notion of public GKE service for exposing a GKE service at all.
Currently the extension only offers the option to expose a public GKE service. Instead it should offer the option to expose a service with an optional public IP address. This would be accomplished by adding another checkbox where the user can select whether the service should be public or not.

This will cause complication of course, if the user updates a private service to be public then it will have to be deleted and re-created to make it public.

ivannaranjo:
This PR adds a notion of having an exposed service that doesn't have a public IP address. This allows us to have private backends that are not exposed to the world.

Having a private, cluster only, service is accomplished by using the "ClusterIP" type for a service. With this type the IP address that the service gets will only be accessible from the cluster itself.

This PR also adds the capability of _updating_ an existing service. If the user requests a different type of service, say the previous service was cluster only and the user requested a public service, then the extension will delete the previous service and create a new one. If the user selects the option of not exposing a service then if there's an existing service it will be deleted.

Fixes #438

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Deployment/GkeDeployment.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Deployment/GkeDeployment.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using GoogleCloudExtension.GCloud;
+using GoogleCloudExtension.GCloud.Models;
 using GoogleCloudExtension.Utils;
 using System;
 using System.Diagnostics;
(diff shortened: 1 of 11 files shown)


ivannaranjo:
// CC @jskeet FYI. This is what is going to allow the user to expose a non-public service from the extension.

jskeet:
SGTM. Happy to give it a whirl when you think it's ready for dogfooding.

ivannaranjo:
Also fixed #436.

PR #448, opened 2017-02-21
Add "Migrage all traffic" command for App Engine versions in Cloud Explorer

### LINKED ISSUE #244: Need to have a command to assign 100% of traffic to a given version.
Going through the allocation dialog for this common operation is too cumbersome. It should be a single command in the context menu for the version, similar to the old "Set as Default version" command.

ivannaranjo:
This PR implements a "Migrate all traffic" command for App Engine versions in Cloud Explorer as a shortcut to change the version that serves all traffic for a service. The same operation could be done by changing the split settings but this short cut will make things much more convenient. This also mirrors the same option in the Cloud Console.

There's also a bit of refactoring the `Poll<T>` class, making it internal to the `GoogleCloudExtension.DataSources` project and exposing `AwaitOperationAsync()` methods for those data sources that offer the option for creating operations. This way the client code only has to `await` a `Task` without having to deal with the details under the covers.

Fixes #244

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionViewModel.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionViewModel.cs
@@ -119,18 +119,10 @@ namespace GoogleCloudExtension.CloudExplorerSources.Gae
         {
             try
             {
-                var split = new TrafficSplit
-                {
-                    Allocations = new Dictionary<string, double?>
-                    {
-                        [_version.Id] = 1.0
-                    }
-                };
-
-                var datasource = _owner.DataSource;
                 IsLoading = true;
                 Caption = String.Format(Resources.CloudExplorerGaeMigratingAllTrafficCaption, _version.Id);
 
+                var split = new TrafficSplit { Allocations = new Dictionary<string, double?> { [_version.Id] = 1.0 } };
                 var operation = await _owner.DataSource.UpdateServiceTrafficSplitAsync(split, _service.Id);
                 await _owner.DataSource.AwaitOperationAsync(operation);
                 _owner.InvalidateService(_service.Id);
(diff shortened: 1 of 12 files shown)


ivannaranjo:
// CC @jskeet @chrsmith, new command for managing App Engine versions. In case you want to play with it.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionViewModel.cs
```diff
@@ -104,7 +105,7 @@ private void UpdateMenu()
             }
 
             // If the version is stopped and has no traffic allocated to it allow it to be deleted.
-            if (!_hasTrafficAllocation && _version.IsStopped())
+            if (HasTrafficAllocation && _version.IsStopped())
```
iantalarico: should this be !has? Or was this a bug before?  Also update comment if so?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionViewModel.cs
```diff
@@ -114,6 +115,26 @@ private void UpdateMenu()
...
+                Caption = String.Format(Resources.CloudExplorerGaeMigratingAllTrafficCaption, _version.Id);
+
+                var split = new TrafficSplit { Allocations = new Dictionary<string, double?> { [_version.Id] = 1.0 } };
+                var operation = await _owner.DataSource.UpdateServiceTrafficSplitAsync(split, _service.Id);
+                await _owner.DataSource.AwaitOperationAsync(operation);
+                _owner.InvalidateService(_service.Id);
+            }
+            catch (DataSourceException ex)
+            {
+                Debug.WriteLine($"Faile to set traffic to 100%: {ex.Message}");
```
iantalarico: nit: failed

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionViewModel.cs
ivannaranjo: Bad typo, thank you for catching it.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionViewModel.cs
ivannaranjo: Done.

PR #454, opened 2017-02-23
Updating ASP.NET 4.x templates to add logging and error reporting.

### LINKED ISSUE #313: Update the project templates to incorporate logging.
Update the project template to incorporate the code to enable logging and error reporting.

ivannaranjo:
This PR updates the ASP.NET 4.x templates to enable logging and error reporting for the WebAPI and MVC ASP.NET 4.x apps.

The PR adds all of the necessary dependencies, .csproj modifications, etc... for the projects to just function out of the box.

Fixes #313


GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/App_Start/FilterConfig.cs
```diff
@@ -1,12 +1,42 @@
...
+using Google.Cloud.Diagnostics.AspNet;
 
 namespace $safeprojectname$
 {
     public class FilterConfig
     {
         public static void RegisterGlobalFilters(GlobalFilterCollection filters)
         {
+	    // To enable Google Cloud Stackdriver Error Reporting:
+            // 1. Enable the Stackdriver Error Reporting API:
```
iantalarico: You actually don't need this now with the latest libraries if you use the defaults, thanks to some magic with error reporting and logging

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/App_Start/WebApiConfig.cs
```diff
@@ -2,16 +2,59 @@
...
+using Google.Cloud.Diagnostics.AspNet;
 
 namespace $safeprojectname$
 {
     public static class WebApiConfig
     {
         public static void Register(HttpConfiguration config)
         {
+	    // To enable Google Cloud Stackdriver Logging and Error Reporting:
+            // 1. Enable the Stackdriver Error Reporting API:
```
iantalarico: same here

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/Web.config
```diff
@@ -7,7 +7,27 @@
...
   </configSections>
+  <!-- // [START log4net_web_config] -->
+  <log4net>
+    <appender name="CloudLogger" type="Google.Cloud.Logging.Log4Net.GoogleStackdriverAppender,Google.Cloud.Logging.Log4Net">
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="%-4timestamp [%thread] %-5level %logger %ndc - %message" />
+      </layout>
+      <projectId value="YOUR-PROJECT-ID" />
+      <serviceName value="NAME-OF-YOUR-SERVICE" />
+      <version value="VERSION-OF-YOUR-SERVCICE" />
```
chrisdunelm: "serviceName" and "version" should not be in the log4net section here.
They are not valid configuration settings for GoogleStackdriverAppender.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/App_Start/FilterConfig.cs
ivannaranjo: Done.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/App_Start/WebApiConfig.cs
ivannaranjo: Done.

ivannaranjo:
PTAL. I have updated the web.config and now I am using the `<appSettings>` section to add the `version` and `serviceName` values to the app. I would do the same thing for the `projectID` but that is already a parameter for `log4net` so it makes sense to only have it there.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/App_Start/FilterConfig.cs
```diff
@@ -1,12 +1,37 @@
...
+	    // Web.config and replace "YOUR-PROJECT-ID" with your Google Cloud
+	    // Project ID
+
+            // [START error_reporting]
+            // Check to ensure that projectId has been changed from placeholder value.
+            var section = (XmlElement)ConfigurationManager.GetSection("log4net");
+            XmlElement element =
+                (XmlElement)section.GetElementsByTagName("projectId").Item(0);
+            string projectId = element.Attributes["value"].Value;
+            if (projectId == ("YOUR-PROJECT-ID"))
```
Deren-Liao: What are all these checking project id doing ? 
I saw it in 3 files.  Looks very inconvenient.  
I think Chris Bacon already added code that auto detect project id.

ivannaranjo: This is also used for the error reporting API, which also requires the project ID. It is done here and it is also done in the Global.asax.cs so it can give the user a meaningful error message if the project ID is not set in the config.

I don't think that we will be able to get the project ID until we have the API for the metadata service released.

Deren-Liao:
I see. I am not holding sign off of your PR.  Just wanted to exchange
information.

For App Engine and VM, Chris Bacon checked in code for detecting project
id.   The code is at  Google.Api.Gax.Platform.  It does not support
container yet.

What metadata service do you refer to?    I vaguely recall I already used
it (on Compute Engine VM ).   I mean we'd better to come up similar
solution for detecting project id.  This simplifies the code a lot.

I think we should let user not to configure project id. If the project id
is different than the underline VM .. the logging service won't even accept
it.

chrisdunelm:
If running on GAE or GCE you can get the projectID using the [Google.Api.Gax.Platform](https://github.com/googleapis/gax-dotnet/blob/master/src/Google.Api.Gax/Platform.cs) class.
Use `Platform platform = Platform.Instance()`, then the use `platform.GaeDetails.ProjectId` or `platform.GceDetails.ProjectId` depending on platform.

(Which makes me think it may be useful to just be able to do: `Platform.Instance().ProjectId`, although that isn't currently available)

ivannaranjo:
I was unaware of being able to get the `ProjectId` from the Gax library. How do we connect that to the log4net configuration since it is fetching the `projectId` from web.config? Does this mean that I do not need to specify the project ID at all on the web.config? Because that would be grand :)

Deren-Liao:
For GAE, GKE , log4net, project id is optional.

ivannaranjo:
Need to be able to support running from the local developer. The project ID will not be available in that case so the user can override it in the config file.

ivannaranjo:
PTAL

Now the template will have code to try to fetch the project ID if no project ID has been specified by the user. This will allow the user to have two Web.config files, one for local testing and another one for production. The user will still have to specify the project ID when running locally, there will be an exception with a (hopefully) friendly error message to indicate what to do to specify the project ID.

All of the code can be deleted of course. Future improvements can include having a singleton class generated in the project that keep the name of the project ID and and identifier of the current environment, GAE vs. GCE.

PR #456, opened 2017-02-24
Error Reporting overview window

Deren-Liao:
The PR adds control content to the overview window.
It has xaml file, view model file etc.


Deren-Liao:
The reported AppVeyor build failure is a bug in the ./tools/ensure_no_unused_strings.sh  .   [#457](https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/issues/457)

I'll change my code to mitigate this error for now.

GoogleCloudExtension/GoogleCloudExtension/Controls/ProgressIndicator.xaml.cs
```diff
@@ -48,6 +38,24 @@ public partial class ProgressIndicator : UserControl, ISupportInitialize
...
+                typeof(int), 
+                typeof(ProgressIndicator), 
+                new PropertyMetadata(500));
+
+        /// <summary>
+        /// Define the FullDuration custom dependency property. 
+        /// This controls the progress moving speed, 
+        /// the smaller this value is, the faster the progress indicator moves.
+        /// </summary>
+        public int FullDuration
```
ivannaranjo: Hmm... Not sure why you want to change the speed of the animation. It is the same images, just moving faster?

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1682,4 +1682,56 @@ Open the project of version {2} to properly locate the logging writter location.
...
+  <data name="ErrorReporting6HoursButtonCaption" xml:space="preserve">
+    <value>6 hours</value>
+    <comment>Caption to the 6 hours button.</comment>
+  </data>
+  <data name="ErrorReporting7DaysButtonCaption" xml:space="preserve">
+    <value>7 days</value>
+    <comment>Caption to the 7 days button.</comment>
+  </data>
+  <data name="ErrorReportingDateTimeFormat" xml:space="preserve">
+    <value>MM-dd-yyyy HH:mm:ss.fff</value>
```
ivannaranjo: This might have to be just using the current locale instead. See if you can change for that.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,136 @@
...
+        public string LastSeenTime => FormatDateTime(ErrorGroup.LastSeenTime);
+
+        /// <summary>
+        /// Initializes a new instance of <seealso cref="ErrorGroupItem"/> class.
+        /// </summary>
+        /// <param name="errorGroup">
+        /// The error group that represents a group of errors.
+        /// <seealso cref="ErrorGroupStats"/>
+        /// </param>
+        public ErrorGroupItem(ErrorGroupStats errorGroup)
```
ivannaranjo: You don't really need to have comments on constructors, it is very clear on what they do.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                throw new ErrorReportingException(new ArgumentNullException(nameof(errorGroup)));
+            }
+
+            ErrorGroup = errorGroup;
+            string[] lines = ErrorGroup.Representative?.Message?.Split(
+                    new string[] { "\r\n", "\n" }, 
+                    StringSplitOptions.RemoveEmptyEntries);
+            Message = lines?[0];
+            Stack = lines?[1];
```
ivannaranjo: Should you check the number of items in `lines`? What if there's only one line?

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,136 @@
...
+    { 
+        /// <summary>
+        /// The error group that represents a group of errors.
+        /// </summary>
+        public ErrorGroupStats ErrorGroup { get; }
+
+        /// <summary>
+        /// Gets the command that navigates to detail window.
+        /// </summary>
+        public ProtectedCommand OnNavigateToDetailCommand { get; }
```
ivannaranjo: This property changes later on, but you are not using the usual pattern of `SetValueAndRaise()`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,136 @@
...
+                throw new ErrorReportingException(new ArgumentNullException(nameof(errorGroup)));
+            }
+
+            ErrorGroup = errorGroup;
+            string[] lines = ErrorGroup.Representative?.Message?.Split(
+                    new string[] { "\r\n", "\n" }, 
+                    StringSplitOptions.RemoveEmptyEntries);
+            Message = lines?[0];
+            Stack = lines?[1];
+            OnNavigateToDetailCommand = new ProtectedCommand(null);
```
ivannaranjo: What? Just leave OnNavigateToDetailCommand set to null. If you change it later use the right value.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        private static string FormatDateTime(object datetime)
```
ivannaranjo: Make this method type specific, formatting a random object to string is probably not what you want to do.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return datetime is DateTime ? 
+                ((DateTime)datetime).ToString(Resources.ErrorReportingDateTimeFormat)
```
ivannaranjo: If the object is not a DateTime you should show a well known error message instead of formatting whatever it is.

ivannaranjo: Also, as mentioned in my comment in `Resources.resx` use a culture specific formatting rather than specifying yourself, unless you want that formatting to be the same everywhere.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindow.cs
```diff
@@ -33,6 +34,11 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
     public class ErrorReportingToolWindow : ToolWindowPane
     {
         /// <summary>
+        /// Gets a <seealso cref="ErrorReportingViewModel"/> object that is associated with the Window.
+        /// </summary>
+        public ErrorReportingViewModel ViewModel => (Content as ErrorReportingToolWindowControl)?.DataContext as ErrorReportingViewModel;
```
ivannaranjo: Why not use the existing pattern of storing the view model and use it as the DAtaContext, etc? Instead of this complicated expression.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindow.cs
```diff
@@ -44,5 +50,16 @@ public ErrorReportingToolWindow() : base(null)
...
         }
+
+        /// <summary>
+        /// Create view mode when window object is created. 
+        /// </summary>
+        public override void OnToolWindowCreated()
+        {
+            base.OnToolWindowCreated();
+            (Content as ErrorReportingToolWindowControl).DataContext = new ErrorReportingViewModel();
+            CredentialsStore.Default.CurrentProjectIdChanged += (sender, e) => ViewModel.OnProjectIdChanged();
```
ivannaranjo: The view model should do this on its own, not outside. This is breaking the encapsulation and layering of the view model. The view model can do this on its constructor.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
```diff
@@ -17,20 +18,146 @@
...
+            </DataGrid>
+
+            <!-- Loading first page progress indicator -->
+            <controls:ProgressIndicator 
+                Grid.Row="2"
+                Panel.ZIndex="10"
+                Margin="0,60,0,0"
+                HorizontalAlignment="Center"                       
+                VerticalAlignment="Top" 
+                FullDuration="1000"
```
ivannaranjo: Not sure why you want a slower time duration here. It should be the same everywhere, that is the point of the animation.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,221 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+        private string _exceptionString;
+        private ObservableCollection<ErrorGroupItem> _groupStatsCollection;
+        private TimeRangeItem _selectedTimeRange;
+
+        /// <summary>
+        /// Gets an exception as string.
+        /// This is for some known possible errors. 
+        /// Mostly is for DataSourceException.
+        /// </summary>
+        public string ExceptionString
```
ivannaranjo: You should never use an exception's message as the source for UI unless there's nothing else you can use. You probably want to have a set of pre-made messages that you can use. Detect the error condition and use the right error message.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
ivannaranjo: Apparently this is not changing, will the value of the property change later?

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,221 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+        /// If the current project id is reset to null or empty, hide the grid. 
+        /// </summary>
+        public bool IsGridVisible => !String.IsNullOrWhiteSpace(CredentialsStore.Default.CurrentProjectId);
+
+        /// <summary>
+        /// Sets the currently selected time range.
+        /// </summary>
+        public TimeRangeItem SelectedTimeRangeItem
+        {
+            set { SetValueAndRaise(ref _selectedTimeRange, value); }
```
ivannaranjo: Where's the getter?

ivannaranjo: Even if it is a private one.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,221 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+            _dataSource = new Lazy<StackdriverErrorReportingDataSource>(CreateDataSource);
+            _groupStatsCollection = new ObservableCollection<ErrorGroupItem>();
+            GroupStatsView = new ListCollectionView(_groupStatsCollection);
+            PropertyChanged += OnPropertyChanged;
+        }
+
+        /// <summary>
+        /// Responds to current project id change event.
+        /// </summary>
+        public void OnProjectIdChanged()
```
ivannaranjo: This is what I was talking about before, this class should be adding the event handlers it needs, not relying on other classes to do it. Therefore this event handler should be private.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,221 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+            if (_nextPageToken == null)
+            {
+                Debug.WriteLine("_nextPageToken is null, there is no more events group to load");
+                return;
+            }
+
+            LoadAsync();
+        }
+
+        private void OnPropertyChanged(object sender, PropertyChangedEventArgs e)
```
ivannaranjo: This is waay overkill. You can do this very same thing on the setter for the `SelectedTimeRangeItem` property right after the call to `SetValueAndRaise()`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,221 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+            _nextPageToken = null;
+            LoadAsync();
+        }
+
+        /// <summary>
+        /// Load data from Google Cloud Error Reporting API service end point.
+        /// It shows a progress control when waiting for data.
+        /// In the end, if there is known type of exception, show the exception.
+        /// </summary>
+        private async Task LoadAsync()
```
ivannaranjo: If there's no point on awaiting for `LoadAsync()` to finish it is ok to return `void`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,221 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+                }
+                results = await _dataSource.Value?.GetPageOfGroupStatusAsync(
+                    _selectedTimeRange.GroupTimeRange,
+                    _selectedTimeRange.TimedCountDuration,
+                    nextPageToken: _nextPageToken);
+            }
+            catch (Exception ex) when (ex is DataSourceException || ex is ErrorReportingException)
+            {
+                ShowException = true;
+                ExceptionString = ex.ToString();
```
ivannaranjo: This is what I was talking about, use pre-made messages not the message coming from the exception. Those messages are not meant to be displayed to the user.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Compare the Service and Version fields of two <seealso cref="ServiceContext"/> objects.
+        /// Service must not be null.
+        /// Version is optional, can be null.
+        /// </summary>
+        public bool Equals(ServiceContext self, ServiceContext other)
+        {
+            if (self == null || other == null || self.Service == null || other.Service == null)
+            {
+                throw new ErrorReportingException(new ArgumentNullException());
```
ivannaranjo: Why? Just return false, or true if both are `null`. You should not throw from here.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+            if (string.CompareOrdinal(self.Service, other.Service) != 0)
+            {
+                return false;
+            }
+            if (self.Version == null || other.Version == null && self.Version != other.Version)
```
ivannaranjo: At this point you know that `self.Version` and `other.Version` is not `null`.

ivannaranjo: You are comparing them twice.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
```diff
@@ -0,0 +1,61 @@
...
+            return string.CompareOrdinal(self.Version, other.Version) == 0;
+        }
+
+        public int GetHashCode(ServiceContext obj)
+        {
+            if (obj == null)
+            {
+                throw new ErrorReportingException(new ArgumentNullException(nameof(obj)));
+            }
+            int serviceCode = obj.Service.GetHashCode();
```
ivannaranjo: Just return this, you don't really need to do anything special for hashes.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
```diff
@@ -0,0 +1,124 @@
...
+        private void timeRangeButton_Click(object sender, RoutedEventArgs e)
+        {
+            var button = sender as Button;
+            if (button == null)
+            {
+                Debug.WriteLine("timeRangeButton_Click, sender is not button. This is not expected. Code bug.");
+                return;
+            }
+
+            SelectedItem = button.DataContext as TimeRangeItem;
```
ivannaranjo: I think that what you really want is `button.Content`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
```diff
@@ -0,0 +1,124 @@
...
+{
+    /// <summary>
+    /// Interaction logic for TimeRangeButtons.xaml
+    /// </summary>
+    public partial class TimeRangeButtons : ItemsControl
+    {
+        /// <summary>
+        /// The list of <seealso cref="TimeRangeItem"/> as the data source of time range buttons.
+        /// </summary>
+        private readonly ObservableCollection<TimeRangeItem> TimeRangeItems =
```
ivannaranjo: This is private field, should be `_timeRangeItems`. It also looks like it could be a static field since you could share the collection amongst all instances.

Also why `ObservableCollection<T>` since you are not changing it?

ivannaranjo: You should be able to make it a `List<TimeRangeItem>` and then use the list initializer syntax:
```C#
new List<TimeRangeItem>
{
  new TimeRangeItem(...),
}
```

GoogleCloudExtension/GoogleCloudExtension/Controls/ProgressIndicator.xaml.cs
Deren-Liao: Moving slower, it's moving too fast like crazy on the error reporting window.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: Together to the comment above.
The type got from the ErrorGroup.FristSeenTime and LastSeenTime is "object". It's not datetime in compile time. 
In runtime, I noticed it is  datetime.
My solution is, if it is actually datetime, I format it.  If it is not, most likely, it is a string. My original code use Tostring.  I think I am going to attempt to convert it into datetime, if it fails, just show nothing.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: The input is from ErrorGroup member that is object.  I added comment and changed function name.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: No, it does not change.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindow.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindow.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
Deren-Liao: While it might be Okay for spinning this speed in the Explorer Tree View, in the Error Reporting window, it was too fast ,  makes me feel ... very bad .   I don't want to change how fast it spins at existing explorer tree view. So I decided to make it customizable.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: I'll use two pre-defined messages then.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: Added.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: async void?  or just void? 
I need async to await to catch data source API exception.   Then, I choose to use async void that follows the advice of [Avoid Async Void](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx)

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
Deren-Liao: I changed. I meant to say, null == null is valid.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
Deren-Liao: Even if there are differences in other fields, ignore them.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
Deren-Liao: Static won't work.   When it is shared with different UI, changing the CurrentSelected will change everywhere which is not desired behavior.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
Deren-Liao: What I need is not only the content. More important is the properties of time range.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: I agree. Removed it.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: Yes, I'll change the code. Normally it has more than one line. Still it's safe to check.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: No, it won't change.  Added comment: // TODO: add in subsequent PR.

Deren-Liao:
Updated, Please check again.

GoogleCloudExtension/GoogleCloudExtension/Controls/ProgressIndicator.xaml.cs
ivannaranjo: Is it different than the cloud explorer?

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
ivannaranjo: So it is set to an invalid command handler?

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
ivannaranjo: Is it because it is bigger?

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
ivannaranjo: The default implementation of GetHashCode works good enough, you don't really need to go beyond that, unless you are doing something more specific.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
ivannaranjo: Is starting to get a bit confusing to see where the data is coming from. Probably better using commands if you can use them.

GoogleCloudExtension/GoogleCloudExtension/Controls/ProgressIndicator.xaml.cs
Deren-Liao: Yes, I make it slower in Error Reporting window.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: Currently, it is null empty, I'll add code in next PR.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
Deren-Liao: No, it's not bigger.  I just feel it doesn't look good when it spins that fast.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
Deren-Liao: This is a customized comparison.  The ServiceContext contains more fields. The comparison cares only two fields :  Service and Version.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
Deren-Liao: Used a command.

Deren-Liao:
Hi Ivan, I changed the button click event to a Command.  Please look again.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
ivannaranjo: Then just remove it, it can be safely `null`.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
ivannaranjo: The thing is that the animation is designed to go at that speed, if it doesn't look good in your case then it doesn't look good in the cloud explorer either. We should then make it slower everywhere.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,217 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+            IsLoadingComplete = false;
+            ListGroupStatsResponse results = null;
+            ShowError = false;
+            IsRefreshing = _nextPageToken == null;
+            IsLoadingNextPage = _nextPageToken != null;
+            try
+            {
+                if (SelectedTimeRangeItem == null)
+                {
+                    throw new ErrorReportingException(new InvalidOperationException(nameof(SelectedTimeRangeItem)));
```
ivannaranjo: You are not actually catching this exception anywhere, why is that you need a special type of exception?

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
```diff
@@ -21,9 +33,217 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
...
+                    SelectedTimeRangeItem.GroupTimeRange,
+                    SelectedTimeRangeItem.TimedCountDuration,
+                    nextPageToken: _nextPageToken);
+            }
+            catch (DataSourceException)
+            {
+                ShowError = true;
+                ErrorString = Resources.ErrorReportingDataSourceGenericErrorMessage;
+            }
+            finally
```
ivannaranjo: Should probably also catch `ErrorReportingException` here, and other places where it is thrown, or you should just delete the type and use an existing exception.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,155 @@
...
+            return dt.ToLocalTime().ToString("F");
+        }
+
+        private string GetSeeIn()
+        {
+            if (ErrorGroup.AffectedServices == null)
+            {
+                return null;
+            }
+            var query = ErrorGroup.AffectedServices.Where(x => x.Service != null).Distinct(new ServiceContextComparer());
```
ivannaranjo: Why do you need a custom comparer? It seems like overkill to ensure that they are they same object. You can also do grouping by a particular property, etc... In short, implementing the `IEqualityComparer<T>` just for this seems too expensive.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml
```diff
@@ -0,0 +1,53 @@
...
+            <utils:VisibilityConverter x:Key="visibilityConverter" />
+
+        </ResourceDictionary>
+    </ItemsControl.Resources>
+    
+    <ItemsControl.ItemTemplate>
+        <DataTemplate>
+            <Grid>
+                <Button 
+                    Panel.ZIndex="10"
```
ivannaranjo: You don't need to specify the `Panel.ZIndex` property. By default `Grid` will use the order of the children.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml
```diff
@@ -0,0 +1,53 @@
+﻿<ItemsControl
```
ivannaranjo: You should do a custom control. This is not a good idea. Really only `UserControl` should be used this way. I know is possible to do this, but is not a good idea. What you really want is a custom control, and you should just do that. With a style to change its settings, template, etc...

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
```diff
@@ -0,0 +1,134 @@
...
+                        TimeRangeEnum.PERIOD1WEEK,
+                        EventTimeRangeEnum.PERIOD1WEEK),
+                    new TimeRangeItem(
+                        GoogleCloudExtension.Resources.ErrorReporting30DaysButtonCaption,
+                        $"{24 * 60 * 60}s",
+                        TimeRangeEnum.PERIOD30DAYS,
+                        EventTimeRangeEnum.PERIOD30DAYS)
+                });
+
+        public static readonly DependencyProperty SelectedItemProperty =
```
ivannaranjo: You should be using actually `Selector` as your base so you have the `SelectedItem` property defined for you in the right way.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// </summary>
+        public TimeRangeButtons()
+        {
+            InitializeComponent();
+        }
+
+        /// <summary>
+        /// Override the <seealso cref="OnApplyTemplate"/> method to initialize controls.
+        /// </summary>
+        public override void OnApplyTemplate()
```
ivannaranjo: This is the wrong approach, you should be defining simply a custom control with a style. You don't do this with code behind.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// If it is not DateTime type or the string is failed to convert to UTC time. 
+        /// </summary>
+        /// <returns>
+        /// Formated time string to Local Time, Local Culture, if input is DateTime type or
+        ///   the input is string and can be converted to UTC time.
+        /// return null otherwise.
+        /// </returns>
+        private static string FormatErrorGroupDateTime(object datetime)
+        {
+            DateTime dt = DateTime.MinValue;
```
Deren-Liao: Yes, but the compiler complains it.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,155 @@
...
+            if (datetime is DateTime)
+            {
+                dt = (DateTime)datetime;
+            }
+            else if (datetime is string)
+            {
+                if (!DateTime.TryParse(datetime as string,
+                    CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out dt))
+                {
+                    return null;
```
Deren-Liao: The UI will then show empty.  Showing "no timestamp found" does  not help anything and it looks odd.

GoogleCloudExtension/GoogleCloudExtension/Controls/ProgressIndicator.xaml.cs
Deren-Liao: I am going to remove this customization.  Spinning a bit fast "subjectively" is Okay for me :).

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1682,4 +1682,56 @@ Open the project of version {2} to properly locate the logging writter location.
     <value>Environment</value>
     <comment>The display name for the Environment property.</comment>
   </data>
+  <data name="ErrorReportingCurrentGroupTimePeriodLabelFormat" xml:space="preserve">
+    <value>Errors in the last {0}</value>
+    <comment>In the error reporting overview window, the current selected time period label text format.</comment>
```
Deren-Liao: This is the period of times.   the value already contains plural.   I'll add comment.

Deren-Liao: Added comment

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: Removed

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <summary>
+        /// The error message displayed in data grid row.
+        /// </summary>
+        public string Error => ErrorGroup.Representative.Message;
+
+        /// <summary>
+        /// Show service context. 
+        /// <seealso cref="ErrorGroupStats.AffectedServices"/>.
+        /// </summary>
+        public string SeenIn => GetSeeIn();
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,155 @@
...
+            if (lines != null)
+            {
+                Message = lines.Count() > 0 ? lines[0] : null;
+                FirstStackFrame = lines.Count() > 1 ? lines[1] : null;
+            }
+            OnNavigateToDetailCommand = new ProtectedCommand(null);     // TODO: add handler in subsequent PR.
+        }
+
+        /// <summary>
+        /// The input in compiling time is object.
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
Deren-Liao: I thought of a better way.  Removed it.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorGroupItem.cs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        private string GetSeeIn()
+        {
+            if (ErrorGroup.AffectedServices == null)
+            {
+                return null;
+            }
+            var query = ErrorGroup.AffectedServices.Where(x => x.Service != null).Distinct(new ServiceContextComparer());
+            return String.Join(Environment.NewLine, query.Select(x => FormatServiceContext(x)));
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindow.cs
```diff
@@ -32,6 +33,13 @@ namespace GoogleCloudExtension.StackdriverErrorReporting
     [Guid("4b3c62b4-2121-40a1-8cd5-8f794760b35e")]
     public class ErrorReportingToolWindow : ToolWindowPane
     {
+        private readonly ErrorReportingViewModel _viewModel;
+
+        /// <summary>
+        /// Gets a <seealso cref="ErrorReportingViewModel"/> object that is associated with the Window.
+        /// </summary>
+        public ErrorReportingViewModel ViewModel => _viewModel;
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
```diff
@@ -17,20 +18,146 @@
...
+            <Grid 
+                Grid.Row="1"
+                Margin="0,12,0,12" 
+                IsEnabled="{Binding IsLoadingComplete}">
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="*" />
+                    <ColumnDefinition Width="Auto"/>
+                </Grid.ColumnDefinitions>
+                <WrapPanel 
+                    Grid.Column="0"
```
Deren-Liao: You are right.  Removed.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
Deren-Liao: I changed back to default behavior.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingToolWindowControl.xaml
```diff
@@ -17,20 +18,146 @@
...
+                Grid.Row="3"
+                Padding="12"
+                Panel.ZIndex="10"
+                Margin="0,-60,0,0"
+                HorizontalAlignment="Center"                       
+                FullDuration="1000"
+                Visibility="{Binding IsLoadingNextPage, Converter={StaticResource visibilityConverter}}" />
+
+            <TextBox 
+                Grid.Row="2"
```
Deren-Liao: Change.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: yes, I need to catch this below.

Deren-Liao: Added catch.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ErrorReportingViewModel.cs
Deren-Liao: Yes, added.  I had catched it and missed it at last code refactoring.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/ServiceContextComparer.cs
Deren-Liao: removed.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml
Deren-Liao: Done.  
A lot of cases, I feel using UserControl makes writing code easier.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
Deren-Liao: That's great. Used Selector.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.xaml.cs
Deren-Liao: Changed.

Deren-Liao:
Hi Ivan, I made the change. Please take a look again.

Deren-Liao:
Hi Ivan, may you take a quick look?

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1684,7 +1684,7 @@ Open the project of version {2} to properly locate the logging writter location.
   </data>
   <data name="ErrorReportingCurrentGroupTimePeriodLabelFormat" xml:space="preserve">
     <value>Errors in the last {0}</value>
-    <comment>In the error reporting overview window, the current selected time period label text format.</comment>
+    <comment>In the error reporting overview window, the current selected time period label text format.  {0} is the time period.  Example:  1 day,  30 days etc.</comment>
```
ivannaranjo: I think you will need a plural vs. singlar version of this string. I will look weird if it says `Error in the last 1 day`.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -1734,4 +1734,8 @@ Open the project of version {2} to properly locate the logging writter location.
     <value>Failed to retrieve data from cloud server. Please try later.</value>
     <comment>Generic error message for Error Reporting when there is DataSource excepiton.</comment>
   </data>
+  <data name="ErrorReportingInternalCodeErrorGenericMessage" xml:space="preserve">
+    <value>Unkown internal error. Please report the issue and/or retry later.</value>
```
ivannaranjo: The string is good for now, but we will go through the string text review later.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// User can click on button to select/change the time range.
+    /// </summary>
+    public class TimeRangeButtons : Selector
+    {
+        public static readonly DependencyProperty OnItemSelectedCommandProperty =
+            DependencyProperty.Register(
+                nameof(OnItemSelectedCommand),
+                typeof(ICommand),
+                typeof(TimeRangeButtons),
+                new FrameworkPropertyMetadata(null));
```
ivannaranjo: You are not executing the command, I guess that code will come in later?

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtonsResources.xaml
```diff
@@ -0,0 +1,59 @@
...
+                            X2="55" 
+                            Stroke="Blue"    
+                            StrokeThickness="2"
+                            Visibility="{Binding IsCurrentSelection, Converter={StaticResource visibilityConverter}}"
+                            VerticalAlignment="Bottom"/>
+                    </Grid>
+                </DataTemplate>
+            </Setter.Value>
+        </Setter>
+        <Setter Property="ItemsPanel">
```
ivannaranjo: You are setting the `ItemsPanel` property twice in the style.

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtons.cs
Deren-Liao: Line 52,  I set SelectedItem = item

GoogleCloudExtension/GoogleCloudExtension/StackdriverErrorReporting/TimeRangeButtons/TimeRangeButtonsResources.xaml
Deren-Liao: Changed.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Sure.  I hope we can have better descriptive text for this.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: I created issue #497 to keep track of this. I'll make the change once I understand well of your comment.

PR #464, opened 2017-02-28
Adding aspnetcore templates

ivannaranjo:
This PR adds the new ASP.NET Core templates with the logging, tracing and error reporting configured for the user.


GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreWebApi/Startup.cs
```diff
@@ -0,0 +1,78 @@
...
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+        {
+            var projectId = GetProjectId();
+
+            loggerFactory.AddConsole(Configuration.GetSection("Logging"));
+            loggerFactory.AddGoogle(projectId);
+            loggerFactory.AddDebug();
+
+            if (env.IsDevelopment())
```
ivannaranjo: Ah, miss read this.

ivannaranjo:
Also, I'm going to send an updated version in a little bit that works in both Gce and Gae. The current version won't deploy to Gae since we only check for project ID in Gce.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreWebApi/Startup.cs
```diff
@@ -64,7 +64,7 @@ private string GetProjectId()
             }
             if (projectId == null)
             {
-                projectId = Google.Api.Gax.Platform.Instance().GceDetails?.ProjectId;
+                projectId = Google.Api.Gax.Platform.Instance().GceDetails?.ProjectId ?? Google.Api.Gax.Platform.Instance().GaeDetails?.ProjectId;
```
iantalarico: The call to Instance can take awhile I would suggest:

var instance = Google.Api.Gax.Platform.Instance().
projectId = instance.GceDetails?.ProjectId ??  instance.GaeDetails?.ProjectId

ivannaranjo: Oh I see, is not catched :(. Ok will do.

PR #470, opened 2017-03-01
Allow "runtime: custom" deployments from VS.

### LINKED ISSUE #465: Should check if app.yaml is a custom runtime.
If the user is specifying a custom runtime then the extension should also copy the existing Dockerfile. The current behavior is to assume that the app.yaml file only refers to an "aspnetcore" runtime.

ivannaranjo:
This change allows back specifying "runtime: custom" in the app.yaml in case the user needs/wants to have a custom Dockerfile for whatever reason.

This PR also adds logic to **not** force the runtime builder if the runtime is not `aspnetcore`, as the only supported runtime is that. This is temporary logic which will go away once the runtime builders are fully released anyway.

Fixes #465


ivannaranjo:
// CC @jskeet in case you want to use the `runtime: custom` et all.

PR #473, opened 2017-03-01
Adds check for Cloud SDK version.

### LINKED ISSUE #471: Add check for minimum gcloud version needed.
The extension should check for the minimum version of gcloud needed before attempting to use it.

ivannaranjo:
This PR adds a check for the version of the Cloud SDK installed in the system so the extension can ensure that the minimal version with the required features is present in the system. If the user has an older version of the Cloud SDK installed a message will show the user the need to update.

Fixes #471

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudValidationResult.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudValidationResult.cs
@@ -1,4 +1,18 @@
-﻿using System;
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
 
 namespace GoogleCloudExtension.GCloud
 {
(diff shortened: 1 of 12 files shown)


GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudValidationResult.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// </summary>
+        public bool IsCloudSdkInstalled { get; }
+
+        /// <summary>
+        /// Whether the Cloud SDK is newer or equal than the required version.
+        /// </summary>
+        public bool IsCloudSdkUpdated { get; }
+
+        /// <summary>
+        /// If a required component was detected as installed or not.
```
Deren-Liao: Do you actually mean "if ALL required component ... " ?

GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudWrapper.cs
```diff
@@ -163,32 +170,48 @@ public static Task<bool> BuildContainerAsync(string buildFilePath, string conten
...
+            {
+                return new GCloudValidationResult(isCloudSdkInstalled: false);
+            }
 
-        /// <summary>
-        /// Returns true if the methods concerning kubectl and GKE can be used safely.
-        /// </summary>
-        /// <returns>A task that will be fullfilled to true if the GKE methods can be used.</returns>
-        public static Task<bool> CanUseGKEAsync() => IsComponentInstalledAsync("kubectl");
+            var cloudSdkVersion = await GetInstalledCloudSdkVersionAsync();
```
Deren-Liao: inside GetInstalledCloudSdkVersionAsync, it checked  IsGCloudCliInstalled  too.   
You might change the code to something like:
 var cloudSdkVersion = await GetInstalledCloudSdkVersionAsync();
if (cloudSdkVersion == null) {
       return new GCloudValidationResult(isCloudSdkInstalled: false);
}

GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudValidationResult.cs
ivannaranjo: No I do actually mean single component. The component that was passed to the `GCloudWrapper.ValidateGCloudAsync()`.

PR #488, opened 2017-03-03
Requiring latest version of gcloud, removing workaround.

### LINKED ISSUE #484: Update the required version of Cloud SDK once it is updated.
Since we are going to require the latest version of the Cloud SDK, update the extension to use the latest version as soon as possible.

Also stop using the `CLOUDSDK_APP_USE_RUNTIME_BUILDERS` and `CLOUDSDK_APP_RUNTIME_BUILDERS_ROOT` environment variables since `gcloud` will have the right behavior built in.

ivannaranjo:
This PR adds the final dependency on the Cloud SDK version 146.0.0. Because that version of the Cloud SDK natively support deploying "aspnetcore" apps we no longer need to set the environment variables to force the deployment.

Fixes #484


ivannaranjo:
// cc @jskeet, in case you want to try this build.

PR #512, opened 2017-03-20
Streaming logs, auto reload error reports.

Deren-Liao:
Use a dispatcher timer to do auto reload.
Logs viewer is a bit more complex.
(1) if it is ascending order, append to end. 
(2) if it is descending order,  add to the start of the list.


GoogleCloudExtension/GoogleCloudExtension/Controls/AutoReloadButton.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        /// <summary>
+        /// Initialize the control template.
+        /// </summary>
+        public override void OnApplyTemplate()
+        {
+            //  DispatcherTimer setup
+            _timer.Interval = TimeSpan.FromSeconds(IntervalSeconds);
+            _timer.Tick += (sender, e) => ExecuteAutoReloadCommand();
```
iantalarico: I'm not very familiar with "OnApplyTemplate" and "ApplyTemplate" but can this be called multiple times for each AutoReloadButton?  If so should this code happen in the constructor?

GoogleCloudExtension/GoogleCloudExtension/Controls/AutoReloadButton.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        private static void OnIntervalSecondsChanged(DependencyObject source, DependencyPropertyChangedEventArgs e)
+        {
+            int newIntervalSeconds = (int)e.NewValue;
```
iantalarico: Should there be checks around what this value can be?

GoogleCloudExtension/GoogleCloudExtension/Controls/AutoReloadButton.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// When it is enabled, the button uses a timer to trigger reload event.
+    /// </summary>
+    public class AutoReloadButton : ImageToggleButton
+    {
+        // TODO: proper dispose of the timer.
+        private readonly DispatcherTimer _timer;
+
+        static AutoReloadButton()
+        {
+            ToggleButton.IsCheckedProperty.OverrideMetadata(
```
iantalarico: I'm not very familiar with static constructors but shouldn't this be done per instance?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -111,23 +112,41 @@
             <RowDefinition Height="Auto" />
             <RowDefinition Height="*" />
         </Grid.RowDefinitions>
+
```
iantalarico: nit: extra whitespace

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
```diff
@@ -111,23 +112,41 @@
...
         <tb:TitleBar />
+
```
iantalarico: nit: extra whitespace

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -276,6 +278,16 @@ public bool IsControlEnabled
         }
 
         /// <summary>
+        /// Gets the command
```
iantalarico: Partial comment?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -400,17 +413,58 @@ private void OnRefreshCommand()
         /// <summary>
         /// Append a set of log entries.
         /// </summary>
-        private void AddLogs(IList<LogEntry> logEntries)
+        private void AddLogs(IList<LogEntry> logEntries, bool autoReload = false)
```
iantalarico: missing param comment?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -400,17 +413,58 @@ private void OnRefreshCommand()
         /// <summary>
         /// Append a set of log entries.
```
iantalarico: add comment about truncating the log list

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -400,17 +413,58 @@ private void OnRefreshCommand()
...
 
-            foreach (var log in logEntries)
+            // Pick the last n items only in case there are too much that makes the viewer too slow.
+            var query = logEntries.Skip(Math.Max(0, logEntries.Count() - MaxLogItems)).Select(x => {
+                    var item = new LogItem(x, SelectedTimeZone);
+                    _lastLogTime = DateTimeUtils.Max(_lastLogTime, item.TimeStamp.ToUniversalTime());
+                    return item;
+                });
+
+            if (autoReload && DateTimePickerModel.IsDescendingOrder)
```
iantalarico: Why does adding log entries need to know about auto reload? If it's IsDescendingOrder don't you always want to insert at the front?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -400,17 +413,58 @@ private void OnRefreshCommand()
...
             {
-                _logs.Add(new LogItem(log));
+                foreach (var item in query.Reverse())
+                {
+                    Debug.WriteLine($"add entry {item.Entry.Timestamp}");
+                    _logs.Insert(0, item);
```
iantalarico: Is this going to be expensive adding at the beginning of the list each time?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
iantalarico: Can we add unit tests for this?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -400,17 +413,58 @@ private void OnRefreshCommand()
...
+                    _logs.Add(item);
+                }
+            }
+
+            RemoveExceedingLimitEntries();
+        }
+
+        /// If the total count of log entries exceeds <seealso cref="MaxLogItems"/>,
+        /// Remove the older log entries.
+        private void RemoveExceedingLimitEntries()
```
iantalarico: Can we add unit tests for this?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -501,7 +557,7 @@ private void SetServerRequestCompleteStatus()
         /// On complex filters, scanning through logs take time. The server returns empty results 
         ///   with a next page token. Continue to send request till some logs are found.
         /// </summary>
-        private async Task LoadLogsAsync(CancellationToken cancellationToken)
+        private async Task LoadLogsAsync(CancellationToken cancellationToken, bool autoReload = false)
```
iantalarico: param comment?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -36,6 +36,8 @@ public class LogsViewerViewModel : ViewModelBase
     {
         private const string AdvancedHelpLink = "https://cloud.google.com/logging/docs/view/advanced_filters";
         private const int DefaultPageSize = 100;
+        private const int MaxLogItems = 500;
```
iantalarico: Why 500? This seems small for log entries.

Can we possibly have more and render them as the user scrolls to them?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
```diff
@@ -714,7 +771,7 @@ private void OnTimeZoneChanged()
         /// <summary>
         /// Aggregate all selections into filter string.
         /// </summary>
-        private string ComposeSimpleFilters()
+        private string ComposeSimpleFilters(bool ignoreTimeStamp = false)
```
iantalarico: param comment?

GoogleCloudExtension/GoogleCloudExtension/Utils/DateTimeUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// Helper functions for <seealso cref="DateTime"/>.
+    /// </summary>
+    public static class DateTimeUtils
+    {
+        public static DateTime Min(DateTime x, DateTime y)
+        {
+            return (x.ToUniversalTime() < y.ToUniversalTime()) ? x : y;
+        }
+
+        public static DateTime Max(DateTime x, DateTime y)
```
iantalarico: comments? I know it's a simple class but it would be nice

iantalarico: You can do:

public static DateTime Max(DateTime x, DateTime y) =>(x.ToUniversalTime() > y.ToUniversalTime()) ? x : y;

GoogleCloudExtension/GoogleCloudExtension/Utils/DateTimeUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.Utils
+{
+    /// <summary>
+    /// Helper functions for <seealso cref="DateTime"/>.
+    /// </summary>
+    public static class DateTimeUtils
+    {
+        public static DateTime Min(DateTime x, DateTime y)
```
iantalarico: I don't see this used anywhere.  Maybe delete it until you need it?

GoogleCloudExtension/GoogleCloudExtensionUnitTests/DateTimeUtilTess.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class DateTimeUtilTess
+    {
+        [TestMethod]
+        public void MinTest()
+        {
+            DateTime t1 = DateTime.Now;
+            DateTime t2 = DateTime.Now;
+            Assert.AreEqual(t1, Min(t1, DateTime.UtcNow));
+            Assert.AreEqual(t1, Min(DateTime.UtcNow, t1));
+            Assert.AreEqual(t1, Min(t1, t2));
```
iantalarico: Are you sure this will always be true can these both be the same time in some cases?

GoogleCloudExtension/GoogleCloudExtensionUnitTests/DateTimeUtilTess.cs
```diff
@@ -0,0 +1,50 @@
...
+            Assert.AreEqual(t2, Min(t2, DateTime.MaxValue));
+            Assert.AreEqual(DateTime.MinValue, Min(t2, DateTime.MinValue));
+        }
+
+        [TestMethod]
+        public void MaxTest()
+        {
+            DateTime t1 = DateTime.Now;
+            DateTime t2 = DateTime.UtcNow;
```
iantalarico: This is utcnow while the other is not?

GoogleCloudExtension/GoogleCloudExtensionUnitTests/DateTimeUtilTess.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    [TestClass]
+    public class DateTimeUtilTess
+    {
+        [TestMethod]
+        public void MinTest()
+        {
+            DateTime t1 = DateTime.Now;
+            DateTime t2 = DateTime.Now;
+            Assert.AreEqual(t1, Min(t1, DateTime.UtcNow));
```
iantalarico: you mix between now and utcnow?

GoogleCloudExtension/GoogleCloudExtensionUnitTests/StringUtilTests.cs
```diff
@@ -42,7 +56,7 @@ public void LastNonSpaceIndexTests()
             Assert.AreEqual(-1, LastNonSpaceIndex("     "));
             Assert.AreEqual(-1, LastNonSpaceIndex(""));
             Assert.AreEqual(-1, LastNonSpaceIndex(null));
-            Assert.AreEqual(8, FirstNonSpaceIndex("   uu  pp  "));
+            Assert.AreEqual(8, LastNonSpaceIndex("   uu  pp  "));
```
iantalarico: This change seems out of place

GoogleCloudExtension/GoogleCloudExtension/Controls/AutoReloadButton.cs
Deren-Liao: The dependency property of WPF control is set globally. In the end, the event is wired to the instance method.  This is how WPF control works.   Refer to https://msdn.microsoft.com/en-us/library/ms754209(v=vs.110).aspx

GoogleCloudExtension/GoogleCloudExtension/Controls/AutoReloadButton.cs
Deren-Liao: The OnApplyTemplate is called only once.  It is the place to add custom initialization after the element visual tree is constructed.  In this particular case,  timer only makes sense after the elements are loaded.  Putting it into constructor may work too, although I doubt if the properties like IntervalSeconds are initialized in that case.  
I mean there is no point to move to constructor while putting it here is safe.

GoogleCloudExtension/GoogleCloudExtension/Controls/AutoReloadButton.cs
Deren-Liao: OnIntervalSecondsChanged  is only called by the WPF framework when the IntervalSeconds property value changes.     The value type is int in this case, it won't be other type. Adding a check would be redundant.

Deren-Liao: Added comment too.   Again, this is a WPF pattern, kind of strange too.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: Removed.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerToolWindowControl.xaml
Deren-Liao: Removed.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: I picked the number by my experience.  
When it has too much, the viewer becomes too slow.    And if doing streaming, we don't want to grow the count indefinitely. 

And yes, when user scroll, new items will be added. In the mean time the oldest items are removed from the view.   

I'd keep the 500 for now and change later if other numbers make more sense.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: Fixed.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: I removed the logic for now. I'll add limiting log count later with a more mature solution.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: Added.

Deren-Liao: I will consider in the future when there is bulk amount of time.  Adding unit test to WPF model could be challenging to do all the mock and most times it does not help catching bugs because they are more UI related.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: In most case, logs are appended to the end. Only if it is descending order and auto reload, log entries are appended to the beginning.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: Either way, adding to the end or the beginning, both are expensive for every time it refreshes the UI. 
The alternative is to make the entire list and switch it to the view.  In that approach, the UI looks pending for a long time and suddenly completely replace the view.  The scroll position changes etc.  The user experience is bad.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: I removed the code.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: Added.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogsViewerViewModel.cs
Deren-Liao: Added.

GoogleCloudExtension/GoogleCloudExtension/Utils/DateTimeUtils.cs
Deren-Liao: Deleted.

GoogleCloudExtension/GoogleCloudExtension/Utils/DateTimeUtils.cs
Deren-Liao: I added a comment.

Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/DateTimeUtilTess.cs
Deren-Liao: yes. Purposely added this test.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/DateTimeUtilTess.cs
Deren-Liao: I made minor change to guarantee they are different.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/DateTimeUtilTess.cs
Deren-Liao: I purposely add it .

GoogleCloudExtension/GoogleCloudExtensionUnitTests/StringUtilTests.cs
Deren-Liao: It's a bug from previous change.  Just fixed it here.

Deren-Liao:
Hi Ian,
Thanks for your code review comment.  I made changes on your feedback.   Adding unit tests to Model is challenging -- mocking the dependencies, check the "expected results" etc.  It's currently not the practice we do although in general it's always better to add more unit tests.

GoogleCloudExtension/GoogleCloudExtension/Controls/AutoReloadButton.cs
iantalarico: Sorry I should have been more clear. I didn't mean about the type I mean about the value.  -10 or 0?

Deren-Liao:
Thanks for the review!

PR #520, opened 2017-03-27
Pubsub feature

ILMTitan:
Merge some old work for browsing, creating and deleting PubSub topics and subscriptions.


ILMTitan:
The set of commits up to 72697f9 have already been reviewed. cf087d3 is just a straight merge, and 4e2cc02 contains a bit of original work to get the pub sub files compatible with changes made to the master branch.

.gitignore
```diff
@@ -10,4 +10,4 @@
 .DS_Store
 **/packages/*
 **/*.pyc
-
+/GoogleCloudExtension/GoogleCloudExtension.DataSources/GoogleCloudExtension.DataSources.csproj.user
```
Deren-Liao: I support the idea of excluding all .user files.   So,  I suggest to removing this change from this PR, add a excluding rule of  *.csproj.user  in a separate PR.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,172 @@
+﻿// Copyright 2016 Google Inc. All Rights Reserved.
```
Deren-Liao: Change to 2017.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    /// <summary>
+    /// Data source that returns information about Pubsub resources.
+    /// </summary>
+    public class PubsubDataSource : DataSourceBase<PubsubService>
+    {
+        private string ProjectResourceName => $"projects/{ProjectId}";
```
Deren-Liao: Use ProjectResourceName from  DataSourceBase.

GoogleCloudExtension/GoogleCloudExtension.Utils/Model.cs
```diff
@@ -44,7 +44,7 @@ protected void SetValueAndRaise<T>(ref T storage, T value, [CallerMemberName] st
         /// </summary>
         /// <param name="propertyName">The name of the property that is changing. If null then the event will signify
         /// that all properties in the object changed.</param>
-        protected void RaisePropertyChanged(string propertyName)
+        protected void RaisePropertyChanged([CallerMemberName] string propertyName = "")
```
Deren-Liao: I searched the code base.  In most cases,  the propertyName is different than the caller member name.   Making it default to [callerMemberName] may introduce bugs more than be more convenient to use.
When the caller name is same, our pattern is to use SetValueAndRaise.

.gitignore
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,175 @@
...
+        public override TreeLeaf NoItemsPlaceholder => s_noItemsPlacehoder;
+        public override TreeLeaf LoadingPlaceholder => s_loadingPlaceholder;
+
+        /// <summary>
+        /// The list of all visible subscriptions of the current project.
+        /// </summary>
+        public IList<Subscription> Subscriptions { get; private set; }
+
+        internal PubsubDataSource DataSource => _dataSource.Value;
+        private Lazy<PubsubDataSource> _dataSource = new Lazy<PubsubDataSource>(CreateDataSource);
```
Deren-Liao: Code style . Move private before public.  I don't recall the rule of internal.   I assume it's find to leave it here.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        /// <summary>
```
Deren-Liao: Code style.  Move private after public methods.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        /// <summary>
+        /// Gets the subscriptions of this topic.
+        /// </summary>
+        private IEnumerable<Subscription> ListSubscriptionViewModels()
+        {
+            return _owner.Subscriptions.Where(subscription => subscription.Topic == _topicItem.FullName);
+        }
+
+        public virtual void Refresh()
```
Deren-Liao: Code style. Move public method before private method.

GoogleCloudExtension/GoogleCloudExtension/GoogleCloudExtension.csproj.user
```diff
@@ -11,6 +11,6 @@
     <StartArguments>/rootsuffix Exp</StartArguments>
   </PropertyGroup>
   <PropertyGroup>
-    <ProjectView>ProjectFiles</ProjectView>
+    <ProjectView>ShowAllFiles</ProjectView>
```
Deren-Liao: Change your .user file back not to ShowAllFiles and save .  Does not seem to be right to set it default to show all.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -95,9 +95,18 @@ public TopicViewModel(PubsubSourceRootViewModel owner, Topic topic)
             }
             catch (DataSourceException e)
             {
-                Debug.Write(e.Message, "New Subscription");
+                string message;
+                if (string.IsNullOrEmpty(e.Message))
```
Deren-Liao: Just be curious,  in what situation you would see null or empty message?

GoogleCloudExtension/GoogleCloudExtension.Utils/Model.cs
ILMTitan: I use this in NewSubscriptionViewModel.Push. I can't use `SetValueAndRaise` there because the value I am setting is a property on another class (which can't be sent as a ref parameter). However, I see now it causes some resharper warnings everywhere (Explicit argument passed to parameter with caller info attribute), so I am chainging it back.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewTopicWindowContent.xaml
```diff
@@ -10,14 +10,40 @@
...
+        </ResourceDictionary>
+    </UserControl.Resources>
+    <theming:CommonDialogWindowBaseContent HasBanner="True">
+        <theming:CommonDialogWindowBaseContent.Buttons>
+            <theming:DialogButtonInfo Command="{Binding CreateCommand}"
+                                      IsDefault="True"
+                                      Caption="{x:Static ext:Resources.UiOkButtonCaption}"/>
+            <theming:DialogButtonInfo IsCancel="True" Caption="{x:Static ext:Resources.UiCancelButtonCaption}"/>
+        </theming:CommonDialogWindowBaseContent.Buttons>
+        <StackPanel Orientation="Vertical" HorizontalAlignment="Stretch" VerticalAlignment="Center">
```
Deren-Liao: Style:  put every property in a separate line.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ILMTitan: Done for new files edited this year.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
         </theming:CommonDialogWindowBaseContent.Buttons>
+
+        <StackPanel HorizontalAlignment="Stretch" VerticalAlignment="Center">
+            <TextBlock HorizontalAlignment="Center" MaxWidth="300" Margin="5,5,5,5"
+                   Text="{Binding TopicName, StringFormat={x:Static ext:Resources.NewSubscriptionWindowHeader}}"/>
+            <DockPanel>
+                <Label DockPanel.Dock="Left" Target="{Binding ElementName=SubscriptionName}" VerticalAlignment="Center"
+                   Content="{x:Static ext:Resources.NewSubscriptionWindowNameLabel}"/>
+                <TextBox Name="SubscriptionName" MinWidth="200" MaxWidth="500" ClipToBounds="True"
+                    ToolTip="{Binding RelativeSource={RelativeSource Self}, Path=(Validation.Errors)[0].ErrorContent}">
```
Deren-Liao: We have common tooltip style.  Search for CommonTooltipStyle, and use that.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
+            </DockPanel>
+            <TextBlock Margin="5" Text="{x:Static ext:Resources.NewSubscriptionWindowDeliveryType}"/>
+            <RadioButton GroupName="PushPull" Margin="5" IsChecked="True"
+                         Content="{x:Static ext:Resources.NewSubscriptionWindowPullLabel}"/>
+            <DockPanel>
+                <RadioButton DockPanel.Dock="Left" Name="PushRadioButton" GroupName="PushPull"
+                             IsChecked="{Binding Push}" Margin="5"
+                             Content="{x:Static ext:Resources.NewSubscriptionWindowPushLabel}"
+                             VerticalAlignment="Center"/>
+                <TextBox Name="PushUrlBox" MinWidth="200" MaxWidth="500" Text="{Binding PushConfig.PushEndpoint}"
```
Deren-Liao: Add  UpdateSourceTrigger=PropertyChanged  too?

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
-                   Content="{x:Static ext:Resources.NewSubscriptionWindowAckDeadlineLabel}"/>
-            <TextBox Name ="Deadline" Width="40"
-                     Text="{Binding Subscription.AckDeadlineSeconds, TargetNullValue=60, UpdateSourceTrigger=PropertyChanged}"/>
-        </DockPanel>
-    </StackPanel>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="../Theming/CommonResources.xaml" />
+            </ResourceDictionary.MergedDictionaries>
+            <Style BasedOn="{StaticResource CommonTextStyle}" TargetType="{x:Type TextBlock}">
```
Deren-Liao: I would ask Ivan to confirm if this is going to cause problem to set style to all elements of the type in this user control. 
Normally, we use x:key="xxxStyle",  and apply the style to the element. 
When setting style like this without x:key , it's universal style. All elements of the type in the control gets this style by default.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ILMTitan: I am passing along GoogleApiException.Error?.Message because it can be a useful error message. If Error is null, the message in the DataSourceException will be empty, signaling the view model to default to the non-helpful error message in the resources file.
Something else I considered doing is adding a Google.Api.Requests.RequestError Error property to DataSourceException, but that seems too heavy for this single use case.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
+                <Setter Property="VerticalAlignment" Value="Center"/>
+            </Style>
+            <Style BasedOn="{StaticResource CommonTextBoxStyle}" TargetType="{x:Type TextBox}">
+                <Setter Property="VerticalAlignment" Value="Center"/>
+                <Setter Property="Margin" Value="3,3,3,3"/>
+            </Style>
+            <Style BasedOn="{StaticResource LabeledRadioButtonStyle}" TargetType="{x:Type RadioButton}">
+                <Setter Property="VerticalAlignment" Value="Center"/>
+            </Style>
+            <Style BasedOn="{StaticResource CommonDialogStyleBase}" TargetType="{x:Type UserControl}"/>
```
Deren-Liao: Following the existing pattern of something like the following. 
    <UserControl.Style>
        <Binding Source="{StaticResource CommonDialogStyleBase}" />
    </UserControl.Style>

This says, for this user control, use this style.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewTopicWindowContent.xaml
```diff
@@ -10,14 +10,40 @@
...
         FocusManager.FocusedElement="{Binding ElementName=TopicName}">
     <UserControl.Resources>
-        <StackPanel Orientation="Vertical" HorizontalAlignment="Stretch" VerticalAlignment="Center" x:Key="Content">
-            <TextBlock HorizontalAlignment="Center"
-                       Text="{Binding Project, StringFormat={x:Static ext:Resources.NewTopicWindowTitle}}"/>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="../Theming/CommonResources.xaml" />
+            </ResourceDictionary.MergedDictionaries>
+            <Style BasedOn="{StaticResource CommonTextStyle}" TargetType="{x:Type TextBlock}">
```
Deren-Liao: Same as above, I am worried of setting global style this way.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/GoogleCloudExtension.csproj.user
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewTopicWindowContent.xaml
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: Do I care if the property on the model is updated while it is being changed? Because I am building and then returning the model, I think I should only care about it having the correct value at the end, right?

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: It being universal is intentional (where the universe is the dialog). I would prefer not to have to put `Style=` on every element.

ILMTitan:
Waiting on Ivann for the global style judgement, but other changes made.
PTAL

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewTopicWindowContent.xaml
ivannaranjo: Yeah I agree, it is better to be explicit. The `TargetType` is better suited for styles for controls and that sort of thing.

.gitignore
ivannaranjo: Please don't remove the .user files just yet. If you do you will break debugging the extension, sadly the process to start when starting debugging is stored in the .user file.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ivannaranjo: Please don't show error messages coming from exceptions. In general those error messages are not as helpful as you think and will never be able to be localized. We need to detect what the error is and show the specific error message.

We have been moving away from displaying exception messages everywhere we can, and show our own messages, where we can control the wording and localize them in the future.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -95,9 +95,18 @@ public TopicViewModel(PubsubSourceRootViewModel owner, Topic topic)
...
+                    message = Resources.PubSubNewSubscriptionErrorMessage;
+                }
+                else
+                {
+                    message = e.Message;
+                }
+                Debug.Write(message, "New Subscription");
                 UserPromptUtils.ErrorPrompt(
-                    Resources.PubSubNewSubscriptionErrorMessage, Resources.PubSubNewSubscriptionErrorHeader);
+                    message, Resources.PubSubNewSubscriptionErrorHeader);
```
ivannaranjo: Same here, please show the message from the resources.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ivannaranjo: We prefer to put Style in every element, be explicit on what elements look like.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
             <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiCancelButtonCaption}" IsCancel="True"/>
         </theming:CommonDialogWindowBaseContent.Buttons>
+
+        <StackPanel HorizontalAlignment="Stretch" VerticalAlignment="Center">
+            <TextBlock HorizontalAlignment="Center" MaxWidth="300" Margin="5,5,5,5"
+                   Text="{Binding TopicName, StringFormat={x:Static ext:Resources.NewSubscriptionWindowHeader}}"/>
+            <DockPanel>
+                <Label DockPanel.Dock="Left" Target="{Binding ElementName=SubscriptionName}" VerticalAlignment="Center"
+                   Content="{x:Static ext:Resources.NewSubscriptionWindowNameLabel}"/>
+                <TextBox Name="SubscriptionName" MinWidth="200" MaxWidth="500" ClipToBounds="True"
```
ivannaranjo: Why the `MinWidth` and `MaxWidth`, you don't expect the size of the `TextBox` to change dynamically no? The dialogs are fixed size, even if they can get their height calculated dynamically.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
+            </Style>
+            <Style BasedOn="{StaticResource LabeledRadioButtonStyle}" TargetType="{x:Type RadioButton}">
+                <Setter Property="VerticalAlignment" Value="Center"/>
+            </Style>
+            <Style BasedOn="{StaticResource CommonDialogStyleBase}" TargetType="{x:Type UserControl}"/>
+        </ResourceDictionary>
     </UserControl.Resources>
-    <theming:CommonDialogWindowBaseContent DialogContent="{StaticResource Content}">
+
+    <theming:CommonDialogWindowBaseContent HasBanner="True">
```
ivannaranjo: For every dialog you need to apply the right dialog style, which will control the resulting size and look of the dialog. We do this with the following snippet:
```
<!-- Apply the style for the dialog. -->
    <UserControl.Style>
        <Binding Source="{StaticResource CommonDialogStyleSmall}" />
    </UserControl.Style>
```

There are various dialog styles to use depending on the look you want to have. This style controls the width, which is the same for all dialogs in a certain dialog class, and allows the height to float or be fixed depending on the look of the dialog.

This snippet, with the right dialog style, needs to be added to every dialog.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
         <theming:CommonDialogWindowBaseContent.Buttons>
             <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiOkButtonCaption}"
-                                      Command="{Binding DataContext.CreateCommand, Source={StaticResource Content}}"
+                                      Command="{Binding CreateCommand}"
                                       IsDefault="True"/>
             <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiCancelButtonCaption}" IsCancel="True"/>
         </theming:CommonDialogWindowBaseContent.Buttons>
+
+        <StackPanel HorizontalAlignment="Stretch" VerticalAlignment="Center">
+            <TextBlock HorizontalAlignment="Center" MaxWidth="300" Margin="5,5,5,5"
```
ivannaranjo: Don't specify width anymore, see my comment about applying the right dialog style.

ivannaranjo: Margins for textboxes are also common now and part of the `CommonTextBoxStyle`. Use those to get the right look as well.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ivannaranjo: Need to add the `CommonTextBoxStyle` to the `TextBox` to get the right look.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
```diff
@@ -6,46 +6,73 @@
...
+                        <Binding Path="Subscription.Name" UpdateSourceTrigger="PropertyChanged">
+                            <Binding.ValidationRules>
+                                <local:PubSubNameValidationRule/>
+                            </Binding.ValidationRules>
+                        </Binding>
+                    </TextBox.Text>
+                </TextBox>
+            </DockPanel>
+            <TextBlock Margin="5" Text="{x:Static ext:Resources.NewSubscriptionWindowDeliveryType}"/>
+            <RadioButton GroupName="PushPull" Margin="5" IsChecked="True"
```
ivannaranjo: We have common margins for `RadioButtons, common styles for `TextBoxes, see for example the PublisDialogSteps/GkeStep/GkeStepContent.xaml file.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ivannaranjo: The only way to ensure that the property is updated is by using `PropertyChanged` because otherwise you depend on the focus changing, which will not happen if you are using keyboard shortcuts, as you should.

GoogleCloudExtension/GoogleCloudExtension/Theming/CommonResources.xaml
```diff
@@ -232,6 +232,16 @@
         </Setter>
     </Style>
 
+    <Style x:Key="LabeledRadioButtonStyle" BasedOn="{StaticResource CommonRadioButton}" TargetType="{x:Type RadioButton}">
```
ivannaranjo: We don't apply styles globally.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,170 @@
...
+                    {
+                        request.PageToken = token;
+                        return request.ExecuteAsync();
+                    },
+                    response => response.Topics,
+                    response => response.NextPageToken);
+            }
+            catch (GoogleApiException e)
+            {
+                throw new DataSourceException("Error listing topics", e);
```
ivannaranjo: You should use the message coming form the exception here. As in:
```C#
throw new DataSourceException(e.Message, e)
```

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,170 @@
...
+                    {
+                        request.PageToken = token;
+                        return request.ExecuteAsync();
+                    },
+                    response => response.Subscriptions,
+                    response => response.NextPageToken);
+            }
+            catch (GoogleApiException e)
+            {
+                throw new DataSourceException("Error listing subscriptions", e);
```
ivannaranjo: Same in all of the `throw` statements in this file.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,170 @@
...
+                    Service.Projects.Subscriptions.Delete(GetSubscriptionFullName(subscriptionName));
+                await request.ExecuteAsync();
+            }
+            catch (GoogleApiException e)
+            {
+                throw new DataSourceException("Error deleting subscription", e);
+            }
+        }
+
+        private string GetTopicFullName(string topicName)
```
ivannaranjo: Nit, consider using the => syntax, as in:
```C#
private string GetTopicFullName(string topicName) => $"projects/{ProjectId}/topics/{topicName}";
```

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            return $"projects/{ProjectId}/topics/{topicName}";
+        }
+
+        private string GetSubscriptionFullName(string subscriptionName)
```
ivannaranjo: Nit, same.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSource.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Source class for the Pubsub tree.
+    /// </summary>
+    internal class PubsubSource : CloudExplorerSourceBase<PubsubSourceRootViewModel>
+    {
+        public PubsubSource(ICloudSourceContext context) : base(context) { }
+
+        /// <summary>
+        /// Gets the last part of the full name i.e. the leaf of the path.
+        /// </summary>
+        internal static string GetPathLeaf(string path)
```
ivannaranjo: Nit, maybe this should be part of the PubSub data source instead?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// Creates a new PubsubDataSource from the default credentials.
+        /// </summary>
+        private static PubsubDataSource CreateDataSource()
+        {
+            if (CredentialsStore.Default.CurrentProjectId != null)
+            {
+                var credential = CredentialsStore.Default.CurrentGoogleCredential;
+                if (credential.IsCreateScopedRequired)
```
ivannaranjo: Did you find that you needed the extra scope? If so we should add it to the list of scopes that we request by default as well when authorizing users. This credential comes from our own credential store, so we already have specified the scopes when logging in our users. I believe this statement is not doing anything.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,175 @@
...
+        public override TreeLeaf LoadingPlaceholder => s_loadingPlaceholder;
+
+        /// <summary>
+        /// The list of all visible subscriptions of the current project.
+        /// </summary>
+        public IList<Subscription> Subscriptions { get; private set; }
+
+        public PubsubSourceRootViewModel()
+        {
+            Subscriptions = new Subscription[0];
```
ivannaranjo: Nit, why not make `Subscriptions` of type `IEnumerable<Subscription>` and just use `Enumerable.Empty<Subscription>`. You don't actually need any of the `IList<T>` methods.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <summary>
+        /// Loads the topics, and creates child topic nodes.
+        /// </summary>
+        protected override async Task LoadDataOverride()
+        {
+            try
+            {
+                Task<IList<Subscription>> subscriptionsTask = GetSubscriptionsAsync();
+                IEnumerable<Topic> topics = await DataSource.GetTopicListAsync();
+                Subscriptions = await subscriptionsTask;
```
ivannaranjo: Reconsider storing the `Subscriptions` in this manner. What other cloud explorer sources are doing is to ensure that each view model has all of the data to initialize itself. In other words, you should pass to the `TopicViewModel` constructor the list of subscriptions to the topic. That way it can create it's own subtree, without this somewhat awkward dependency on the `_onwer` to get the subscriptions.

In case of a local refresh, the `TopicViewModel` can act completely on its own without having to affect the overall tree of view models.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionItem.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public SubscriptionItem(Subscription subscription) :
+            base(Resources.CloudExplorerPubSubSubscriptionCategory, subscription.Name)
+        {
+            _subscription = subscription;
+        }
+
+        [LocalizedCategory(nameof(Resources.CloudExplorerPubSubSubscriptionCategory))]
+        [LocalizedDescription(nameof(Resources.CloudExplorerPubSubSubscriptionNameDescription))]
+        public string Name => PubsubSource.GetPathLeaf(_subscription.Name);
```
ivannaranjo: You should also have a `[LocalizedDisplayName]` used here. We are trying to control explicitly all of the strings displayed to the user.

ivannaranjo: Same in all properties.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionViewModel.cs
```diff
@@ -0,0 +1,111 @@
...
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    /// <summary>
+    /// A leaf node that describes a Pubsub subscription.
+    /// </summary>
+    internal class SubscriptionViewModel : TreeLeaf, ICloudExplorerItemSource
+    {
+        // TODO(Jimwp): Change to Pubsub specific icon.
+        private const string IconResourcePath = "CloudExplorerSources/Gcs/Resources/bucket_icon.png";
```
ivannaranjo: You should consider doing this before merging into master, it signals that the feature is not done yet.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionViewModel.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <summary>
+        /// Returns the context in which this view model is working.
+        /// </summary>
+        public ICloudSourceContext Context => _owner.Context;
+
+        /// <summary>
+        /// The datasource for the item.
+        /// </summary>
+        public PubsubDataSource DataSource => _owner.DataSource;
+        public event EventHandler ItemChanged;
```
ivannaranjo: Missing new line between property and event.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicItem.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public TopicItem(Topic topic) :
+            base(Resources.CloudExplorerPubSubTopicCategory, PubsubSource.GetPathLeaf(topic.Name))
+        {
+            _topic = topic;
+        }
+
+        [LocalizedCategory(nameof(Resources.CloudExplorerPubSubTopicCategory))]
+        [LocalizedDescription(nameof(Resources.CloudExplorerPubSubTopicNameDescription))]
+        public string Name => PubsubSource.GetPathLeaf(_topic.Name);
```
ivannaranjo: Use the `[LocalizedDisplayName]` attribute.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,161 @@
...
+
+namespace GoogleCloudExtension.CloudExplorerSources.PubSub
+{
+    /// <summary>
+    /// A node that describes a Pubsub topic. A child of the Pubsub root and has Pubsub subscription childeren.
+    /// </summary>
+    internal class TopicViewModel : TreeHierarchy, ICloudExplorerItemSource
+    {
+        // TODO(Jimwp): Use new pubsub specific icon.
+        private const string IconResourcePath = "CloudExplorerSources/Gcs/Resources/bucket_icon.png";
```
ivannaranjo: Same, please add the right icons.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,161 @@
...
+        private static readonly Lazy<ImageSource> s_topicIcon =
+            new Lazy<ImageSource>(() => ResourceUtils.LoadImage(IconResourcePath));
+        private PubsubSourceRootViewModel _owner;
+
+        private TopicItem _topicItem;
+        public PubsubDataSource DataSource => _owner.DataSource;
```
ivannaranjo: Missing line between field and properties. Please clearly separate fields from properties.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        public object Item => _topicItem;
+        /// <summary>
```
ivannaranjo: Missing line.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        public event EventHandler ItemChanged;
+
+        public TopicViewModel(PubsubSourceRootViewModel owner, Topic topic)
+        {
+            _owner = owner;
+            _topicItem = new TopicItem(topic);
+            Caption = _topicItem.Name;
+            Icon = s_topicIcon.Value;
+            foreach (var subscription in ListSubscriptionViewModels())
```
ivannaranjo: You should consider passing into this constructor the list of subscriptions that apply to this topic, then create the view models for them here and add them as children.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionViewModel.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                if (value != Push)
+                {
+                    Subscription.PushConfig = value ? PushConfig : null;
+                    RaisePropertyChanged(nameof(Push));
+                }
+            }
+        }
+
+        public PushConfig PushConfig { get; }
```
ivannaranjo: Comments on the properties.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewTopicViewModel.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Data object that backs the new topic window. Contains necessicary data for creating a new topic.
+    /// </summary>
+    public class NewTopicViewModel : ViewModelBase
+    {
+        private readonly CommonDialogWindowBase _owner;
+        private string _topicName;
+
+        public string Project { get; }
```
ivannaranjo: Please add comments on the properties.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/PubSubNameValidationRule.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Windows.Controls;
+
+namespace GoogleCloudExtension.PubSubWindows
+{
+    /// <summary>
+    /// A validation rule that checks that a name fits the common google naming rules.
+    /// </summary>
+    public class PubSubNameValidationRule : ValidationRule
```
ivannaranjo: We're no longer using validation rules to validate input, we use explicit methods to validate the state of the view model and show error dialogs to the user with the specific errors found.

Using validation actually is more fickle, it requires the user to tab out and can show invalid states for in between states.

We are using explicitly validation at the point of commit of the data, before the action is taken we validate the view model and if everything checks out the action is taken, otherwise we will show error messages to the user.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ILMTitan: Done.

ILMTitan:
FYI, you can add files to .gitignore without removing them from the index. Doing that means local changes no longer show up in the commit list, but it still exists for an initial checkout.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/Theming/CommonResources.xaml
ILMTitan: This isn't applied globally. This is just a new keyed Style for a RadioButton that uses a Label rather than a TextBlock.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModel.cs
ILMTitan: Done. The error message was nice and specific, but it was only tagged with code 400 (Bad Request), so I can't import it into our local resources without doing a string compare.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: Done.
CommonTextBoxStyle has many good things, but Margins are not one of them.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ILMTitan: Done? It does not fit on one line. Is it ok to have an expression body over two?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ILMTitan: Mostly to avoid warnings about multiple enumeration of IEnumerable.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionItem.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionViewModel.cs
ILMTitan: Copied two of the generic "cloud" icons into a PubSub/Resources folder and referenced them.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicItem.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ILMTitan: It appears you are right.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/SubscriptionViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewTopicViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/PubSubNameValidationRule.cs
ILMTitan: Done.
I had wanted to try to keep the WPF validation, but was shocked how badly it actually worked. Apparently the view model only gets updated if validation passes (which means the last valid value exists in the model at all times).

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/PubSubWindows/NewSubscriptionWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSource.cs
ILMTitan: Done.

ILMTitan:
PTAL

ivannaranjo:
Thank you for making the changes.

PR #524, opened 2017-03-30
Fix xaml errors

ILMTitan:
Either Resharper or the built in visual studio designer does not like the explicit implementation of `IList` by `BindableList<T>`. The error "The type does not implement IList: BindableList\`1" exists on every `<theming:CommonDialogWindowBaseContent.Buttons>` element. This PR fixes all of those errors by making `BindableList<T>` implicitly implement `IList` by making it a subclass of `ObservableCollection<T>`.

Instead of overriding `Add()` and `Set()` to call `SetupDataContextBinding()`, it overrides `InsertItem()` and `SetItem()`.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Theming/CommonDialogWindowBaseContent.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Theming/CommonDialogWindowBaseContent.cs
@@ -14,7 +14,6 @@
 
 using GoogleCloudExtension.Utils;
 using System;
-using System.Collections;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Media;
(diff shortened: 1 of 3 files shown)


GoogleCloudExtension/GoogleCloudExtension/Theming/CommonDialogWindowBaseContent.cs
```diff
@@ -34,7 +33,7 @@ public class CommonDialogWindowBaseContent : ContentControl
         public static readonly DependencyProperty ButtonsProperty =
             DependencyProperty.Register(
                 nameof(Buttons),
-                typeof(IList),
+                typeof(BindableList<DialogButtonInfo>),
```
ivannaranjo: Can we make it `IList<DialogButtonInfo>` instead? I don't want to burn the `BindableList<T>` implementation in the property if we can.

ILMTitan: Oddly, changing it to `IList<DialogButtonInfo>` causes other xaml errors, but leaving it as `IList` does not. Changing it back to `IList`.

ILMTitan:
PTAL

PR #525, opened 2017-03-30
Use 1.0 tag instead.

ivannaranjo:
Trivial change to use the `1.0` tag instead of the `1.0.3` tag. By using the `1.0` tag we can ensure that we will always use the latest version in the 1.0.x branch.

Now we should think about how to pin to a specific version of the image within that version, how to list the different versions, etc... But pinning to `1.0` is the right default.


ivannaranjo:
Thank you for the review. One day we will have not even have this in the code, but smalls steps.

PR #529, opened 2017-03-31
M1 1 adding vs2017 support

ivannaranjo:
This PR prepares the extension to support VS 2017. It uses the latest build tools to produce a .vsix package that is compatible with both VS 2015 and VS 2017.

The PR also updates all of the `Google.*` packages to their latest release.

Because of an issue in the the build tasks I had to disable deploying the .vsix after every build. The deployment fails because of an issue with the paths being too long, this is a regression of the VS 2017 tooling as this used to work fine for VS 2015. To install the .vsix manually we can use a tool like [Root-Vsix](https://github.com/ivannaranjo/Root-VSIX) which I have modified to support installing on top of an already existing .vsix. More instructions on how to do this to come.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/source.extension.vsixmanifest
+++ b/GoogleCloudExtension/GoogleCloudExtension/source.extension.vsixmanifest
@@ -5,7 +5,7 @@
     The Version attribute of the Identity element *must* match the version number in Properties\AssemblyInfo.cs, to ensure 
     accurate metrics.
     -->
-    <Identity Id="GoogleAppEngine.Google.d3d3eeb8-3710-4bd9-97ba-1401bf2acd22" Version="1.1.7.0" Language="en-US" Publisher="Google Inc." />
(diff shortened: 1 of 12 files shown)


GoogleCloudExtension/GoogleCloudExtension/source.extension.vsixmanifest
```diff
@@ -24,5 +24,14 @@
   <Assets>
     <Asset Type="Microsoft.VisualStudio.VsPackage" d:Source="Project" d:ProjectName="%CurrentProject%" Path="|%CurrentProject%;PkgdefProjectOutputGroup|" />
     <Asset Type="Microsoft.VisualStudio.ProjectTemplate" d:Source="File" Path="ProjectTemplates" />
+    <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="%CurrentProject%" Path="|%CurrentProject%|" />
   </Assets>
+  <Prerequisites>
```
Deren-Liao: What's going to happen if the user does not install these components?   Does the installer remind user to install them or simply fails?

ivannaranjo: It will ask you to install the missing components.

PR #530, opened 2017-03-31
Add a select file dialog

Deren-Liao:
When there are multiple possible matching files, show a dialog and let user to pick the right one.


Deren-Liao:
Thank  you for the review.

PR #534, opened 2017-04-01
Search for local git repository

Deren-Liao:
- Add Libgit2Sharp lib.  It requires x86 native .dll to be copied over to target explicitly.
- Add GitCommit that searches for Local Git Repo .
- Add GitTempFiles helper class that manages temporary files for source file revisions.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/ActionCancelledException.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/ActionCancelledException.cs
@@ -1,0 +1,26 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace GoogleCloudExtension.SourceBrowsing
+{
+    /// <summary>
+    /// This exception indicates the user canceled the browsing source operation.
+    /// </summary>
+    internal class ActionCancelledException : Exception
+    {
+        internal ActionCancelledException() : base() { }
+    }
+}
--- a/GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
@@ -1,0 +1,168 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GoogleCloudExtension.Utils;
+using LibGit2Sharp;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+namespace GoogleCloudExtension.SourceBrowsing
+{
+    /// <summary>
+    /// Represent a local git commit on a given git SHA.
+    /// </summary>
+    internal class GitCommit
+    {
+        private readonly Commit _commit;
+        private readonly Repository _repo;
+        private readonly string _localRoot;
+        
+        internal GitCommit(Commit commit, Repository repo, string localRoot)
+        {
+            _commit = commit.ThrowIfNull(nameof(commit));
+            _repo = repo.ThrowIfNull(nameof(repo));
+            _localRoot = localRoot.ThrowIfNullOrEmpty(nameof(localRoot));
+        }
+
+        /// <summary>
+        /// Search for a Git Commit on a directory. 
+        /// The method traversal the directory tree bottom up to check 
+        /// if the directory is a valid local repo and contains the git SHA.
+        /// </summary>
+        /// <param name="path">The path to search for local git repo.</param>
+        /// <param name="sha">The git commit SHA to be searched for.</param>
+        /// <returns>
+        /// A <seealso cref="GitCommit"/> object that represents the git SHA commit at a local repository.
+        /// null if not found.
+        /// </returns>
+        public static GitCommit FindCommit(string path, string sha)
+        {
+            string dir = path.ThrowIfNullOrEmpty(nameof(path));
+            sha.ThrowIfNullOrEmpty(nameof(sha));
+            do
+            {
+                if (Repository.IsValid(dir))
+                {
+                    var repo = new Repository(dir);
+                    Commit commit = repo.Lookup<Commit>(sha);
+                    if (commit != null)
+                    {
+                        return new GitCommit(commit, repo, dir);
+                    }
+                }
+                dir = Directory.GetParent(dir)?.FullName;
+            } while (dir != null);
+            return null;
+        }
+
+        /// <summary>
+        /// Get the file of the revision.
+        /// This method firstly open the revision of the file in a temporary path.
+        /// Comparing the temporary file with the current active version of file.
+        /// If the current active file is same as the revision of the file, return the current active file.
+        /// Otherwise, return the temporary file.
+        /// </summary>
+        /// <param name="filePath">The file to be searched for.</param>
+        public string GetFile(string filePath)
+        {
+            TreeEntry treeEntry;
+            var matchingFiles = FindMatchingEntry(filePath).ToList();
+            if (matchingFiles.Count == 0)
+            {
+                throw new Exception(String.Format(
+                    Resources.SourceVersionUtilsFailedToLocateFileInRepoMessage,
+                    filePath,
+                    _commit.Sha,
+                    _localRoot));
+            }
+            if (matchingFiles.Count() > 1)
+            {
+                var index = PickFileDialog.PickFileWindow.PromptUser(matchingFiles.Select(x => x.Path));
+                if (index < 0)
+                {
+                    throw new ActionCancelledException();
+                }
+
+                treeEntry = matchingFiles[index];
+            }
+            else
+            {
+                treeEntry = matchingFiles.First();
+            }
+            var tmpPath = GitTempFiles.Current.GetOrSave(_commit.Sha, treeEntry.Path, (tmpFile) => SaveTempFile(treeEntry, tmpFile));
+            var currentPath = CurrentPath(treeEntry);
+            return _repo.RetrieveStatus(treeEntry.Path) == FileStatus.Unaltered && FileCompare(currentPath, tmpPath) ? currentPath : tmpPath;            
+        }
+
+        private void SaveTempFile(TreeEntry treeEntry, string filePath)
+        {
+            var blob = (Blob)treeEntry.Target;
+            using (var contentStream = blob.GetContentStream())
+            {
+                Debug.Assert(blob.Size == contentStream.Length);
+                contentStream.Seek(0, SeekOrigin.Begin);
+                using (var fileStream = File.Create(filePath))
+                {
+                    contentStream.CopyTo(fileStream);
+                }
+            }
+        }
+
+        private string CurrentPath(TreeEntry treeEntry) => Path.Combine(_localRoot, treeEntry.Path);
+
+        private IEnumerable<TreeEntry> FindMatchingEntry(string filePath) =>
+            SubPaths(filePath).Select(x => _commit[x]).Where(y => y != null);
+
+        private static IEnumerable<string> SubPaths(string filePath)
+        {
+            filePath.ThrowIfNullOrEmpty(nameof(filePath));
+
+            int index = filePath.IndexOf(Path.DirectorySeparatorChar);
+            for (; index >= 0; index = filePath.IndexOf(Path.DirectorySeparatorChar, index+1))
+            {
+                yield return filePath.Substring(index + 1);
+            }
+        }
+
+        /// <summary>
+        /// This method compare two files line by line.
+        /// </summary>
+        /// <returns>
+        /// True: the two files content are same.
+        /// False: thw two files differs.
+        /// </returns>
+        private bool FileCompare(string file1, string file2)
+        {
+            string line1, line2;
+            using (var fs1 = new StreamReader(file1))
+            using (var fs2 = new StreamReader(file2))
+            {
+                do
+                {
+                    line1 = fs1.ReadLine();
+                    line2 = fs2.ReadLine();
+                    if (line1 != line2)
+                    {
+                        return false;
+                    }
+                }
+                while (!fs1.EndOfStream && !fs2.EndOfStream);
+                return fs1.EndOfStream && fs2.EndOfStream;
+            }
+        }
+    }
+}
--- a/GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
@@ -1,0 +1,102 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GoogleCloudExtension.Utils;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GoogleCloudExtension.SourceBrowsing
+{
+    /// <summary>
+    /// Helper methods for managing temporary files of git revision.
+    /// 
+    /// Giving a git SHA and relative file path, returns a temporary file name.
+    ///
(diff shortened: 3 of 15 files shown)


GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <param name="relativePath">Relative path of the target file.</param>
+        /// <param name="save">The callback to save the content of the file to the temporary file path.</param>
+        /// <returns>Temporary file path.</returns>
+        public string GetOrSave(string gitSha, string relativePath, Action<string> save)
+        {
+            gitSha.ThrowIfNullOrEmpty(nameof(gitSha));
+            relativePath.ThrowIfNullOrEmpty(nameof(relativePath));
+            save.ThrowIfNull(nameof(save));
+
+            var key = $"{gitSha}/{relativePath}";
```
ILMTitan: Using a composite string as a key can lead to subtle bugs, if not now than in the future. Using a `Tuple` is much more bulletproof.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+        public static GitTempFiles Current => s_instance.Value;
+
+        private GitTempFiles()
+        {
+            if (!File.Exists(_folder.Value))
+            {
+                Directory.CreateDirectory(_folder.Value);
+            }
+            Clean();
+            ShellUtils.RegisterShutdownEventHandler(Clean);
```
ILMTitan: As is, `_folder` doesn't do enough setup to need to be lazy. If you move all of this setup to the `_folder` lazy initialization, you could remove the lazy around `s_instance`.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+            } while (dir != null);
+            return null;
+        }
+
+        /// <summary>
+        /// Get the file of the revision.
+        /// This method firstly open the revision of the file in a temporary path.
+        /// Comparing the temporary file with the current active version of file.
+        /// If the current active file is same as the revision of the file, return the current active file.
+        /// Otherwise, return the temporary file.
```
ILMTitan: This comment needs a rework. 

```
This method first opens the revision of the file in a temporary path and
compares the temporary file with the current version of the file.
```

Under what circumstances would it popup a dialog for the user? Why would someone want to get different versions based on the comparison check?

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/ActionCancelledException.cs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+using System;
+
+namespace GoogleCloudExtension.SourceBrowsing
+{
+    /// <summary>
+    /// This exception indicates the user canceled the browsing source operation.
+    /// </summary>
+    internal class ActionCancelledException : Exception
```
ivannaranjo: There's already such an exception, `OperationCancelledException`, why is this one better?

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+
+using GoogleCloudExtension.Utils;
+using LibGit2Sharp;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+namespace GoogleCloudExtension.SourceBrowsing
```
ivannaranjo: In general, there's really no need for these classes to be in the main assembly, why not create a separate project with all of the source browsing classes. That way you can also create a separate assembly with tests, etc, etc...

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                if (Repository.IsValid(dir))
+                {
+                    var repo = new Repository(dir);
+                    Commit commit = repo.Lookup<Commit>(sha);
+                    if (commit != null)
+                    {
+                        return new GitCommit(commit, repo, dir);
+                    }
+                }
```
ivannaranjo: I would agree with that. If a repo is found but it doesn't contain the commit SHA then we're done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// Otherwise, return the temporary file.
+        /// </summary>
+        /// <param name="filePath">The file to be searched for.</param>
+        public string GetFile(string filePath)
+        {
+            TreeEntry treeEntry;
+            var matchingFiles = FindMatchingEntry(filePath).ToList();
+            if (matchingFiles.Count == 0)
+            {
+                throw new Exception(String.Format(
```
ivannaranjo: We said it before, please **dot not** use exceptions to communicate error messages this way. Use the catch to detect the situation and format the error message. Do not do this.

Instead create your own custom exception type with the right information.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+                    filePath,
+                    _commit.Sha,
+                    _localRoot));
+            }
+            if (matchingFiles.Count() > 1)
+            {
+                var index = PickFileDialog.PickFileWindow.PromptUser(matchingFiles.Select(x => x.Path));
+                if (index < 0)
+                {
+                    throw new ActionCancelledException();
```
ivannaranjo: I would prefer if you return `bool` or just `null` when the user cancels. Throwing is more for errors than normal conditions. The user cancelling is **not** an error.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+                }
+
+                treeEntry = matchingFiles[index];
+            }
+            else
+            {
+                treeEntry = matchingFiles.First();
+            }
+            var tmpPath = GitTempFiles.Current.GetOrSave(_commit.Sha, treeEntry.Path, (tmpFile) => SaveTempFile(treeEntry, tmpFile));
```
ivannaranjo: Does this create a temporary file? How does it get cleaned up?

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+            var currentPath = CurrentPath(treeEntry);
+            return _repo.RetrieveStatus(treeEntry.Path) == FileStatus.Unaltered && FileCompare(currentPath, tmpPath) ? currentPath : tmpPath;            
+        }
+
+        private void SaveTempFile(TreeEntry treeEntry, string filePath)
+        {
+            var blob = (Blob)treeEntry.Target;
+            using (var contentStream = blob.GetContentStream())
+            {
+                Debug.Assert(blob.Size == contentStream.Length);
```
ivannaranjo: This seems a useless check, unless you are checking for bugs in the git library.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+                contentStream.Seek(0, SeekOrigin.Begin);
```
ivannaranjo: Is the content stream position at the end? Why do you need to seek?

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        /// <summary>
+        /// This method compare two files line by line.
+        /// </summary>
+        /// <returns>
+        /// True: the two files content are same.
+        /// False: thw two files differs.
+        /// </returns>
+        private bool FileCompare(string file1, string file2)
```
ivannaranjo: Wouldn't be better to compare hashes instead? You can also probably shortcut it by comparing file sizes first.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+
+using GoogleCloudExtension.Utils;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GoogleCloudExtension.SourceBrowsing
+{
+    /// <summary>
+    /// Helper methods for managing temporary files of git revision.
```
ivannaranjo: This comment indicates a `static` class, but this is not a static class. The comment needs to be re-worked to match the class.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
ivannaranjo: Any time you have a singleton you need to use the `Lazy<T>` pattern really, unfortunate as it is.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// The class cleanup files on VS exit event.
+    /// </summary>
+    internal class GitTempFiles
+    {
+        private const string TempSubFolder = "GoogleToolsForVS";
+
+        // Use () => new GitTempFiles() here, because the constructor is private.
+        private static Lazy<GitTempFiles> s_instance = new Lazy<GitTempFiles>(() => new GitTempFiles());
+
+        private Lazy<string> _folder = new Lazy<string>(
```
ivannaranjo: This is **not** a true temporary path however, you need to generate a random path to ensure that there are no collisions. See other parts the extension when this is done, such as the publishing code.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
ivannaranjo: Use something like `var stageDirectory = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());`, to ensure that the directory is truly unique.

ivannaranjo: Don't do this please, delete the directories as you go, as I do in the publishing. If you do this then the shutdown of VS is going to take longer for no real reason as it will have to wait for the directories to be deleted.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+            save.ThrowIfNull(nameof(save));
+
+            var key = $"{gitSha}/{relativePath}";
+            if (_tmpFilesMap.ContainsKey(key))
+            {
+                return _tmpFilesMap[key];
+            }
+            else
+            {
+                var filePath = NewTempFileName(Path.GetFileName(relativePath));
```
ivannaranjo: Bad name, something more like `GetTempFileName()` would be better.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+                save(filePath);
```
ivannaranjo: Bad name, **all** names start with uppercase.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+        public string GetOrSave(string gitSha, string relativePath, Action<string> save)
+        {
+            gitSha.ThrowIfNullOrEmpty(nameof(gitSha));
+            relativePath.ThrowIfNullOrEmpty(nameof(relativePath));
+            save.ThrowIfNull(nameof(save));
+
+            var key = $"{gitSha}/{relativePath}";
+            if (_tmpFilesMap.ContainsKey(key))
+            {
+                return _tmpFilesMap[key];
```
ivannaranjo: You are doing the lookup twice, use `_tempFilesMap.TryGetValue()` instead.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+                _tmpFilesMap[key] = filePath;
+                return filePath;
+            }
+        }
+
+        private void Clean()
+        {
+            try
+            {
+                Array.ForEach(Directory.GetFiles(_folder.Value), File.Delete);
```
ivannaranjo: Once you fix the name of the directory to be truly unique you can just delete the directory in a single call. See `CommonUtils.Cleanup` for this.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+            catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException
+                || ex is DirectoryNotFoundException || ex is PathTooLongException)
+            { }
+        }
+
+        private string NewTempFileName(string name) => 
+            Path.Combine(_folder.Value, $"{Guid.NewGuid()}_{name}");
```
ivannaranjo: Don't use guids for this, use `Path.GetRandomFileName()`.

GoogleCloudExtension/GoogleCloudExtension/Utils/ShellUtils.cs
```diff
@@ -28,6 +28,8 @@ namespace GoogleCloudExtension.Utils
     /// </summary>
     public static class ShellUtils
     {
+        private static Guid s_GuidMicrosoftCsharpEditor = new Guid("{A6C744A8-0E4A-4FC6-886A-064283054674}");
```
ivannaranjo: `s_guidMicrosoftCsharpEditor` actually.

GoogleCloudExtension/GoogleCloudExtension/Utils/ShellUtils.cs
```diff
@@ -132,6 +132,17 @@ public static void SaveAllFiles()
             dte.ExecuteCommand("File.SaveAll");
         }
 
+        /// <summary>
+        /// Register a Visual Studio Shutdown event handler.
+        /// Normally for some quick cleanup tasks.
+        /// </summary>
+        /// <param name="onExitEventHandler">The event handler.</param>
+        public static void RegisterShutdownEventHandler(Action onExitEventHandler)
```
ivannaranjo: Don't do this unless absolutely necessary, this event is not realiable and will make VS take longer to shutdown. As of now you don't need it so, please delete this.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
Deren-Liao: This is a passed in action

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
Deren-Liao: The chance to bugs is close to Zero. SHA won't contain / character.
 I just tried to use tuple then I would have to give up the StringComparer.IgnoreCase option as a benefit. Then the code becomes more redundant.
If there is no other strong reasons, I'd keep it as is.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
Deren-Liao: I have many places that need distinguish case of "User Cancelled"  or "null return please try other way".  Just like c++ often uses global error code for this purpose, in C#, exception is a way. 
I agree this is not ideal way to write the code. In my this lib, there are many places I want to signal "user canceled the action". Use this exception in my particular case simplifies the code.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
Deren-Liao: The classes have dependencies to PromptUser etc.  I'd keep them as is for now.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
Deren-Liao: Did I do this before? I think this is first time.   I am going to replace it with a error prompt.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
Deren-Liao: N/A

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
Deren-Liao: N/A

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTempFiles.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/Utils/ShellUtils.cs
Deren-Liao: N/A

GoogleCloudExtension/GoogleCloudExtension/Utils/ShellUtils.cs
Deren-Liao: Removed.

PR #535, opened 2017-04-03
Adding an abstraction layer for VS2015 and VS2017 tools and .json and .csproj projects.

ivannaranjo:
This PR adds two abstraction layers to the extension which allows us to isolate the deployment code from the differences in the underlying project systems employed by VS 2015 and VS 2017.

### Host abstraction layer.
The first layer is the host abstraction layer. This layer provides functionality that varies depending on the version of VS that is hosting the extension. Currently the only service offered by this layer is the `IToolsPathProvider` interface. This interface provides the paths to the tools used by the deployment classes to build and deploy the apps to Windows VMs, App Engine and GKE.

This layer is composed of two namespaces, `HostAbstraction.VS14` and `HostAbstraction.VS15`. Each namespace contains implementations for the host abstraction interfaces, now only `IToolsPathProvider`. There's a common class called `HostAbstractionLayer` which takes care of detecting the version of VS that is hosting the extension and returning the right implementation of the common interfaces.

### Project abstraction layer
The project abstraction layer allows the deployment modules to accept project instances without having to care for the specifics for the project, whether it is a `project.json` or a `.csproj` project, for example.

There two namespaces, `Projects.Net4` and `Projects.NetCore` which contain the project implementation for each supported runtime. Under those there are the various types of projects supported. `CsprojProject` and `JsonProject`, which represent the support for the `.csproj` and `project.json` projects.

There's a common class, `ProjectParser` which takes care of opening the project file and determining what kind of project it is. It will then create the right instance of project to manage that project.

With this layer the deployment code does not have to care about the particulars of each project.

### Review details
The review contains 30 files, but most of the changes are actually just renaming of classes and moving files around. The main meat of the review is in the common classes, `HostAbstractionLayer` and `ProjectParser`.


GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            var guids = projectGuids.Split(';');
+            if (guids.Contains(WebApplicationGuid))
+            {
+                return new Net4.CsprojProject(project);
+            }
+            return null;
+        }
+
+        private static IParsedProject ParseProjectJson(Project project)
```
ILMTitan: This function should be name `ParseJsonProject` to match `ParseMsbuildProject`.

GoogleCloudExtension/GoogleCloudExtension/Projects/NetCore/JsonProject.cs
```diff
@@ -12,30 +12,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using GoogleCloudExtension.Deployment;
 using System.IO;
 
-namespace GoogleCloudExtension.SolutionUtils
+namespace GoogleCloudExtension.Projects.NetCore
```
ILMTitan: I think the namespace should be `...Projects.DotNetCore`. The "dot" in .NET is pronounced, and should be part of the namespace. Even more for the `...Projects.DotNet4`, because `Net4` sounds like it could be just a fourth type of networking library.

GoogleCloudExtension/GoogleCloudExtension.Deployment/KnownProjectTypes.cs
```diff
@@ -25,13 +25,18 @@ public enum KnownProjectTypes
...
-        NetCoreWebApplication,
+        NetCoreWebApplication1_0,
+
+        /// <summary>
+        /// An ASP.NET Core 1.1 app
+        /// </summary>
+        NetCoreWebApplication1_1,
     }
-}
+}
```
ivannaranjo: That's a unixism that really doesn't apply on Windows. It has no consequence other than the anoying warning so I never chase them around.

GoogleCloudExtension/GoogleCloudExtension/SolutionUtils/SolutionHelper.cs
```diff
@@ -47,57 +52,50 @@ public static SolutionHelper CurrentSolution
...
                 return null;
             }
 
-            // .NET Core projects are not fully supported, only exist as a project.json file in a directory, check
-            // if the file exists.
-            var possibleProjectJson = Path.Combine(selectedProjectDirectory, ProjectJsonName);
-            if (File.Exists(possibleProjectJson))
-            {
-                return new NetCoreProject(possibleProjectJson);
-            }
```
ivannaranjo: Yes, that is now moved to the `ProjectParser` class.

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
ivannaranjo: That's a really good point, I also renamed the other method to `ParseCsprojProject` to match the naming of the project classes.

GoogleCloudExtension/GoogleCloudExtension/Projects/NetCore/JsonProject.cs
ivannaranjo: Agreed, and done.

ivannaranjo:
PTAL

GoogleCloudExtension/GoogleCloudExtension.Deployment/IToolsPathProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace GoogleCloudExtension.Deployment
+{
+    /// <summary>
+    /// Implementation of this interface provide the path to the tools needed by the publish
+    /// operations.
+    /// </summary>
+    public interface IToolsPathProvider
+    {
+        /// <summary>
+        /// Returns the path to use for external tools to be used during deployment.
```
Deren-Liao: By reading the comment, this sounds like a general path for a bunch of tools.  Then I wonder , the return value should be a list of strings.   I assume you only expect one path for return.  Can you modify the name and comment to make it more clear?

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/HostAbstractionLayer.cs
```diff
@@ -0,0 +1,50 @@
...
+            switch (GoogleCloudExtensionPackage.VsVersion)
+            {
+                case VisualStudio2015Version:
+                    return new VS14.ToolsPathProvider();
+
+                case VisualStudio2017Version:
+                    return new VS15.ToolsPathProvider(GoogleCloudExtensionPackage.VsEdition);
+
+                default:
+                    throw new NotSupportedException($"Version {GoogleCloudExtensionPackage.VsVersion} is not supported.");
```
Deren-Liao: Should this string be localized?

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/HostAbstractionLayer.cs
```diff
@@ -0,0 +1,50 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GoogleCloudExtension.Deployment;
+using System;
+
+namespace GoogleCloudExtension.HostAbstraction
```
Deren-Liao: nit,  I'd recommend a more descriptive name like VSVersion VSVersionUtil .

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/HostAbstractionLayer.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// This class provides services that are Visual Studio version specific.
+    /// </summary>
+    internal static class HostAbstractionLayer
```
Deren-Liao: nit. This is a Factory or Builder,  use a name like ToolsPathProviderFactory or something sounds better to me.

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/VS15/ToolsPathProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var programFilesPath = Environment.ExpandEnvironmentVariables("%ProgramW6432%");
+            var result = Path.Combine(programFilesPath, @"dotnet\dotnet.exe");
+            GcpOutputWindow.OutputDebugLine($"Dotnet path: {result}");
+            return result;
+        }
+
+        public string GetExternalToolsPath()
+        {
+            return "";
```
Deren-Liao: When it is empty, does it use the current directory?  Or it's not used for VS2017?  if the later case, add NotImplementedException()?

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/VS14/ToolsPathProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+using GoogleCloudExtension.Utils;
+using System;
+using System.IO;
+
+namespace GoogleCloudExtension.HostAbstraction.VS14
+{
+    /// <summary>
+    /// The implementation of <seealso cref="IToolsPathProvider"/> for Visual Studio 2015 (v14.0).
+    /// </summary>
+    internal class ToolsPathProvider : IToolsPathProvider
```
Deren-Liao: nit, cache the paths after it is called once so that later call won't need to do all the string operation.

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
```diff
@@ -0,0 +1,134 @@
...
+        private const string TargetFrameworkElementName = "TargetFramework";
+
+        // The PropertyTypeGuids element contains the GUID that identifies the type of project, console, web, etc...
+        private const string PropertyTypeGuidsElementName = "ProjectTypeGuids";
+
+        // The Sdk attribute points to the SDK used to build this app, console or web.
+        private const string SdkAttributeName = "Sdk";
+
+        /// <summary>
+        /// Parses the given <seealso cref="Project"/> instance and returned a friendlier and more usable type to use for
```
Deren-Liao: nit,   "returns " ?

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        private static IParsedProject ParseCsprojProject(Project project)
+        {
+            GcpOutputWindow.OutputDebugLine($"Parsing .csproj {project.FullName}");
+
+            var dom = XDocument.Load(project.FullName);
+            var sdk = dom.Root.Attribute(SdkAttributeName);
+            if (sdk != null)
+            {
+                GcpOutputWindow.OutputDebugLine($"Found a new style .csproj {sdk.Value}");
```
Deren-Liao: "new" style?   If we use it a year later, is it still "new" style?  I feel it might be confusing.

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
```diff
@@ -0,0 +1,134 @@
...
+                default:
+                    return null;
+            }
+        }
+
+        private static IParsedProject ParseCsprojProject(Project project)
+        {
+            GcpOutputWindow.OutputDebugLine($"Parsing .csproj {project.FullName}");
+
+            var dom = XDocument.Load(project.FullName);
```
Deren-Liao: Catch XML parser error.

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
```diff
@@ -0,0 +1,134 @@
...
+                    Debug.WriteLine($"Processing a .csproj: {project.FullName}");
+                    return ParseCsprojProject(project);
+
+                default:
+                    return null;
+            }
+        }
+
+        private static IParsedProject ParseCsprojProject(Project project)
+        {
```
Deren-Liao: Use a try catch for all the XML parse error?

GoogleCloudExtension/GoogleCloudExtension.Deployment/IToolsPathProvider.cs
ivannaranjo: The name of the method is exactly what it says, it is returning the path to the external tools to use, it is not a list of paths. I will update the comment to make that clear.

The external tools are all installed in VS 2015 under a single path, which is represented by the name. These tools do **not** exist in VS 2017 and forward, which is why it just returns "" on the VS15 implementation.

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/HostAbstractionLayer.cs
ivannaranjo: No, it is not shown to the user. We never show strings coming from exceptions to the user.

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/HostAbstractionLayer.cs
ivannaranjo: That's your java programmer life talking :). C# programmers should hate those names, provider is as far as you get. This class is meant to provide multiple services based on the version of VS hosting the extension, I'll think of a better name then.

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/VS15/ToolsPathProvider.cs
ivannaranjo: The value of "" means no path, nothing is added to the PATH environment variable then. The deployment tools will see this value and use it.

What this method mean is that there's no tools directory for VS 2017.

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/VS14/ToolsPathProvider.cs
ivannaranjo: This is not done in a hot loop, the caching will add unnecessary complexity.

ivannaranjo:
Will look at the class name changes, I will not however use a Java-ism in the code :).

GoogleCloudExtension/GoogleCloudExtension/HostAbstraction/HostAbstractionLayer.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension/Projects/ProjectParser.cs
ivannaranjo: Done, the documentation is not very precise about what exceptions are going to be thrown, so I am being broad.

ivannaranjo:
PTAL

GoogleCloudExtension/GoogleCloudExtension/VsVersion/HostAbstractionLayer.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
```
Deren-Liao: Nit, change the file name to match the class name.

ivannaranjo:
Thank you, once the appveyor build finishes I'll merge the PR.

PR #537, opened 2017-04-04
M2 gcs browser merging master

ivannaranjo:
This PR brings the `m2_gcs_browser` branch up to date with the `master` branch in preparation for more work done in this branch. This branch had become a bit stale, which shows by the number of files changed.

There are no new changes included in this PR, all changes were reviewed and come directly from the `master` branch. Of course there were merge conflicts so a spot check here or there would be quite helpful.


ivannaranjo:
@ILMTitan you are absolutely right, there's no reason why the .vcst file changes in this merge. I'll just accept the changes from `master` and move on. This branch is not adding any more VS level commands.

ivannaranjo:
Actually that .vsct file is the same as the one in the `master` branch. The difference is because of the staleness of the branch.

PR #554, opened 2017-04-08
use git revision id to locate source file.

Deren-Liao:
Made changes according to prior review feedback.
I am creating a new PR for I added a second private branch into this PR too. 
Major changes 
 * use git command to replace libgit2sharp for license problem.
* Delete temporary files immediately after file is opened.


GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommandWrapper.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <param name="gitLocalRoot">The file path to be checked.</param>
+        /// <returns>
+        /// <seealso cref="GitCommandWrapper"/> object or null if the path is not a valid git repository path.
+        /// </returns>
+        public static GitCommandWrapper GetGitCommandWrapperForPath(string gitLocalRoot)
+        {
+            return IsGitRepository(gitLocalRoot) ? new GitCommandWrapper(gitLocalRoot) : null;
+        }
+
+        public static bool IsGitRepository(string gitLocalRoot) => RunGitCommand("log -1", gitLocalRoot)?.Count > 0;
```
ivannaranjo: Add comments for the public method.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommandWrapper.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// </summary>
+        public string Root { get; }
+
+        /// <summary>
+        /// Gets the git.exe full path if it is installed properly.
+        /// Return null if git.exe is not found.
+        /// </summary>
+        public static string GitPath => s_gitPathLazy.Value;
+
+        private GitCommandWrapper(string gitLocalRoot)
```
ivannaranjo: Since the constructor is private you should probably move after the public methods.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommandWrapper.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <seealso cref="GitCommandWrapper"/> object or null if the path is not a valid git repository path.
+        /// </returns>
+        public static GitCommandWrapper GetGitCommandWrapperForPath(string gitLocalRoot)
+        {
+            return IsGitRepository(gitLocalRoot) ? new GitCommandWrapper(gitLocalRoot) : null;
+        }
+
+        public static bool IsGitRepository(string gitLocalRoot) => RunGitCommand("log -1", gitLocalRoot)?.Count > 0;
+
+        public bool ContainsCommit(string sha) => ExecCommand($"cat-file -t {sha}").FirstOrDefault() == "commit";
```
ivannaranjo: Comments for public methods in the rest of the file as well.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommandWrapper.cs
```diff
@@ -0,0 +1,110 @@
...
+        public List<string> GetRevisionFile(string sha, string relativePath)
+            => ExecCommand($"show {sha}:{relativePath.Replace('\\', '/')}");
+
+        private List<string> ExecCommand(string command) => RunGitCommand(command, Root);
+
+        private static string GetGitPath()
+        {
+            // Firstly check default installation location.
+            string programPath = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
+            if (programPath.EndsWith("(x86)"))
```
ivannaranjo: No, you want the bitness independent program files, the value for that is also stored in the "ProgramFiles" environment variable. Don't assume any value for Program Files, the user can change it.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommandWrapper.cs
```diff
@@ -0,0 +1,110 @@
...
+            return Environment.GetEnvironmentVariable("PATH")
+                .Split(';')
+                .Select(x => Path.Combine(x, GitExecutable))
+                .FirstOrDefault(x => File.Exists(x));
+        }
+
+        /// <summary>
+        /// Run a git command and return the output or error output.
+        /// </summary>
+        private static List<string> RunGitCommand(
```
ivannaranjo: You need to make this method asynchronous, otherwise the UI waiting for the git command to run is going to hang. If you really want to have timeout then use a CancellationToken and add that functionality to the ProcessUtils class.

As it is right now this is going to make the UI unresponsive, which is not good.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        private readonly GitCommandWrapper _gitCommand;
+        private readonly string _sha;
+        private readonly Lazy<HashSet<string>> _fileTree;
+        
+        internal GitCommit(GitCommandWrapper gitCommand, string sha)
+        {
+            _gitCommand = gitCommand.ThrowIfNull(nameof(gitCommand));
+            _sha = sha.ThrowIfNullOrEmpty(nameof(sha));
+            _fileTree = new Lazy<HashSet<string>>(
```
ivannaranjo: You need to think about how you can do this using asynchronous code. This can potentially be a very expensive operation that is going to hang the UI while it is done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// The path is from the machine that build the assembly.
+        /// So the file path has different root path.
+        /// </param>
+        public Window OpenFileRevision(string filePath)
+        {
+            string relativePath;
+            var matchingFiles = FindMatchingEntry(filePath).ToList();
+            if (matchingFiles.Count == 0)
+            {
+                UserPromptUtils.ErrorPrompt(
```
ivannaranjo: You should separate the Git layer from the UI layer. The Git layer should not directly prompt the user, ideally the Git later should go into its own project actually.

Think about how you can separate the concerns.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommit.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GoogleCloudExtension.SourceBrowsing
+{
+    /// <summary>
+    /// Represent a local git commit on a given git SHA.
+    /// </summary>
+    internal class GitCommit
```
ivannaranjo: This class seems to be merging UI with Git information, I think you need to separate this class into two different layers, one for the pure Git management and another one which uses the Git layer and can prompt the user when needed.

The Git layer seems like something that can be re-used in other parts of the extension in the future.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitTemporaryFiles.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// Giving a git SHA and relative file path, save the file to a temporary path.
+    /// Open the file in a Document window, return the window object.
+    ///
+    /// Windows temporary files won't be cleaned up automatically.
+    /// The class deletes the temporary file immediately after the document is opened.
+    /// 
+    /// The the document window for the same file is still open,
+    /// return the document window so as not to open a new window.
+    /// </summary>
+    internal class GitTemporaryFiles
```
ivannaranjo: This class is really not managing the temp files but rather doing that and managing the UI and editors for those files. You should again separate the concerns of this class into two. The temporary file manager could be re-used in other parts of the extension.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -14,18 +14,93 @@
...
+                    var locatedFilePath = project.FindSourceFile(logItem.SourceFilePath)?.FullName;
+                    window = ShellUtils.Open(locatedFilePath);
+                    if (window == null)
+                    {
+                        FailedToOpenFilePrompt(logItem.SourceFilePath);
+                        return;
+                    }
+                }
+            }
+            catch (OperationCanceledException)
```
ivannaranjo: Where can you get this from? This is not an async method at all.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -34,84 +109,116 @@ internal static class SourceVersionUtils
...
                 {
-                    return null;    // Still does not find the project, quit.
+                    throw new OperationCanceledException();
                 }
             }
 
             if (project.Version != logItem.AssemblyVersion && !ContinueWhenVersionMismatch(project, logItem.AssemblyVersion))
             {
-                return null;
+                throw new OperationCanceledException();
```
ivannaranjo: No, you should never throw this exception yourself. You should rely on the CancellationToken. This `throw` is a symptom that you have not implemented asynchronous code at all.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -34,84 +109,116 @@ internal static class SourceVersionUtils
...
+        /// that writes the log entry.
+        /// </summary>
+        /// <param name="sha">The git commit SHA.</param>
+        /// <param name="filePath">The full file path that generates the log item or error event.</param>
+        /// <returns>
+        /// The file path of the source file revision.
+        /// null: Operation is cancelled or failed to open the file revision.
+        /// </returns>
+        /// <exception cref="ActionCancelledException">Does not find the revision of file.</exception>
+        private static EnvDTE.Window SearchGitRepoAndGetFileContent(string sha, string filePath)
```
ivannaranjo: You are not getting the content, you are getting an editor, rename the method.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/GitCommandWrapper.cs
Deren-Liao: On x64 which I assume is most commonly used,  the git is installed to wow64 ProgramFiles folder by default.  
Basically, I am going to just check the PATH then if you don't think this is safe.

Deren-Liao:
Modified please check.

Deren-Liao:
Ian, Jim, 
Could either of you take a look since Ivan is on vacation?

I've made changes according to Ivan's request.  
+ Created a GitUtils project for git commands.
+ Use async execution for git commands.  If a command takes relatively longer , user will see a progress dialog.
+ Others (comment etc)

iantalarico:
@Deren-Liao is this blocking any other work?  It's a fairly large PR which Ivan has all the context for.

Deren-Liao:
Thank you for looking into this.
Not completely blocking, but waiting for another week is too long.  Also I
have bug fixes etc that I want to submit PR too.  I hope to submit the bug
fixes PR on top of this rather than doing merges later which is quite
painful. Actually I've been holding a few bug fixes.
The overall logic, design etc Ivan already reviewed, I think it's Okay if
you focus on individual method implementations etc.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -134,5 +250,72 @@ private static bool ContinueWhenVersionMismatch(ProjectHelper project, string as
...
+            var task = ProgressDialogWindow.PromptUser<GitCommit>(
+                GitCommit.FindCommitAsync(filePath, sha),
+                s_gitOperationOption);
+            task.Wait();
+            var gitCommit = task.Result;
+            if (gitCommit != null)
+            {
+                s_localCache[sha] = gitCommit;
+            }
+            return gitCommit;
```
ILMTitan: You are already making the gitCommit available with `s_localCache`. It feels like this should return true or false.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -75,38 +155,74 @@ public static ProjectHelper FindOrOpenProject(this LogItem logItem)
...
+            {
+                OpenProjectFromLocalRepositoryPrompt();
+            }
+            if (!IsCurrentSolutionOpen())
+            {
+                return null;    // Still no solution is open, give up.
+            }
+
+            var solution = SolutionHelper.CurrentSolution;
+            commit = solution.Projects.Select(x => SearchCommitAtPath(x.ProjectRoot, sha)).Where(y => y != null).FirstOrDefault();
```
ILMTitan: Can you make this project searches run in parallel?

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -134,5 +250,72 @@ private static bool ContinueWhenVersionMismatch(ProjectHelper project, string as
...
         }
+
+        private static bool IsCurrentSolutionOpen() => SolutionHelper.CurrentSolution?.Projects?.Count > 0;
+
+        private static GitCommit SearchCommitAtPath(string filePath, string sha)
+        {
+            var task = ProgressDialogWindow.PromptUser<GitCommit>(
+                GitCommit.FindCommitAsync(filePath, sha),
+                s_gitOperationOption);
+            task.Wait();
```
ILMTitan: This is not the correct place to end async. It should be async all the way up to `NavigateToSourceLineCommand`.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -134,5 +250,72 @@ private static bool ContinueWhenVersionMismatch(ProjectHelper project, string as
...
+            {
+                s_localCache[sha] = gitCommit;
+            }
+            return gitCommit;
+        }
+
+        private static EnvDTE.Window OpenGitFile(GitCommit commit, string filePath)
+        {
+            string relativeFile;
+            var task = ProgressDialogWindow.PromptUser<IEnumerable<string>>(
```
ILMTitan: There are a lot of user prompts being opened. Can they be condensed into one?

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -134,5 +250,72 @@ private static bool ContinueWhenVersionMismatch(ProjectHelper project, string as
...
+                commit.FindMatchingEntryAsync(filePath),
+                s_gitOperationOption);
+            task.Wait();
```
ILMTitan: Same as above, `await` this, don't `Wait()` it.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -134,5 +250,72 @@ private static bool ContinueWhenVersionMismatch(ProjectHelper project, string as
...
+                relativeFile,
+                (tmpFile) => SaveFile(commit, tmpFile, relativeFile));
+        }
+
+        private static void SaveFile(GitCommit commit, string tempFile, string relativePath)
+        {
+            var task = ProgressDialogWindow.PromptUser(
+                commit.SaveTempFileAsync(tempFile, relativePath),
+                s_gitOperationOption);
+            task.Wait();
```
ILMTitan: And here.

GoogleCloudExtension/GoogleCloudExtension/Utils/ShellUtils.cs
```diff
@@ -99,18 +99,14 @@ public static void OpenProject()
...
 
         /// <summary>
-        /// Opens a project item in Visual Studio.
+        /// Opens a source file in Visual Studio.
         /// </summary>
-        /// <param name="projectItem"><seealso cref="ProjectItem"/></param>
+        /// <param name="sourceFile">Source file path</param>
         /// <returns>The Window that displays the project item.</returns>
-        public static Window Open(ProjectItem projectItem)
+        public static Window Open(string sourceFile)
```
ILMTitan: Why not add a delegating overload?

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -134,5 +250,72 @@ private static bool ContinueWhenVersionMismatch(ProjectHelper project, string as
...
+        }
+
+        private static EnvDTE.Window OpenGitFile(GitCommit commit, string filePath)
+        {
+            string relativeFile;
+            var task = ProgressDialogWindow.PromptUser<IEnumerable<string>>(
+                commit.FindMatchingEntryAsync(filePath),
+                s_gitOperationOption);
+            task.Wait();
+            var matchingFiles = task.Result;
```
ILMTitan: `matchingFiles` is an enumerable from a generator function. Enumerating it multiple times should fail.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            Root = gitLocalRoot;
+        }
+
+        private static string GetGitPath()
+        {
+            return Environment.GetEnvironmentVariable("PATH")
+                .Split(';')
+                .Select(x => Path.Combine(x, GitExecutable))
+                .FirstOrDefault(x => File.Exists(x));
```
ILMTitan: If "git" is on the path, then why do you need to find it? Won't `Process.Start` find it itself?
Also, what happened to using a git C# library?

Deren-Liao:
Thank you for the comment. 
I'll address the comment one by one.  To the overall question
* We are not allowed to use gitlib2sharp for it contains GLP license. 
* I'll see if I can change the code to all the way async up. If it is kind too much code change, unless there is clear benefit of doing so, I'd like to keep it as is.  Do you agree?
* Please note, for those that I return null , they are valid cases.  I search multiple paths for local git root.   if it is not there, I return null and try next one. In the end, if none of them are found, it is still valid user case. I'll simply display a dialog to tell user the information.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitCommit.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <returns>
+        /// A <seealso cref="GitCommit"/> object that represents the git SHA commit at a local repository.
+        /// null if the search is not successful.
+        /// </returns>
+        public static async Task<GitCommit> FindCommitAsync(string path, string sha)
+        {
+            string dir = path.ThrowIfNullOrEmpty(nameof(path));
+            sha.ThrowIfNullOrEmpty(nameof(sha));
+            var gitCommand = await GitRepository.GetGitCommandWrapperForPathAsync(dir);
+            if (gitCommand == null)
```
Deren-Liao: It is valid case if this path is not valid local git root or if it does not contain the git sha.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitCommit.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                return null;
+            }
+            return (await gitCommand.ContainsCommitAsync(sha)) == true? new GitCommit(gitCommand, sha) : null;
```
Deren-Liao: Fixed.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitCommit.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// So the file path has different root path.
+        /// </param>
+        /// <returns>File path relative to git root. </returns>
+        public async Task<IEnumerable<string>> FindMatchingEntryAsync(string filePath)
+        {
+            var fileTree = await GetFileTreeAsync();
+            return SubPaths(filePath)
+                .Select(x => {
+                    string y;
+                    return fileTree.TryGetValue(x, out y) ? y : null;
```
Deren-Liao: Some prefers to use that style to reduce a line. I am fine either way.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitCommit.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /// <summary>
+        /// Save the relative file content to a temporary file.
+        /// </summary>
+        /// <param name="tmpFile">Temporary file path.</param>
+        /// <param name="relativePath">The file path relative to the git root.</param>
+        public async Task SaveTempFileAsync(string tmpFile, string relativePath)
+        {
+            var content = await _repo.GetRevisionFileAsync(Sha, relativePath);
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitCommit.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            if (_fileTree == null)
+            {
+                _fileTree = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                foreach (var file in await _repo.ListTreeAsync(Sha))
+                {
+                    _fileTree.Add(file.Replace('/', '\\'), file);
+                }
+            }
+            return _fileTree;
```
Deren-Liao: Yes, I did not like the fancy stuff and decided to write this dumb code.  Seems to me everybody is used to it, I am going to make the change.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// Returns a <seealso cref="GitRepository"/> object 
+        /// if the <paramref name="dir"/> is a valid locat Git repository path.
+        /// </summary>
+        /// <param name="dir">The file path to be checked.</param>
+        /// <returns>
+        /// <seealso cref="GitRepository"/> object or null if the path is not a valid git repository path.
+        /// </returns>
+        public static async Task<GitRepository> GetGitCommandWrapperForPathAsync(string dir)
+        {
+            if (await IsGitRepositoryAsync(dir))
```
Deren-Liao: The critical part here is the IsGitRepositoryAsync.  
Only if this is a valid git path, return the GitRepository object.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                var root =  (await RunGitCommandAsync("rev-parse --show-toplevel", dir)).FirstOrDefault()?.Replace('/', '\\');
+                return new GitRepository(root);
+            }
+            return null;
```
Deren-Liao: As explained, it is a valid scenario that the path is not what we want. I'll search from other paths.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
Deren-Liao: I simply followed the pattern that Ivan already be using.  As I don't see anything wrong doing so, I'd keep it is.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
```diff
@@ -0,0 +1,122 @@
...
+            if (!File.Exists(gitLocalRoot) && !Directory.Exists(gitLocalRoot))
+            {
+                return null;
+            }
+            List<string> output = new List<string>();
+            return await ProcessUtils.RunCommandAsync(
+                file: GitPath,
+                args: command,
+                handler: (o, e) => output.Add(e.Line),
+                workingDir: gitLocalRoot) ? output : null;
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/OpenGitFile.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        // Use () => new GitTemporaryFiles() here, because the constructor is private.
+        private static Lazy<OpenGitFile> s_instance = new Lazy<OpenGitFile>(() => new OpenGitFile());
+
+        private readonly string _tmpFolder;
+
+        /// <summary>
+        /// Key is git_sha/relative_path, value is the opened document window.
+        /// </summary>
+        private Dictionary<string, Window> _fileRevisionWindowMap = new Dictionary<string, Window>(StringComparer.OrdinalIgnoreCase);
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/OpenGitFile.cs
```diff
@@ -0,0 +1,114 @@
...
+        private Dictionary<Window, string> _documentWindows = new Dictionary<Window, string>();
+
+        /// <summary>
+        /// Returns the singleton of the <seealso cref="OpenGitFile"/> class.
+        /// </summary>
+        public static OpenGitFile Current => s_instance.Value;
+
+        /// <summary>
+        /// Returns a cached file path that is saved in prior calls.
+        /// Or create the temporary file and save the conteint.
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/OpenGitFile.cs
```diff
@@ -0,0 +1,114 @@
...
+            Directory.CreateDirectory(_tmpFolder);
+            ShellUtils.RegisterWindowCloseEventHandler(OnWindowClose);
+        }
+
+        private Window OpenDocument(string filePath, string key)
+        {
+            var window = ShellUtils.Open(filePath);
+            window.Document.ReadOnly = true;
+            File.Delete(filePath);
+            if (window != null)
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -12,20 +12,104 @@
...
 
 namespace GoogleCloudExtension.SourceBrowsing
 {
     /// <summary>
-    /// Helper methods to find, open the project of a source version.
+    /// Helper methods to find, open the source file of the matching revision.
     /// </summary>
     internal static class SourceVersionUtils
     {
+        private const string SourceContextIDLabel = "git_revision_id";
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -12,20 +12,104 @@
...
+
+        private static readonly ProgressDialogWindow.Options s_gitOperationOption =
+            new ProgressDialogWindow.Options
+            {
+                Message = GoogleCloudExtension.Resources.SourceVersionProgressDialogMessage,
+                Title = GoogleCloudExtension.Resources.uiDefaultPromptTitle,
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -12,20 +12,104 @@
...
+                IsCancellable = false
+            };
+
+        /// <summary>
+        /// A map of git sha to <seealso cref="GitCommit"/> object.
+        /// </summary>
+        private static readonly Dictionary<string, GitCommit> s_localCache = new Dictionary<string, GitCommit>(StringComparer.OrdinalIgnoreCase);
```
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
```diff
@@ -12,20 +12,104 @@
...
+                    // fallback to using assembly version information.
+                    var project = FindOrOpenProject(logItem);
+                    if (project == null)
+                    {
+                        Debug.WriteLine($"Failed to find project of {logItem.AssemblyName}");
+                        return;
+                    }
+
+                    var locatedFilePath = project.FindSourceFile(logItem.SourceFilePath)?.FullName;
+                    window = ShellUtils.Open(locatedFilePath);
```
Deren-Liao: Fixed.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
Deren-Liao: N/A, refactored code.

Deren-Liao: I'd rather not unless there is performance problem.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
Deren-Liao: Changed.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
Deren-Liao: I need to display dialogs in between. 
If I use one progressDialogWindow.PromptUser,  then user will see a dialog above dialog. The UI experience is not good.  Designing a dedicated Dialog is way more overhead, I don't think that's necessary.

In my test, these dialog does not even be noticed because the commands are fast.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
Deren-Liao: I tested, it works fine.

GoogleCloudExtension/GoogleCloudExtension/SourceBrowsing/SourceVersionUtils.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/Utils/ShellUtils.cs
Deren-Liao: I don't need overload, just replace it.

Deren-Liao:
Hi Jim,  
Thank you for doing the code review. I updated code on your comment. Please take a look again.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/LogItem.cs
```diff
@@ -215,7 +215,8 @@ public LogItem(LogEntry logEntry, TimeZoneInfo timeZoneInfo)
                 }
 
                 SourceLinkVisible = true;
-                OnNavigateToSourceCommand = new ProtectedCommand(() => SourceVersionUtils.NavigateToSourceLineCommand(this));
+                OnNavigateToSourceCommand = new ProtectedCommand(
+                    () => SourceVersionUtils.NavigateToSourceLineCommandAsync(this).Wait());
```
ILMTitan: You don't need the lambda expression or the `.Wait()`. Just call it like you did previously.

Deren-Liao: I need to wait, otherwise user will be able to interact with the Logs Viewer which will go to unknown complex interaction .

Deren-Liao:
When it is not waited, user is able to click on other part of the UI.  I
intentionally want it be blocked . Then user will see a progress dialog if
it takes time.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitCommit.cs
ILMTitan: I would argue that it is invalid to call this on a path that is not a repository. Since it is called `Find` and not `Get`, it is valid to take a commit that does not exist.

Deren-Liao: I understand your point. I would say this is more about how to come up a better method name so as  the code here makes more sense.
The code itself logically is needed to be this way. If I don't do the check here, I'll have to write another wrapper function to check again instead of throwing an exception.  Then I'll come up another name for the wrapper.  
And I don't have a better name than this one for the moment. I'd keep it as is.

Deren-Liao:
Thanks Ivan and Jim for doing the review!

Issue #566, opened 2017-04-15
Logs Viewer Detail View : JValue element in JArray causes exception.


Deren-Liao:

Deren-Liao:
This is the fix. If element is JValue, the ParseArray would throw exception.
ObjectNodeTree.cs#L178
And then
ObjectNodeTree.cs#L165
ObjectNodeTree.cs#L143

PR #567, opened 2017-04-15
fix : possible JValue element causes exception.

Deren-Liao:
https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/issues/566

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/JArrayNode.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/JArrayNode.cs
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace GoogleCloudExtension.StackdriverLogsViewer
 {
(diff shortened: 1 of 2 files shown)


ILMTitan:
I don't understand what this is trying to fix. Is there a case that wasn't JArray, JObject, JProperty or JValue? Was there a null pointer at `jItem.Value.GetType().Name`?

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/JObjectNode.cs
```diff
@@ -40,16 +41,13 @@ protected override void ParseObjectTree(object obj)
             JObject jsonObj = obj as JObject;
             foreach (var jItem in jsonObj)
             {
-                switch (jItem.Value.GetType().Name)
+                if (jItem.Value is JArray || jItem.Value is JObject)
```
Deren-Liao: This is code refactoring.   Not a bug.

GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/JArrayNode.cs
```diff
@@ -38,7 +40,19 @@ public JArrayNode(string name, JArray jArrayObj, ObjectNodeTree parent)
         protected override void ParseObjectTree(object obj)
         {
             JArray jsonArray = obj as JArray;
-            ParseArray(jsonArray.Children());
+            List<object> objectArray = new List<object>();
```
Deren-Liao: This is the fix.  If element is JValue,  the ParseArray would throw exception.  
https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/blob/master/GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs#L178
And then
https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/blob/master/GoogleCloudExtension/GoogleCloudExtension/StackdriverLogsViewer/TreeViewConverters/ObjectNodeTree.cs#L143

Deren-Liao:
Added comment, let me know if it's still not clear.

Deren-Liao:
Thank you.

Issue #584, opened 2017-04-25
Generated source context file was removed after build on publishing 


Deren-Liao:
Currently, the source context files were generated before build. The next step is msbuild and it will remove the file.  As the result, the source context file is not published.
Need to move the source context file generation step after msbuild.

PR #585, opened 2017-04-25
Switch source context file generation order. #584

Deren-Liao:
Fix https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/issues/584


Deren-Liao:
Thanks

PR #600, opened 2017-05-05
Add attach debugger window

Deren-Liao:
* Add the window as container of attaching steps. Steps will be added later.
* Add attach debugger menu item to GCE instance.
* Modified the ProtectedCommand to accept Func<Task>.  
  This is to make sure it won't escape unhanded exceptions from async tasks.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
@@ -22,17 +22,10 @@ namespace GoogleCloudExtension.AttachRemoteDebugger
     /// <summary>
     /// Base class for attaching remote debugger step view models.
     /// </summary>
-    public class AttachDebuggerStepBase:  ViewModelBase, IAttachDebuggerStep
+    public abstract class AttachDebuggerStepBase : ViewModelBase, IAttachDebuggerStep
     {
-        private static readonly ContentControl _contentControl = new ContentControl();
-
         private bool _isOKButtonEnabled;
         private bool _isCancelButtonEnabled;
-
-        /// <summary>
-        /// Define a global exit step object that implements <seealso cref="IAttachDebuggerStep"/> interface.
-        /// </summary>
-        public static AttachDebuggerStepBase ExitStep { get; } = new AttachDebuggerStepBase();
 
         #region implement interface IAttachDebuggerStep
         public virtual bool IsCancelButtonEnabled
@@ -41,7 +34,7 @@ namespace GoogleCloudExtension.AttachRemoteDebugger
             protected set { SetValueAndRaise(ref _isCancelButtonEnabled, value); }
         }
 
-        public virtual ContentControl Content => _contentControl;
+        public virtual ContentControl Content => null;
 
         public virtual bool IsOKButtonEnabled
         {
@@ -51,18 +44,12 @@ namespace GoogleCloudExtension.AttachRemoteDebugger
 
         public virtual async Task<IAttachDebuggerStep> OnCancelCommand()
         {
-            return ExitStep;
+            return null;
         }
 
-        public virtual async Task<IAttachDebuggerStep> OnStart()
-        {
-            throw new NotImplementedException();
-        }
+        public abstract Task<IAttachDebuggerStep> OnStart();
 
-        public virtual async Task<IAttachDebuggerStep> OnOKCommand()
-        {
-            throw new NotImplementedException();
-        }
+        public abstract Task<IAttachDebuggerStep> OnOKCommand();
         #endregion
     }
 }
--- a/GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
@@ -1,0 +1,68 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GoogleCloudExtension.Utils;
+using System;
+using System.Threading.Tasks;
+using System.Windows.Controls;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
+{
+    /// <summary>
+    /// Base class for attaching remote debugger step view models.
+    /// </summary>
+    public class AttachDebuggerStepBase:  ViewModelBase, IAttachDebuggerStep
+    {
+        private static readonly ContentControl _contentControl = new ContentControl();
+
+        private bool _showOKButton = true;
+        private bool _showCancelButton = true;
+
+        /// <summary>
+        /// Define a global exit step object that implements <seealso cref="IAttachDebuggerStep"/> interface.
+        /// </summary>
+        public static AttachDebuggerStepBase ExitStep { get; } = new AttachDebuggerStepBase();
+
+        #region implement interface IAttachDebuggerStep
+        public virtual bool IsCancelButtonEnabled
+        {
+            get { return _showCancelButton; }
+            protected set { SetValueAndRaise(ref _showCancelButton, value); }
+        }
+
+        public virtual ContentControl Content => _contentControl;
+
+        public virtual bool IsOKButtonEnabled
+        {
+            get { return _showOKButton; }
+            protected set { SetValueAndRaise(ref _showOKButton, value); }
+        }
+
+        public virtual IAttachDebuggerStep OnCancelCommand()
+        {
+            return ExitStep;
+        }
+
+        public virtual Task<IAttachDebuggerStep> OnStart()
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual Task<IAttachDebuggerStep> OnOKCommand()
+        {
+            throw new NotImplementedException();
+        }
+        #endregion
+    }
+}
--- a/GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
@@ -26,8 +26,8 @@ namespace GoogleCloudExtension.AttachRemoteDebugger
     {
         private static readonly ContentControl _contentControl = new ContentControl();
 
-        private bool _showOKButton = true;
-        private bool _showCancelButton = true;
+        private bool _isOKButtonEnabled;
+        private bool _isCancelButtonEnabled;
 
         /// <summary>
         /// Define a global exit step object that implements <seealso cref="IAttachDebuggerStep"/> interface.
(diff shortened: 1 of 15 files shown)


GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Windows.Controls;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
+{
+    /// <summary>
+    /// Base class for attaching remote debugger step view models.
+    /// </summary>
+    public class AttachDebuggerStepBase:  ViewModelBase, IAttachDebuggerStep
+    {
+        private static readonly ContentControl _contentControl = new ContentControl();
```
ivannaranjo: Sharing controls is not going to be a good idea. You can re-create them when needed. Also nit, should be `s_contentControl`.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
```diff
@@ -0,0 +1,68 @@
...
+        public static AttachDebuggerStepBase ExitStep { get; } = new AttachDebuggerStepBase();
+
+        #region implement interface IAttachDebuggerStep
+        public virtual bool IsCancelButtonEnabled
+        {
+            get { return _isCancelButtonEnabled; }
+            protected set { SetValueAndRaise(ref _isCancelButtonEnabled, value); }
+        }
+
+        public virtual ContentControl Content => _contentControl;
```
ivannaranjo: Yep, multiple instances sharing the same controls is not a good idea. Make this a lazy field so you only create it on demand, they are not that expensive that you need to do this sharing.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        public virtual async Task<IAttachDebuggerStep> OnCancelCommand()
+        {
+            return ExitStep;
+        }
+
+        public virtual async Task<IAttachDebuggerStep> OnStart()
+        {
+            throw new NotImplementedException();
```
ivannaranjo: This is temporary right? Add a comment to that effect.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowContent.xaml
```diff
@@ -0,0 +1,55 @@
...
+        <theming:CommonDialogWindowBaseContent.Buttons>
+            <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiOkButtonCaption}"
+                                      Command="{Binding OKCommand}" />
+            <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiCancelButtonCaption}"
+                                      Command="{Binding CancelCommand}"
+                                      IsCancel="True" />
+        </theming:CommonDialogWindowBaseContent.Buttons>
+
+        <StackPanel IsEnabled="{Binding IsReady}">
+            <ContentControl Content="{Binding Content}" />
```
ivannaranjo: This really tells me that we need a wizard library for the extension. We have the publishing wizard and now the debugger doing the very same thing.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -0,0 +1,217 @@
...
+            if (step != null && step != ExitStep)
+            {
+                Content = step.Content;
+                step.PropertyChanged += OnStepPropertyChanged;
+
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+                Content.Loaded += (sender, _) 
+                    => ErrorHandlerUtils.HandleExceptionsAsync(
+                        () => ExceuteAsync(
+                            () => OnContentLoaded(sender)));
```
ivannaranjo: I'm sorry, this is weird, consider rethinking this structure. I will wager that you don't need to wait for the Xaml engine to load the content in this way.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -0,0 +1,217 @@
...
+#pragma warning restore CS4014
+            }
+            _currentStep = step;    // set _currentStep if step == ExitStep
+        }
+
+        private async Task OnContentLoaded(object sender)
```
ivannaranjo: If you don't the caller to await this method why does it return `Task`? You can have an `async void` method.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -0,0 +1,217 @@
...
+                (ex is DataSourceException ||
+                 ex is AttachDebuggerException)
+            {
+                // TODO: Add help page or generic error page
+                if (ex is AggregateException)
+                {
+                    UserPromptUtils.ErrorPrompt(
+                        title: Resources.ExceptionPromptTitle,
+                        message: String.Format(Resources.ExceptionPromptMessage, ex.InnerException.Message),
+                        errorDetails: ex.InnerException.ToString());
```
ivannaranjo: We do not show exception messages in the UI, this is once of our UI core tenets. Use a string to show this message to the user, if you cannot detect the error situation then use a generic string. The strings coming from the exceptions are useful for us the developers but not our users.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -0,0 +1,217 @@
...
+                }
+                else
+                {
+                    UserPromptUtils.ErrorPrompt(
+                        title: Resources.ExceptionPromptTitle,
+                        message: String.Format(Resources.ExceptionPromptMessage, ex.Message),
+                        errorDetails: ex.ToString());
```
ivannaranjo: Same thing, this is a nono. Don't show the strings coming from the exceptions.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -0,0 +1,217 @@
...
+                _currentStep = ExitStep;
+            }
+            finally
+            {
+                IsReady = true;
+                ShowProgressIndicator = false;
+                UpdateButtons();
+            }
+
+            if (_currentStep == ExitStep)
```
ivannaranjo: This is more something that the ExitStep itself should do, rather than comparing against it I think.

GoogleCloudExtension/GoogleCloudExtension/Utils/ProtectedCommand.cs
```diff
@@ -45,7 +58,7 @@ public ProtectedCommand(Action handler, bool canExecuteCommand = true)
 
         public void Execute(object parameter)
         {
-            _action.Invoke();
+            _action?.Invoke();
```
ivannaranjo: Can legally `_action` be null? I think that we should crash in this case, to know of the bug.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
Deren-Liao: I did not mean to share it.  It's going to be used by dummy ExitStep only.  I'll refactor the class as abstract to avoid the confusion.

GoogleCloudExtension/GoogleCloudExtension/Utils/ProtectedCommand.cs
Deren-Liao: Reverted.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerStepBase.cs
```diff
@@ -22,26 +22,19 @@ namespace GoogleCloudExtension.AttachRemoteDebugger
...
-
         #region implement interface IAttachDebuggerStep
         public virtual bool IsCancelButtonEnabled
         {
             get { return _isCancelButtonEnabled; }
             protected set { SetValueAndRaise(ref _isCancelButtonEnabled, value); }
         }
 
-        public virtual ContentControl Content => _contentControl;
+        public virtual ContentControl Content => null;
```
ivannaranjo: I think you want to do an abstract property, like so:
```C#
public abstract ContentControl Content { get; }
```
It is better that way, since I assume that all steps will have to have an actual `Content` property.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -77,14 +75,12 @@ public ContentControl Content
...
-        public AttachDebuggerWindowViewModel(
-            AttachDebuggerWindow owner,
-            Instance gceInstance)
+        public AttachDebuggerWindowViewModel(Instance gceInstance)
         {
-            _owner = owner;
             _gceInstance = gceInstance;
             OKCommand = new ProtectedCommand(taskHandler: () => ExceuteAsync(OnOKCommand), canExecuteCommand: false);
             CancelCommand = new ProtectedCommand(taskHandler: () => ExceuteAsync(OnCancelCommand), canExecuteCommand: false);
+            GotoStep(null);     // Add initial step.
```
ivannaranjo: This seems weird, is `null` a special value of the initial step? Why not specify the initial step here?

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -95,7 +91,10 @@ public ContentControl Content
                 return;
             }
             IAttachDebuggerStep nextStep = await _currentStep.OnOKCommand();
-            GotoStep(nextStep);
+            if (nextStep != null)
```
ivannaranjo: You can simplify this by allowing the steps to call `GotoStep()` instead. That way you don't need to do this weird `null` handling. The steps can just call to navigate if needed, otherwise no changes in the navigation stack are perfomed.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -141,26 +143,34 @@ private void GotoStep(IAttachDebuggerStep step)
                 _currentStep.PropertyChanged -= OnStepPropertyChanged;
             }
 
-            if (step != null && step != ExitStep)
+            if (step != null)
             {
                 Content = step.Content;
                 step.PropertyChanged += OnStepPropertyChanged;
 
+                Content.Loaded += OnContentLoaded;
```
ivannaranjo: This is the bit we need to talk about, I do not think that you actually need to listen to the UI events for this. You can do this simply by having your mini wizard framework call a method at the right time for the initialization of the step to happen.

The UI events are fickle and reserved for things that need to ensure that the screen is updated, which is not a concern of the view model.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -141,26 +143,34 @@ private void GotoStep(IAttachDebuggerStep step)
...
+        }
+
+        private void OnContentLoaded(object sender, System.Windows.RoutedEventArgs e)
+        {
 #pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
-                Content.Loaded += (sender, _) 
-                    => ErrorHandlerUtils.HandleExceptionsAsync(
-                        () => ExceuteAsync(
-                            () => OnContentLoaded(sender)));
+            var action = new Action(() => ErrorHandlerUtils.HandleExceptionsAsync(
```
ivannaranjo: This #pragma is actually not needed, see my PR #604

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -141,26 +143,34 @@ private void GotoStep(IAttachDebuggerStep step)
...
+                () => ExceuteAsync(() => OnStart(sender))));
 #pragma warning restore CS4014
-            }
-            _currentStep = step;    // set _currentStep if step == ExitStep
+
+            // A tricky way to perform tasks after the user control is rendered.
```
ivannaranjo: That's why you shouldn't do this, let's talk.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
Deren-Liao: Initial step will be added later.

Deren-Liao:
Even using "the trick" did not work .   I thought it worked , turned out I left the Task.Delay in the test code.   Using thread does not help much.  I am removing that part of code.  I'll think of other ways later.

Deren-Liao:
Thank you Ivan for doing the review.

PR #601, opened 2017-05-09
M2 gcs browser refactoring

ivannaranjo:
With this PR I am refactoring the file operations that the GCS browser depends on to a new separate library. This library is independent on the UI. This will also simplify greatly the view model for the GCS browser moving forward.

This is part of a UI cleanup work that I am doing as I finish this work item.


GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
```diff
@@ -0,0 +1,253 @@
...
+                .Where(x => x.IsFile)
+                .Select(x => new GcsFileOperation(
+                    localPath: Path.Combine(destinationDir, GcsPathUtils.GetFileName(x.Name)),
+                    gcsItem: x)));
+
+            // Collect all of the files from all of the directories to download, collects all of the
+            // local file system directories that need to be created to mirror the source structure.
+            var subDirs = new HashSet<string>();
+            foreach (var dir in sources.Where(x => x.IsDirectory))
+            {
```
ILMTitan: It feels like the `await` in this loop could be parallelized. If you move the loop contents into a separate method, you could collect the resultant tasks, and then do an `await Task.WhenAll(...)`.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// <param name="sources">The list of sources to delete.</param>
+        /// <param name="cancellationToken">The cancellation token to use for the operations.</param>
+        /// <returns>The list of operations started.</returns>
+        public async Task<IList<GcsFileOperation>> StartDeleteOperationsAsync(
+            IEnumerable<GcsItemRef> sources,
+            CancellationToken cancellationToken)
+        {
+            List<GcsFileOperation> deleteOperations = new List<GcsFileOperation>();
+
+            deleteOperations.AddRange(sources
```
ILMTitan: Again, this could be done in the constructor.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// Creates the <seealso cref="GcsFileOperation"/> necessary to upload all of the sources from the local
+        /// file system to GCS. The sources can be files or directories.
+        /// </summary>
+        /// <param name="sources">The list of local files or directories to upload.</param>
+        /// <returns></returns>
+        static private IList<GcsFileOperation> CreateUploadOperations(IEnumerable<string> sources, string bucket, string bucketPath)
+        {
+            return sources
+               .Select(src =>
+               {
```
ILMTitan: I am not a fan of an anonymous function this big in the middle of a set of statements. Perhaps refactor it into a method?

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// </summary>
+        /// <param name="sourceDir">The local dir to process.</param>
+        /// <param name="bucket">The name of the bucket.</param>
+        /// <param name="baseGcsPath">The base gcs path where to copy the files.</param>
+        static private IEnumerable<GcsFileOperation> CreateUploadOperationsForDirectory(
+            string sourceDir,
+            string bucket,
+            string baseGcsPath)
+        {
+            return Enumerable.Concat(
```
ILMTitan: Split the two arguments into local variables.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/GcsPathUtils.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </summary>
+    public static class GcsPathUtils
+    {
+        /// <summary>
+        /// Returns the leaf on a GCS path.
+        /// </summary>
+        /// <param name="name">The path within a bucket.</param>
+        public static string GetFileName(string name)
+        {
+            var cleanName = name.Last() == '/' ? name.Substring(0, name.Length - 1) : name;
```
ILMTitan: `name.Last()` will throw on an empty string.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/GcsItemRef.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// Wether this item is a directory.
+        /// </summary>
+        public bool IsDirectory => String.IsNullOrEmpty(Name) || Name.Last() == '/';
+
+        /// <summary>
+        /// Whether this item is file.
+        /// </summary>
+        public bool IsFile => !IsDirectory;
```
ILMTitan: I guess we don't care about poorly named storage objects with names that end with `/`.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -157,30 +159,24 @@ public void StartDroppedFilesUpload(IEnumerable<string> files)
...
-            {
-                _dataSource.StartFileUploadOperation(
-                    sourcePath: operation.Source,
-                    bucket: operation.Bucket,
-                    name: operation.Destination,
-                    operation: operation,
-                    token: tokenSource.Token);
-            }
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+            var uploadOperations = _fileOperationsEngine.StartUploadOperations(
```
ILMTitan: What is the type of uploadOperations?

GoogleCloudExtension/GoogleCloudExtension/GcsFileProgressDialog/GcsFileProgressDialogWindow.cs
```diff
@@ -42,14 +43,14 @@ public class GcsFileProgressDialogWindow : CommonDialogWindowBase
         /// <param name="caption">The caption to use for the dialog.</param>
         /// <param name="message">The message to use in the dialog.</param>
         /// <param name="operations">The list of operations to track.</param>
-        /// <param name="tokenSource">The <seealso cref="CancellationTokenSource"/> to be used to cancel the operations.</param>
+        /// <param name="cancellationTokenSource">The <seealso cref="CancellationTokenSource"/> to be used to cancel the operations.</param>
```
ILMTitan: I am not sure of the benefits of the longer name. I felt the shorter name was perfectly clear.

GoogleCloudExtension/GoogleCloudExtension.DataSources/GcsDataSource.cs
```diff
@@ -206,7 +206,7 @@ public Task<IList<Bucket>> GetBucketListAsync()
             {
                 operation.Error(new DataSourceException(ex.Message, ex));
             }
-            catch (TaskCanceledException)
+            catch (TaskCanceledException ex)
```
ivannaranjo: Oops, left over from a debugging session. Will fix it for this round but you raise an interesting question about warnings, the fact that they are ignorable. I'll create a PR to set TreatWarningsAsErrors in appveyor and fix the build. That way we will not have this issue anymore.

GoogleCloudExtension/GoogleCloudExtension.Deployment/GoogleCloudExtension.Deployment.csproj.user
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <ProjectView>ShowAllFiles</ProjectView>
+    <ProjectView>ProjectFiles</ProjectView>
```
ivannaranjo: I think that this makes sense, but I will do this in a separate cleaning PR on master. I don't want to do this sort of general thing on a separate feature branch.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// <param name="bucketPath">The destination path in the bucket.</param>
+        /// <param name="cancellationToken">The cancellation token source to use for the operations.</param>
+        /// <returns>The list of operations started.</returns>
+        public IList<GcsFileOperation> StartUploadOperations(
+            IEnumerable<string> sources,
+            string bucket,
+            string bucketPath,
+            CancellationToken cancellationToken)
+        {
+            var uploadOperations = CreateUploadOperations(sources, bucket: bucket, bucketPath: bucketPath);
```
ivannaranjo: I take that you are not a fan of the `var` statement then? It is what `CreateUploadOperations()` returns which is some sort of enumerable type. In short, it doesn't matter as long as the type can be enumerated. That is what I am specifying in the code.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/GcsItemRef.cs
ivannaranjo: True, those objects will be considered directory markers. That's what a directory marker is actually, a blob that ends in /.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Hover over the variable :).

GoogleCloudExtension/GoogleCloudExtension/GcsFileProgressDialog/GcsFileProgressDialogWindow.cs
ivannaranjo: Fully specifying is always better in my opinion, which is why I fully specify what it is. Longer names are not bad, intellisense is there to type them out for you.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// <param name="cancellationToken">The cancellation token for the operation.</param>
+        /// <returns>The list of operations started.</returns>
+        public async Task<IList<GcsFileOperation>> StartDownloadOperationsAsync(
+            IEnumerable<GcsItemRef> sources,
+            string destinationDir,
+            CancellationToken cancellationToken)
+        {
+            List<GcsFileOperation> downloadOperations = new List<GcsFileOperation>();
+
+            downloadOperations.AddRange(sources
```
ivannaranjo: Good point, done.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
ivannaranjo: Very good point, done.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
ivannaranjo: Done.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
ivannaranjo: Done, good point actually. I let the anonymous function grow too much.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/GcsPathUtils.cs
ivannaranjo: True, but that is a bad path that should not have gone through this function. I have changed it to throw an ArgumentException so we know when this happen in a more reliable way.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: Maybe we should have a bigger discussion over the use of `var`. It seems that you don't particularly agree with it in these cases.

ivannaranjo:
PTAL

ivannaranjo:
Please PTAL.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/FileOperationsEngine.cs
ILMTitan: I believe, when using `var`, you should not have to look more than a couple of lines to figure out the type. If it takes any more work, that is just wasted mental effort that could be fixed by making things more explicit.

I think you could use one of the `var` declarations here, but not both `var uploadOperations = ...` and ` var operation in ...`.

PR #604, opened 2017-05-10
Remove all warnings, enable warnings as errors on appveyor

ivannaranjo:
This PR enables the `TreatWarningsAsErrors` setting to fail the build on compiler warnings. The PR also contain the necessary changes to eliminate the existing warnings from the code base.

The decision to only enable this setting on the CI server is to allow low friction local development. You can always duplicate the setting locally before sending the PR or let the CI system build the code when submitting the PR.


Deren-Liao:
The article  [here](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx?f=255&MSPPError=-2147217396) talks about "don't use async void" . 
Everything else looks good.

ivannaranjo:
Yes, but that means that we need to await all async methods. The methods that I change are changed that way because they are not awaited, which means their result didn't matter. If that is not the case then we need to make sure that we await them and do something with the result.

ivannaranjo:
I'm going to enable this as well for the local development, talking with others realized that waiting for the CI build is counter productive.

Issue #574, opened 2017-04-19
Google app settings should be grouped into a section of appsettings.json.


SurferJeffAtGoogle:
I created a new GoogleAspNetCoreMvc application and the appsettings looked like this:
```json
{
  "ProjectId": "YOUR-PROJECT-ID",
  "Logging": {
    "IncludeScopes": false,
    "LogLevel": {
      "Default": "Debug",
      "System": "Information",
      "Microsoft": "Information"
    }
  },
  "GoogleErrorReporting": {
    "ServiceName": "NAME-OF-YOUR-SERVICE",
    "Version": "VERSION-OF-YOUR-SERVICE"
  }
}
```

Here's how I want it to look:
```json
{
  "Google": {
    "ProjectId": "YOUR-PROJECT-ID",
    "ErrorReporting": {
      "ServiceName": "NAME-OF-YOUR-SERVICE",
      "Version": "VERSION-OF-YOUR-SERVICE"
    }
  },
  "Logging": {
    "IncludeScopes": false,
    "LogLevel": {
      "Default": "Debug",
      "System": "Information",
      "Microsoft": "Information"
    }
  }
}
```

PR #607, opened 2017-05-11
Project template cleanup

ILMTitan:
This PR adds the Template projects to the solution. It adds a wizard that will popup a dialog to let the user input a project id that will be set in the template.


ivannaranjo:
Please fix the build, it seems you might be missing a file from your commits.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -0,0 +1,140 @@
...
+            foreach (PackageReference package in _packages)
+            {
+                allTasks.Add(InstallPackageAsync(project, package));
+            }
+            await Task.WhenAll(allTasks);
+        }
+
+        private async Task InstallPackageAsync(Project project, PackageReference package)
+        {
+            await Task.Run(() => _installer.InstallPackage(null, project, package.Id, package.Version, true));
```
ivannaranjo: Are you certain that the `InstallPackage` is thread safe? Can we install multiple packages at the same time without corrupting files?

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -0,0 +1,140 @@
...
+            foreach (XElement packageElement in packageElements)
+            {
+                string id = packageElement.Attribute("id")?.Value;
+                string version = packageElement.Attribute("version")?.Value;
+                yield return new PackageReference(id, version);
+            }
+        }
+    }
+
+    internal class PackageReference
```
ivannaranjo: Add class comments. Consider moving the class to its own file.

GoogleCloudExtension/GoogleCloudExtension/Theming/CommonDataTemplates.xaml
```diff
@@ -0,0 +1,17 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:crm="clr-namespace:Google.Apis.CloudResourceManager.v1.Data;assembly=Google.Apis.CloudResourceManager.v1">
+
+    <!-- Data template for the GCP project. -->
+    <DataTemplate DataType="{x:Type crm:Project}">
+        <TextBlock>
+            <TextBlock.Text>
+                <MultiBinding StringFormat="{}{0} ({1}:{2})">
```
ivannaranjo: Actually, move this to the resources. All string formatting should really be in the resources so we can change it directly. I feel you, but it is better to centralize it that way. You can do this formatting in the view model and reference the right string from the resources.

GoogleCloudExtension/GoogleCloudExtension/Utils/DataSourceUtils.cs
```diff
@@ -0,0 +1,23 @@
+using Google.Apis.Auth.OAuth2;
```
ivannaranjo: Missing licenses.

GoogleCloudExtension/GoogleCloudExtension/Utils/DataSourceUtils.cs
```diff
@@ -0,0 +1,23 @@
...
+using GoogleCloudExtension.Accounts;
+using GoogleCloudExtension.DataSources;
+
+namespace GoogleCloudExtension.Utils
+{
+    public static class DataSourceUtils
```
ivannaranjo: Class comment.

GoogleCloudExtension/GoogleCloudExtension/Utils/DataSourceUtils.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public static ResourceManagerDataSource CreateResourceManagerDataSource()
```
ivannaranjo: Method comment.

ivannaranjo: Also, not sure that this code is duplicated enough to grant its own method. This seems to me a bit of overkill.

GoogleCloudExtension/GoogleCloudExtension/Utils/NotifyTaskCompletion.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System.ComponentModel;
```
ivannaranjo: Missing licenses.

GoogleCloudExtension/GoogleCloudExtension/Utils/NotifyTaskCompletion.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.Utils
+{
+    public class NotifyTaskCompletion : INotifyPropertyChanged
```
ivannaranjo: Missing class comment.

ivannaranjo: This class looks very similar to [`AsyncPropertyValue`](https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/blob/master/GoogleCloudExtension/GoogleCloudExtension.Utils/AsyncPropertyValue.cs). Can we re-use it?

GoogleCloudExtension/GoogleCloudExtension/packages.config
```diff
@@ -32,9 +32,14 @@
   <package id="Microsoft.VisualStudio.Shell.Interop.9.0" version="9.0.30729" targetFramework="net452" />
   <package id="Microsoft.VisualStudio.TextManager.Interop" version="7.10.6070" targetFramework="net452" />
   <package id="Microsoft.VisualStudio.TextManager.Interop.8.0" version="8.0.50727" targetFramework="net452" />
-  <package id="Microsoft.VisualStudio.Threading" version="14.1.131" targetFramework="net452" />
+  <package id="Microsoft.VisualStudio.Threading" version="14.1.111" targetFramework="net452" />
```
ivannaranjo: Is this downgrade expected?

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Startup.cs
```diff
@@ -68,22 +68,17 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
 
         private string GetProjectId()
         {
-            string projectId = Configuration["ProjectId"];
-            if (projectId == ("YOUR-PROJECT-ID"))
+            var instance = Google.Api.Gax.Platform.Instance();
+            var projectId =
+                instance.GceDetails?.ProjectId ??
+                instance.GaeDetails?.ProjectId ??
+                Configuration["ProjectId"];
```
ivannaranjo: This is the wrong order. If the user specifies the project in the configuration file then it should be used. The app should only fetch the _ambient_ project id if no other source for the project id is found. This is so the user can actually override the project id if necessary/wanted.

ivannaranjo: The user can have multiple configuration files to control this.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/appsettings.json
```diff
@@ -1,5 +1,5 @@
 ﻿{
-  "ProjectId": "YOUR-PROJECT-ID",
+  "ProjectId": "$gcpprojectid$",
```
ivannaranjo: Since you are fixing this, you should also consider fixing #574 and move all Google specific values to its own section.

ivannaranjo: We should not force the user to have a value here however. This is only for debugging when running locally. This setting should be considered as an override for the ambient value, not a fallback.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/appsettings.json
```diff
@@ -9,7 +9,7 @@
     }
   },
   "GoogleErrorReporting": {
-    "ServiceName": "NAME-OF-YOUR-SERVICE",
-    "Version": "VERSION-OF-YOUR-SERVICE"
+    "ServiceName": "$safeprojectname$",
```
ivannaranjo: Why not detect the `ServiceName` as well? In other words, let the code determine the `ServiceName` from the project ID **unless** the user specifies a particular `ServiceName`.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvcVs2017/Startup.cs
```diff
@@ -68,22 +68,17 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
 
         private string GetProjectId()
         {
-            string projectId = Configuration["ProjectId"];
-            if (projectId == ("YOUR-PROJECT-ID"))
+            var instance = Google.Api.Gax.Platform.Instance();
```
ivannaranjo: Obviously, same comment as before, the order of precedence is wrong.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/App_Start/FilterConfig.cs
```diff
@@ -11,36 +11,29 @@ public class FilterConfig
...
             // [START error_reporting]
             // Check to ensure that projectId has been changed from placeholder value.
             var section = (XmlElement)ConfigurationManager.GetSection("log4net");
-            XmlElement element =
+            var projectIdElement =
                 (XmlElement)section.GetElementsByTagName("projectId").Item(0);
-            string projectId = element?.Attributes["value"].Value;
-            if (projectId == ("YOUR-PROJECT-ID"))
+            string projectId =
+                Google.Api.Gax.Platform.Instance().GceDetails?.ProjectId ??
```
ivannaranjo: Same as before, the local setting (in the config file) should override the ambient value. The order that you have is therefore wrong.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/Web.config
```diff
@@ -21,8 +21,8 @@
            Project id if you want to run on local When running on
            Google Cloud the code will be able to determine the project
            ID automatically. -->
-      <!-- <projectId value="YOUR-PROJECT-ID" /> -->
-      <logId value="YOUR-LOG-NAME" />
+       <projectId value="$gcpprojectid$" />
```
ivannaranjo: Leave this commented out, that way the user can uncomment it if they wish to override the project ID detection.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/Web.config
```diff
@@ -40,8 +40,8 @@
     <add key="ClientValidationEnabled" value="true" />
     <add key="UnobtrusiveJavaScriptEnabled" value="true" />
     <!-- [START google_error_reporting_config] -->
-    <add key="google_error_reporting:serviceName" value="NAME-OF-YOUR-SERVICE" />
-    <add key="google_error_reporting:version" value="VERSION-OF-YOUR-SERVICE" />
+    <add key="google_error_reporting:serviceName" value="$safeprojectname$" />
```
ivannaranjo: Same comment as before, if the suggested `servicename` is a function on the project ID then leave this blank and commented and auto detect it in the code. If the user sets this value then they will be overriding the detection code.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/App_Start/WebApiConfig.cs
```diff
@@ -18,34 +18,27 @@ public static class WebApiConfig
     {
         public static void Register(HttpConfiguration config)
         {
-	    // To enable Google Cloud Stackdrive Logging and Error Reporting
-	    // while running on your local machine edit Web.config and uncomment
-	    // the <projectId> value under the <log4net> section. Ensure that
-	    // the <projectId> is set to a valid Google Cloud Project Id.
+            // To enable Google Cloud Stackdrive Logging and Error Reporting
+            // while running on your local machine edit Web.config and uncomment
+            // the <projectId> value under the <log4net> section. Ensure that
```
ivannaranjo: See my previous comment on this, the project ID value in the .config file is not currently commented out but it should be.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/App_Start/WebApiConfig.cs
```diff
@@ -18,34 +18,27 @@ public static class WebApiConfig
...
             // [START logging_and_error_reporting]
             // Check to ensure that projectId has been changed from placeholder value.
             var section = (XmlElement)ConfigurationManager.GetSection("log4net");
-            XmlElement element =
+            XmlElement projectIdElement =
                 (XmlElement)section.GetElementsByTagName("projectId").Item(0);
-            string projectId = element?.Attributes["value"].Value;
-            if (projectId == ("YOUR-PROJECT-ID"))
+            string projectId =
+                Google.Api.Gax.Platform.Instance().GceDetails?.ProjectId ??
```
ivannaranjo: Again this order is wrong.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/Web.config
```diff
@@ -5,39 +5,56 @@
...
   <configSections>
-    <!-- For more information on Entity Framework configuration, visit http://go.microsoft.com/fwlink/?LinkID=237468 -->
-    <section name="entityFramework" type="System.Data.Entity.Internal.ConfigFile.EntityFrameworkSection, EntityFramework, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" requirePermission="false" />
-    <!-- // [START log4net_web_config] -->
+    <!--
+      For more information on Entity Framework configuration, visit http://go.microsoft.com/fwlink/?LinkID=237468
+    -->
+    <section
+      name="entityFramework"
+      type="System.Data.Entity.Internal.ConfigFile.EntityFrameworkSection,
```
ivannaranjo: Why? This is not the standard Xml formatting that Microsoft uses.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/Web.config
```diff
@@ -5,39 +5,56 @@
...
       <appender-ref ref="CloudLogger" />
     </root>
   </log4net>
-  <!-- // [END log4net_web_config] -->
+  <!-- [END log4net_web_config] -->
   <connectionStrings>
-    <add name="DefaultConnection" connectionString="Data Source=(LocalDb)\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\aspnet-$safeprojectname$-20160523013341.mdf;Initial Catalog=aspnet-$safeprojectname$-20160523013341;Integrated Security=True"
+    <add
+      name="DefaultConnection"
+      connectionString="Data Source=(LocalDb)\MSSQLLocalDB;
```
ivannaranjo: Please no, use the standard Xml formatting.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/Web.config
```diff
@@ -53,7 +70,12 @@
       <remove name="ExtensionlessUrlHandler-Integrated-4.0" />
       <remove name="OPTIONSVerbHandler" />
       <remove name="TRACEVerbHandler" />
-      <add name="ExtensionlessUrlHandler-Integrated-4.0" path="*." verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />
+      <add
```
ivannaranjo: No, please, standard formatting.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        private IEnumerable<PackageReference> GetPackageRefernces(string vsTemplatePath)
+        {
+            XDocument vsTemplate = XDocument.Load(vsTemplatePath);
+            if (vsTemplate.Root == null)
+            {
+                throw new InvalidOperationException();
+            }
+            IEnumerable<XElement> dataElement =
+                vsTemplate.Root.Elements().Where(e => e.Name.LocalName == "WizardData");
```
ivannaranjo: This doesn't appear to be used in the PR. Is this a future feature not yet fullfilled? Are the packages not already installed in the templates themselves?

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+            : this(owner, DataSourceUtils.CreateResourceManagerDataSource, ManageAccountsWindow.PromptUser) { }
+
+        /// <summary>
+        /// For Testing.
+        /// </summary>
+        /// <param name="owner">The window that owns this ViewModel.</param>
+        /// <param name="dataSourceFactory">The factory of the source of projects.</param>
+        /// <param name="promptAccountManagement">Action to prompt managing accounts.</param>
+        internal PickProjectIdViewModel(
+            IPickProjectIdWindow owner, Func<IResourceManagerDataSource> dataSourceFactory, Action promptAccountManagement)
```
ILMTitan: Because during testing, you can mock an interface, but can't mock a class.

GoogleCloudExtension/GoogleCloudExtension.Interop/IGoogleProjectTemplateWizard.cs
```diff
@@ -0,0 +1,5 @@
+﻿using Microsoft.VisualStudio.TemplateWizard;
+
+namespace GoogleCloudExtension.TemplateWizards {
+    public interface IGoogleProjectTemplateWizard : IWizard { }
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/ResourceManagerDataSource.cs
```diff
@@ -22,11 +22,25 @@
 
 namespace GoogleCloudExtension.DataSources
 {
+    public interface IResourceManagerDataSource
```
ILMTitan: Done.
Yes, it is for mocking.

GoogleCloudExtension/GoogleCloudExtension.Interop/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension.TemplateWizards/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Reflection;
```
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -0,0 +1,124 @@
+﻿using Google.Apis.CloudResourceManager.v1.Data;
```
ILMTitan: Done

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+using GoogleCloudExtension.ManageAccounts;
+using GoogleCloudExtension.Utils;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.TemplateWizards.Dialogs.ProjectIdDialog
+{
+    public class PickProjectIdViewModel : ViewModelBase
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        public string Result { get; private set; }
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        public NotifyTaskCompletion LoadTask
+        {
+            get { return _loadTask; }
+            set { SetValueAndRaise(ref _loadTask, value); }
+        }
+
+        /// Property backing field.
+        private IList<Project> _projects;
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+            _promptAccountManagement = promptAccountManagement;
+            ChangeUserCommand = new ProtectedCommand(OnChangeUser);
+            SelectProjectCommand = new ProtectedCommand(OnSelectProject, false);
+            SkipProjectInputCommand = new ProtectedCommand(OnSkip);
+            LoadTask = new NotifyTaskCompletion(LoadProjectsAsync());
+        }
+
+        private async Task LoadProjectsAsync()
+        {
+            Projects = await _resourceManagerDataSourceFactory().GetProjectsListAsync();
```
ILMTitan: Yes. The main part of the dialog shows a "Loading Projects..." message. That is replaced when the load is complete, either with an error message if the load failed or was canceled, or the list of projects from which to select.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+            if (string.IsNullOrEmpty(ProjectId) || ProjectId.Equals(SelectedProject?.ProjectId))
```
ILMTitan: Done with the first part. We can consider the second part later.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindow.cs
```diff
@@ -0,0 +1,27 @@
+﻿using GoogleCloudExtension.Theming;
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindow.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace GoogleCloudExtension.TemplateWizards.Dialogs.ProjectIdDialog
+{
+    public interface IPickProjectIdWindow
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindow.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        void Close();
+    }
+
+    public class PickProjectIdWindow : CommonDialogWindowBase, IPickProjectIdWindow
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
```diff
@@ -0,0 +1,112 @@
...
+                            Binding LoadTask.IsFaulted, Converter={
+                                utils:BooleanConverter
+                                    TrueValue={x:Static Visibility.Visible}, FalseValue={x:Static Visibility.Collapsed}
+                            }
+                        }"
+                        VerticalAlignment="Center"
+                        HorizontalAlignment="Center"/>
+                    <TextBlock
+                        Text="{x:Static ext:Resources.WizardPickProjectIdCanceledTask}"
+                        Visibility="{
```
ILMTitan: I have converted the attribute to nested elements to format under 120 characters that way.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Xml.Linq;
+using Task = System.Threading.Tasks.Task;
+
+namespace GoogleCloudExtension.TemplateWizards
+{
+    /// <summary>
+    /// Wizard for a project template.
+    /// </summary>
+    [Export(typeof(IGoogleProjectTemplateWizard))]
+    public class GoogleProjectTemplateWizard : IGoogleProjectTemplateWizard
```
ILMTitan: You are correct. I added comments to the interface. Given that this bit of code cleared that up for you, I am not changing any comments on this implementation.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -0,0 +1,140 @@
...
+        public void RunStarted(
+            object automationObject,
+            Dictionary<string, string> replacementsDictionary,
+            WizardRunKind runKind,
+            object[] customParams)
+        {
+            string vsTemplatePath = customParams.FirstOrDefault() as string;
+            if (vsTemplatePath != null)
+            {
+                _packages = GetPackageRefernces(vsTemplatePath);
```
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        ///<inheritdoc />
+        public void RunStarted(
+            object automationObject,
+            Dictionary<string, string> replacementsDictionary,
+            WizardRunKind runKind,
+            object[] customParams)
+        {
+            string vsTemplatePath = customParams.FirstOrDefault() as string;
```
ILMTitan: The official NuGet Template Wizard does it this way. I didn't actually see a reference for what customParams includes.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
ILMTitan: I am not using any of this code. I am deleting all of the Package code.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
ILMTitan: It is not. Deleting code.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
ILMTitan: Deleted

GoogleCloudExtension/GoogleCloudExtension/Theming/CommonDataTemplates.xaml
ILMTitan: I am moving the StringFormat to the resources. I do want keep the CommonDataTemplates.xaml, as it still contains code we would otherwise repeat and that we probably want to keep in sync.

GoogleCloudExtension/GoogleCloudExtension/Utils/DataSourceUtils.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/Utils/DataSourceUtils.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/appsettings.json
ILMTitan: Surfacing offline discussion:
$safeprojectname$ is the name of the Visual Studio project, not the GCP project, and is probably a good default for the ServiceName.

GoogleCloudExtension/GoogleCloudExtension/Utils/NotifyTaskCompletion.cs
ILMTitan: Rewrote it to make it the base class of AsyncPropertyValue<T>.

GoogleCloudExtension/GoogleCloudExtension/packages.config
ILMTitan: Resetting everything back to .131.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Startup.cs
ILMTitan: I think this is the correct order. Taking discussion offline.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/Web.config
ILMTitan: This is needed if we want the templates runnable immediately after creation. Part of the whole project Id precedence offline discussion.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/Web.config
ILMTitan: Undone. All too-long attributes are back to one line.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/Web.config
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/Web.config
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/IResourceManagerDataSource.cs
```diff
@@ -0,0 +1,37 @@
...
+using Google.Apis.CloudResourceManager.v1.Data;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.DataSources
+{
+    /// <summary>
+    /// The interface of the ResourceManagerDataSource. Mock this object in unit tests.
+    /// </summary>
+    public interface IResourceManagerDataSource
```
ivannaranjo: I like this, we should consider doing this for the rest of the sources at some point.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
```diff
@@ -57,47 +57,60 @@
...
-                                utils:BooleanConverter
-                                    TrueValue={x:Static Visibility.Collapsed}, FalseValue={x:Static Visibility.Visible}
-                            }
-                        }"
                         VerticalAlignment="Center"
-                        HorizontalAlignment="Center"/>
+                        HorizontalAlignment="Center">
+                        
+                        <TextBlock.Visibility>
+                            <Binding Path="LoadTask.IsCompleted">
```
ivannaranjo: Nit and more of a personal preference, use the `{Binding ...}` expression is in my opinion cleaner, but this works too.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
```diff
@@ -57,47 +57,60 @@
...
+                                <Binding.Converter>
+                                    <utils:BooleanConverter
```
ivannaranjo: There is a dedicated `VisiblityConverter` for this, as it is a very common need for the boolean converter.

GoogleCloudExtension/GoogleCloudExtension.Utils/AsyncProperty.cs
```diff
@@ -0,0 +1,157 @@
...
+    {
+        /// <summary>
+        /// Returns true if the wrapped task is in error.
+        /// </summary>
+        public bool IsError => _task?.IsFaulted ?? false;
+
+        /// <summary>
+        /// Gets a relevant exception error message, if it exists.
+        /// </summary>
+        public string ErrorMessage =>
```
ivannaranjo: We should not use error messages coming from exceptions, always use strings coming from the resources. You could specify the string to use when you create the `AsyncProperty` in the constructor, have a default one in case no string is specified.

GoogleCloudExtension/GoogleCloudExtension.Utils/AsyncProperty.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// </summary>
+        public T Value { get; private set; }
+
+        /// <summary>
+        /// Returns a task that will be completed once the wrapped task is completed. This task is
+        /// not directly connected to the wrapped task and will never throw and error.
+        /// </summary>
+        public Task<T> ValueTask => _completionSource.Value.Task;
+
+        public AsyncProperty(Task<T> valueSource, T defaultValue = default(T)) : base(valueSource, false)
```
ivannaranjo: This arrangement seems a bit strange, the base class and the derived class both have separate tasks for the same "logical" task. I propose an idea. Have an abstract class, say `AsyncPropertyBase`, which contains all of the common functionality and an abstract method that returns a `Task`. Derived classes `AsyncProperty` and `AsyncProperty<T>` can have their own storage for the `Task`. The base class can do the basic stuff while the derived class will the specific one, like having the `Value` property.

I am worried about this business with the `TaskCompletionSource`, it really adds quite a bit of complexity to what is otherwise a very simple class. Quite frankly I would rather have the `AsyncProperty` and `AsyncProperty<T>` be two similar but distinct classes instead.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -39,52 +84,51 @@ public Project SelectedProject
...
-                SelectProjectCommand.CanExecuteCommand = !string.IsNullOrEmpty(ProjectId);
+                OkCommand.CanExecuteCommand = !string.IsNullOrEmpty(ProjectId);
             }
         }
 
-        public NotifyTaskCompletion LoadTask
+        /// <summary>
+        /// The property that surfaces task completion information for the Load Projects task.
+        /// </summary>
+        public AsyncProperty LoadTask
```
ivannaranjo: The `LoadTask` property doesn't really change during the life of the instance, so you could just have a readonly property for it, avoid the field as well.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -99,17 +143,17 @@ public PickProjectIdViewModel(IPickProjectIdWindow owner)
             }
             else
             {
-                SelectedProject = null;
+                SelectedProject = Projects.FirstOrDefault(p => p.ProjectId.Equals(ProjectId));
```
ivannaranjo: In C# the string == works "the right way", no need to use `Equals()`.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
```diff
@@ -57,47 +57,60 @@
...
-                                    TrueValue={x:Static Visibility.Visible}, FalseValue={x:Static Visibility.Collapsed}
-                            }
-                        }"
                         VerticalAlignment="Center"
-                        HorizontalAlignment="Center"/>
+                        HorizontalAlignment="Center">
+                        
+                        <TextBlock.Visibility>
+                            <Binding Path="LoadTask.IsError">
+                                <Binding.Converter>
```
ivannaranjo: Same as before, use `VisiblityConverter`.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/appsettings.json
```diff
@@ -8,8 +7,11 @@
       "Microsoft": "Information"
     }
   },
-  "GoogleErrorReporting": {
-    "ServiceName": "$safeprojectname$",
-    "Version": "1.0.0"
+  "Google": {
```
ivannaranjo: Love this.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
ILMTitan: It changes because the list of projects is reloaded when the user changes the active account.

GoogleCloudExtension/GoogleCloudExtension.Utils/AsyncProperty.cs
ILMTitan: I created the abstract base class.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.Utils/AsyncProperty.cs
ILMTitan: I am changing my dialog to have a generic error message, but I am going to leave this here. I think at some point we need to let the user see the error message from the API, as that can be more useful than anything we can put in the resources.

ILMTitan:
PTAL

GoogleCloudExtension/GoogleCloudExtension.Utils/VisibilityConverter.cs
```diff
@@ -40,17 +40,22 @@ public class VisibilityConverter : MarkupExtension, IValueConverter
...
-        /// The prefix string to use for the log messages, useful for finding the entries in the 
+        /// The prefix string to use for the log messages, useful for finding the entries in the
         /// Output window.
         /// </summary>
         public string LoggingPrefix { get; set; }
 
+        /// <summary>
+        /// When set to true, hides the element rather than collapsing it.
+        /// </summary>
+        public bool Hide { get; set; } = false;
```
ivannaranjo: Collapsed is usually better because it doesn't take part on the layout. What is the case for Hide?

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
```diff
@@ -58,58 +58,28 @@
...
+
                     <TextBlock
                         Text="{x:Static ext:Resources.WizardPickProjectIdCanceledTask}"
                         VerticalAlignment="Center"
-                        HorizontalAlignment="Center">
+                        HorizontalAlignment="Center"
+                        Visibility="{Binding LoadTask.IsCanceled, Converter={utils:VisibilityConverter}}"/>
+                    <ScrollViewer
+                        VerticalScrollBarVisibility="Auto"
+                        Visibility="{Binding LoadTask.IsSuccess, Converter={utils:VisibilityConverter Hide=True}}">
```
ivannaranjo: Not sure I agree with this change, why not collapse the ScrollViewer instead? Do you still want it to occupy space?

GoogleCloudExtension/GoogleCloudExtension/source.extension.vsixmanifest
```diff
@@ -6,7 +6,7 @@
     accurate metrics.
     -->
     <Identity Id="GoogleAppEngine.Google.d3d3eeb8-3710-4bd9-97ba-1401bf2acd22" Version="1.2.3.0" Language="en-US" Publisher="Google Inc." />
-    <DisplayName>Google Cloud Tools for Visual Studio</DisplayName>
+    <DisplayName>ct-vs</DisplayName>
```
ivannaranjo: Why are you changing the DisplayName of the extension at ALL?!?!?

ILMTitan: A workaround for the path lengths bug, allowing me to deploy the extension using the MSBuild tasks.

PR #609, opened 2017-05-15
Master sorting organzing usings formatting warnings

ivannaranjo:
This PR was created just by running "Sort and organize usings" in the entire solution, followed by running the .NET code formatter. With this PR the code is cleaner.

In this PR I have also turned on warnings and errors on the project files themselves, that way we will see the warnings right away instead of having to wait for the appveyor build to finish.


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/Gae/VersionViewModel.cs
```diff
@@ -147,7 +147,7 @@ private void OnBrowseStackdriverLogCommand()
                 IsLoading = true;
                 Caption = String.Format(Resources.CloudExplorerGaeMigratingAllTrafficCaption, _version.Id);
 
-                var split = new TrafficSplit { Allocations = new Dictionary<string, double?> { [_version.Id] = 1.0 } };
+                var split = new TrafficSplit { Allocations = new Dictionary<string, double?> {[_version.Id] = 1.0 } };
```
ILMTitan: I'm not sure I agree with this change.

ivannaranjo: Ah, the code formatter got confused. Good eye, will fix this.

PR #614, opened 2017-05-16
Add steps of enable debugger port and remote powershell port

Deren-Liao:
* It firstly test if the firewall rule is there
* Add the firewall rule if it is not there
* Then it tests if local machine can "telnet" to the ip:port
* Add rule one by one, such that we allow user to manually install/start remote debugger tool without requiring to use remote powershell.


GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+using GoogleCloudExtension.Utils;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Net.Sockets;
+using System.Threading.Tasks;
+using System.Windows.Controls;
+using static GoogleCloudExtension.Utils.ArgumentCheckUtils;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
```
ivannaranjo: Nit, this is not a very good namespace name. Maybe `RemoteDebugger` is a better name for this. This is a problem because this is an action phrase, which is good for a method but not for a class or naemspace.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+{
+    /// <summary>
+    /// This class contains common data, methods for attach remote debugger steps.
+    /// </summary>
+    internal class AttachDebuggerContext
+    {
+        private const int ConnectivityTestTimeout = 2000;   // In milliseconds
+        private const int FirewallRuleWaitMaxTime = 5 * 60;    // 5 minutes
+
+        private static Lazy<AttachDebuggerContext> s_Instance;
```
ivannaranjo: Nit, should be `s_instance`.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+using static GoogleCloudExtension.Utils.ArgumentCheckUtils;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
+{
+    /// <summary>
+    /// This class contains common data, methods for attach remote debugger steps.
+    /// </summary>
+    internal class AttachDebuggerContext
+    {
+        private const int ConnectivityTestTimeout = 2000;   // In milliseconds
```
ivannaranjo: Why not use a `TimeSpan` instead? It is clear what it is and you can use it later on to get the time in any unit that you need.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        /// <summary>
+        /// The GCE VM <seealso cref="Instance"/> object.
+        /// </summary>
+        public Instance GceInstance { get; private set; }
+
+        /// <summary>
+        /// The GCE VM instance public ip address.
+        /// </summary>
+        public string PublicIp { get; }
```
ivannaranjo: Isn't this part of the GceInstance? Instead of duplicating it you could get it from it. I think that there's an extension method to do it for you.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        /// <summary>
+        /// Create the <seealso cref="Context"/> to associate the <paramref name="gceInstance"/>.
+        /// </summary>
+        /// <param name="gceInstance">GCE VM instance object.</param>
+        /// <param name="dialogWindow">The dialog window</param>
+        public static void CreateContext(Instance gceInstance, AttachDebuggerWindow dialogWindow)
+        {
+            s_Instance = new Lazy<AttachDebuggerContext>(() => new AttachDebuggerContext(gceInstance, dialogWindow));
```
ivannaranjo: Why does this need to be a singleton? Why not just a regular dialog?

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// The GCE VM instance public ip address.
+        /// </summary>
+        public string PublicIp { get; }
+
+        /// <summary>
+        /// Create the <seealso cref="Context"/> to associate the <paramref name="gceInstance"/>.
+        /// </summary>
+        /// <param name="gceInstance">GCE VM instance object.</param>
+        /// <param name="dialogWindow">The dialog window</param>
+        public static void CreateContext(Instance gceInstance, AttachDebuggerWindow dialogWindow)
```
ivannaranjo: I don't think we should have singleton Windows like this. We should create them when we need them and let them be destroyed afterwards.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// False: Failed to connect.
+        /// </returns>
+        public async Task<bool> ConnectivityTest(int port)
+        {
+            using (TcpClient client = new TcpClient())
+            {
+                try
+                {
+                    var task = client.ConnectAsync(PublicIp, port);
+                    return ((await Task.WhenAny(task, Task.Delay(ConnectivityTestTimeout))) == task);
```
ivannaranjo: The extra parenthesis can be removed I think, as in:
```C#
return (await Task.WhenAny(task, Task.Delay(ConnectivityTestTimeout))) == task;
```
But you could also have a temporary variable, that will make it more readable.

ivannaranjo: I'm worried about what would happen if the `Task.Delay()` call returns first. What happens with the connection task that is still happening in the backgroud? Instead of manufacturing our own timeout, is there a way to specify a timeout in `TcpClient` itself?

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// <summary>
+        /// Add firewall rule to unblock the port to the GCE instance.
+        /// </summary>
+        /// <param name="portInfo">A <seealso cref="PortInfo"/> object that represents the port.</param>
+        public async Task EnablePort(PortInfo portInfo)
+        {
+            portInfo.ThrowIfNull(nameof(portInfo));
+            string portTag = portInfo.GetTag(GceInstance);
+
+            List<FirewallPort> toEnables = new List<FirewallPort>();
```
ivannaranjo: Typo, `toEnable`.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+            toEnables.Add(new FirewallPort(portTag, portInfo.Port));
+            var operation = Context.DataSource.UpdateInstancePorts(
+                GceInstance,
+                portsToEnable: toEnables,
+                portsToDisable: new List<FirewallPort>());
+            await operation.OperationTask;
+
+            _lastPortEnableTime.Add(portInfo.Port, DateTime.Now);
+            GceInstance = await Context.DataSource.RefreshInstance(GceInstance);
```
ivannaranjo: Does the UI need to be invalidated in any way?

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            GceInstance = gceInstance.ThrowIfNull(nameof(gceInstance));
+            DialogWindow = dialogWindow.ThrowIfNull(nameof(dialogWindow));
+            PublicIp = gceInstance.GetPublicIpAddress();
+            _dataSource = new Lazy<GceDataSource>(CreateDataSource);
+        }
+
+        private static T CreateObject<T>(params object[] args)
+        {
+            return (T)Activator.CreateInstance(typeof(T), args);
```
ivannaranjo: This is not good, you should prefer something like an `Action<T>` to create the object. Using reflection for object creation completely breaks the language encapsulation and negates any advantage of having this method. Please don't do this.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Threading.Tasks;
+using System.Windows.Controls;
+using static GoogleCloudExtension.Utils.ArgumentCheckUtils;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
+{
+    /// <summary>
+    /// This class contains common data, methods for attach remote debugger steps.
+    /// </summary>
+    internal class AttachDebuggerContext
```
ivannaranjo: The purpose of thiis class is not really clear to me. It seems that this class is just a mix of UI, view model, etc... I think that this needs a bit more of design work.  I think that you should separate the ports management into its own class, and move the UI parts to a view model.

Also, you are still using static/singletons for UI which is not the right pattern for WPF. Do not reuse windows.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
```diff
@@ -75,12 +74,13 @@ public ContentControl Content
             private set { SetValueAndRaise(ref _content, value); }
         }
 
-        public AttachDebuggerWindowViewModel(Instance gceInstance)
+        public AttachDebuggerWindowViewModel(Instance gceInstance, AttachDebuggerWindow dialogWindow)
         {
-            _gceInstance = gceInstance;
+            AttachDebuggerContext.CreateContext(gceInstance, dialogWindow);
```
ivannaranjo: Why do you need to make `AttachDebuggerContext` a singleton then? Just put the same state in the view model.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnableDebuggerPortStepViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+using static GoogleCloudExtension.VsVersion.VsVersionUtils;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
+{
+    public class EnableDebuggerPortStepViewModel : EnablePortStepViewModel
+    {
+        /// <summary>
+        /// The <seealso cref="PortInfo"/> for Visual Studio Remote Debugger tool port.
+        /// </summary>
+        public static readonly PortInfo s_DebuggerPortInfo =
```
ivannaranjo: Wrong name, `s_debuggerPortInfo`.

ivannaranjo: Also, you should never have public static fields, or any other field. If you need to access this data make it a public static property.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnableDebuggerPortStepViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+            new PortInfo("VSRemoteDebugger", RemoteDebuggerPort, Resources.PortManagerRemoteDebuggerDescription);
+
+        public EnableDebuggerPortStepViewModel(UserControl content)
```
ivannaranjo: Use the specific type for the view that this view model is associated with.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+            protected set { SetValueAndRaise(ref _askingToTestConnectivityLater, value); }
+        }
+
+        /// <summary>
+        /// The message when operation is in progress.
+        /// </summary>
+        public string ProgressMessage
+        {
+            get { return _progressMessage; }
+            set { SetValueAndRaise(ref _progressMessage, value); }
```
ivannaranjo: Should this setter also be protected?

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+                    IsCancelButtonEnabled = true;
+                    IsOKButtonEnabled = true;
+                    IsAskingToTestConnectivityLater = true;
+                    break;
+            }
+        }
+
+        protected enum Stage
+        {
+            Init = -1,
```
ivannaranjo: Set it to 0 instead, that the is the expected default for enums anyway.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+                    break;
+                case Stage.AskToCheckConnectivityLater:
+                    IsCancelButtonEnabled = true;
+                    IsOKButtonEnabled = true;
+                    IsAskingToTestConnectivityLater = true;
+                    break;
+            }
+        }
+
+        protected enum Stage
```
ivannaranjo: Add comment for the enum, what is its purpose.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            Init = -1,
+            CheckingFirewallRule,
```
ivannaranjo: Ideally a line of comment for each value, to know what it means.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// if the GCE instance also has the tag, the rule is applied to the GCE instance.
+        /// </summary>
+        /// <param name="portInfo">A <seealso cref="PortInfo"/> object that represents the port.</param>
+        /// <param name="gceInstance">GCE Instance</param>
+        private async Task<bool> IsPortEnabled(PortInfo portInfo)
+        {
+            string portTag = portInfo.GetTag(Context.GceInstance);
+
+            // If the instance does not contain the tag, the firewall rule is not set.
+            if (Context.GceInstance.Tags?.Items?.Contains(portTag) == true)
```
ivannaranjo: I think that you can simplify this logic a bit like this:
```C#
if (Context.GceInstance.Tags?.Items?.Contains(portTag) ?? false)
{
...
}
```

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+            {
+                var rules = await Context.DataSource.GetFirewallListAsync();
+                foreach (var rule in rules)
+                {
+                    // Left oprand is nullable bool.  null == false is false. null == true is false.
+                    if (!(rule.TargetTags?.Contains(portTag) == true))
```
ivannaranjo: And you can apply the same thing here, the ?? operator will allow you to transform the `null` value into `false`, and then use that in the `if` normally.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+                    {
+                        continue;   // Skip, rules does not contain the tag.
+                    }
+                    foreach (var allowed in rule.Allowed)
```
ivannaranjo: I think that you could simplify this with a LINQ expression.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePowerShellPortStepViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+// limitations under the License.
+
+using GoogleCloudExtension.FirewallManagement;
+using static GoogleCloudExtension.AttachRemoteDebugger.AttachDebuggerContext;
+using System.Threading.Tasks;
+using System.Windows.Controls;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
+{
+    public class EnablePowerShellPortStepViewModel : EnablePortStepViewModel
```
ivannaranjo: Add a class comment.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePowerShellPortStepViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /// <summary>
+        /// The <seealso cref="PortInfo"/> for remote PowerShell HTTPs port.
+        /// </summary>
+        public static readonly PortInfo s_RemotePowerShellPortInfo =
```
ivannaranjo: No public static fields, if you need it make it a public property.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: I firstly used RemoteDebugger, later I felt this sounds like I am implementing a debugger.  I am going to use AttachDebuggerDialog.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: I intend to let this class to save some states cross steps.  
I am going to move all methods out this class and leave only properties to make it clear.  It is not a view, not a view model, just a class that helps keep some states that every step can access.  
It is valid when the dialog is open, and become invalid when the dialog is closed.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: I just caches the GetPublicIpAddress() result.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: No, it does not cache the singleton window, it just keeps the reference for every steps to access.  When the window close, the context become invalid. When a new windows is created, a new context is created.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: The TcpClient is closed and the ConnectAsync task will be stopped.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
ILMTitan: Perhaps RemoteDebug would not have the same debugger implementation connotation?

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: No, Attach debugger dialog does not need to refresh.  In the last step, the Cloud Explorer instance node will be refreshed.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: Every step has CreateStep method, I wanted to create a generic method to simplifies the steps. Then it requires Activator for creating Template types.  I agree with you, I'll make the change.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: I tried. Using LINQ makes my code harder to read.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: N/A

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: N/A

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerWindowViewModel.cs
Deren-Liao: N/A

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnableDebuggerPortStepViewModel.cs
Deren-Liao: I knew, just did not pay attention.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnableDebuggerPortStepViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+    {
+        // TODO: update the link when we have the doc ready.
+        private const string EnablePortHelpLink = "https://cloud.google.com/tools/visual-studio/docs/how-to";
+
+        private bool _portEnabled;
+        private string _progressMessage;
+        private bool _askingToConfirm;
+        private bool _askingToTestConnectivityLater;
+        private PortInfo _portInfo;
+        private Stage _stage;
```
ILMTitan: `_stage` is never used.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// The message when operation is in progress.
+        /// </summary>
+        public string ProgressMessage
+        {
+            get { return _progressMessage; }
+            set { SetValueAndRaise(ref _progressMessage, value); }
+        }
+
+        /// <summary>
+        /// Initializes the <seealso cref="EnableDebuggerPortViewModel"/>
```
ILMTitan: Should this be `EnablePortStepViewModel`?

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+
+        public override Task<IAttachDebuggerStep> OnCancelCommand()
+        {
+            Context.DialogWindow.Close();
+            return Task.FromResult<IAttachDebuggerStep>(null);    // TODO:  return help page as next step.
+        }
+
+        public override async Task<IAttachDebuggerStep> OnOKCommand()
```
ILMTitan: Nit: Should Be `OnOkCommand`

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+            AskToCheckConnectivityLater,
+        }
+
+        /// <summary>
+        /// Check if GCE firewall rules include a rule that enables the port to target GCE VM.
+        /// A firewall rule contains tag, 
+        /// if the GCE instance also has the tag, the rule is applied to the GCE instance.
+        /// </summary>
+        /// <param name="portInfo">A <seealso cref="PortInfo"/> object that represents the port.</param>
+        /// <param name="gceInstance">GCE Instance</param>
```
ILMTitan: The param `gceInstance` doesn't exist.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+            CheckingFirewallRule,
+            AskingPermitToAddRule,
+            AddingFirewallRule,
+            CheckingConnectivity,
+            AskToCheckConnectivityLater,
+        }
+
+        /// <summary>
+        /// Check if GCE firewall rules include a rule that enables the port to target GCE VM.
+        /// A firewall rule contains tag,
```
ILMTitan: Nit: Extra space.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePowerShellPortStepViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public EnablePowerShellPortStepViewModel(UserControl content)
+            : base(content, s_RemotePowerShellPortInfo)
+        { }
+
+        /// <summary>
+        /// Create the the step that enables Visual Studio remote debugging tool port.
+        /// </summary>
+        public static IAttachDebuggerStep CreateStep() => 
+            Context.CreateStep<EnablePowerShellPortStepViewModel, EnablePortStepContent>();
```
ILMTitan: Again, just use constructors here directly.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/AttachDebuggerContext.cs
Deren-Liao: I thought only array can do initialization this way. Done.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnableDebuggerPortStepViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+            : base(content, s_DebuggerPortInfo)
+        { }
+
+        /// <summary>
+        /// Create the the step that enables Visual Studio remote debugging tool port.
+        /// </summary>
+        public static IAttachDebuggerStep CreateStep() => 
+            Context.CreateStep<EnableDebuggerPortStepViewModel, EnablePortStepContent>();
```
Deren-Liao: There are many steps. I just wanted to share the code.  I've refactored the code that removed this.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepContent.xaml
```diff
@@ -0,0 +1,51 @@
...
+             d:DesignHeight="150" d:DesignWidth="320"
+             Width="320">
+
+    <UserControl.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="../Theming/CommonResources.xaml" />
+            </ResourceDictionary.MergedDictionaries>
+
+            <utils:VisibilityConverter x:Key="visibilityConverter" />
```
Deren-Liao: We've been using lower case for other xamls.  I am making the change here.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepContent.xaml
```diff
@@ -0,0 +1,51 @@
...
+            <utils:NullEmptyInvisibleConverter x:Key="nullInvisibleConverter" />
+
+        </ResourceDictionary>
+    </UserControl.Resources>
+    
+    <Grid>
+        <TextBlock 
+            Margin="0,6,0,6"
+            Text="{Binding ProgressMessage}"
+            Visibility="{Binding ProgressMessage, Converter={StaticResource nullInvisibleConverter} }" />
```
Deren-Liao: I just follows the convention.  I think the advantage I can think of is that if we later update the implementation of the converter, it's easier to replace at one place.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepContent.xaml.cs
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+using System.Windows.Controls;
+
+namespace GoogleCloudExtension.AttachRemoteDebugger
+{
+    /// <summary>
+    /// Interaction logic for EnablePortStepContent.xaml
+    /// </summary>
+    public partial class EnablePortStepContent : UserControl
```
Deren-Liao: By default it is appended by Visual studio I think.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: removed

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: Updated

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: Done.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePortStepViewModel.cs
Deren-Liao: N/A after code refactoring.

GoogleCloudExtension/GoogleCloudExtension/AttachRemoteDebugger/EnablePowerShellPortStepViewModel.cs
Deren-Liao: Changed

Deren-Liao:
Please take a look again. I did the code refactoring as requested.

GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerFirewallPort.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// Checks if current machine can establish TCP connection to the remote applicaton.
+        /// The test succeeds only if 
+        ///   (a) The network is connected (i.e no firewall blocks it),
+        ///   (b) The target application is started and listening at the TCP port.
+        /// </summary>
+        /// <returns>
+        /// True: Local machine is able to connect to the target GCE VM at the port number.
+        /// False: Failed to connect.
+        /// </returns>
+        public async Task<bool> ConnectivityTest()
```
ILMTitan: I feel like this method could be written much cleaner if written synchronously:
```
using(...)
{
    try
    {
         return client.ConnectAsync(...).Wait(ConnectivityTestTimeout);
    }
    catch
    {
        return false;
    }
}
```

Deren-Liao: Yes, using synchronous make it easier.  We need it be async so that the UI can show progressing, otherwise the UI looks hang.

ILMTitan:
```
using(...)
{
    try
    {
         var task = client.ConnectAsync(...);
         if(task == await Task.WhenAny(task, Task.Delay(...)))
         {
             await task;
             return true;
        }
        else
        {
            return false;
        }
    }
    catch
    {
        return false;
    }
}
```

GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerFirewallPort.cs
```diff
@@ -0,0 +1,163 @@
...
+            using (TcpClient client = new TcpClient())
+            {
+                Exception exception = null;
+                Func<Task<bool>> connectAsync = async () =>
+                {
+                    try
+                    {
+                        await client.ConnectAsync(_gceInstance.GetPublicIpAddress(), PortInfo.Port);
+                        Debug.WriteLine("ConnectivityTest, Succeeded");
+                        return true;
```
ILMTitan: Where is this return value used?

Deren-Liao: connectTask.Result

GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerFirewallPort.cs
Deren-Liao: This is better. Updated.

Deren-Liao:
Hi Jim, I updated the ConnectivityTest code as your suggestion. Please take a look again when you get a chance.

Deren-Liao:
Thank you Ivan and Jim for doing the review!

PR #617, opened 2017-05-17
Update the main GCS UI 

ivannaranjo:
This PR updates the main UI of the GCS browser to better match the UX design, by:
* Removing the whitespace around the `DataGrid`.
* Making the widths of the columns relatives to the overall width of the grid.
  + This allows the filenames to be longer than the rest of the columns.
* Formatting the size of the files in human readable form.
  + This is using the Win32 API [`StrFormatByteSizeW`](https://msdn.microsoft.com/en-us/library/windows/desktop/bb759975%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396).
  + This uses the same formatting that the Windows Explorer does.


GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsFileBrowserWindowControl.xaml
```diff
@@ -124,20 +126,53 @@
...
-                <DataGridTextColumn Binding="{Binding ContentType}"
-                                    utils:DataGridBehaviors.CustomSort="{local:GcsColumnSorter ContentType}"
-                                    Header="{x:Static ext:Resources.GcsFileBrowserContentTypeHeader}" />
+                <DataGridTemplateColumn Width="*"
+                                        SortMemberPath="Size"
+                                        utils:DataGridBehaviors.CustomSort="{local:GcsColumnSorter Size}"
+                                        Header="{x:Static ext:Resources.GcsFileBrowserSizeHeader}">
+                    <DataGridTemplateColumn.CellTemplate>
+                        <DataTemplate>
+                            <TextBlock Text="{Binding FormattedSize}"
```
Deren-Liao: nit: You may create a style for all the textblock that share this same textTriming , textWrapping settings.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringFormatUtils.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
```
Deren-Liao: Add license header.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringFormatUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Threading.Tasks;
+
+namespace GoogleCloudExtension.Utils
+{
+    /// <summary>
+    /// This class contains helpers for formatting strings.
+    /// </summary>
+    public static class StringFormatUtils
+    {
+        private const int MaxStringSize = 20;
```
Deren-Liao: As a utility class, internally limit string size may surprise other users if it is used by others in the future.  You can add it as input parameter and set the size limit by caller.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringFormatUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        /// <param name="size">The size in bytes to format.</param>
+        /// <returns>The human readable string, for example 3KB, 45 bytes, etc...</returns>
+        public static string FormatByteSize(ulong size)
+        {
+            StringBuilder sb = new StringBuilder(MaxStringSize);
+            var result = StrFormatByteSize(size, sb, sb.Capacity);
+            if (result == IntPtr.Zero)
+            {
+                throw new InvalidOperationException($"Failed to convert value: {size}");
```
Deren-Liao: What input can cause this error?   size is too large?  < 0? 
If that's case, it might be more appropriate to throw ArgumentOutOfRange Exception.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringFormatUtils.cs
ivannaranjo: So true, will be done.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringFormatUtils.cs
ivannaranjo: True, done.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringFormatUtils.cs
ivannaranjo: Basically a size so big that it cannot be formatted by the method. I think that you are right, `ArgumentOutOfRange` is more appropriate. Done.

ivannaranjo:
Tests are not appropriate here because I didn't implement the algorithm for the `StrFormatByteSizeW` function. However I will look into adding a smoke test to make sure that small and large values do go through.

ivannaranjo:
Added the tests, PTAL.

PR #623, opened 2017-05-19
Manage concurrent operations for GCS operations.

ivannaranjo:
This PR adds a new class to manage concurrent operations for GCS operations. Before the PR the extension would try to queue all of the operations in the GCS service at once. This will end up queuing all of the operations in the HttpClient underlying the GCS service, which will in turn completely overload the thread pool that is used to run async tasks. If the user is, for example, uploading a large number of files this would cause the process to pause as the operations are queued. Because the timeout starts the moment the operations are queued a lot of the operations would then be cancelled because it took a long time to get to them.

With this PR we now only queue as many operations as allowed by the `ServicePointManager.DefaultConnectionLimit` setting. This setting controls how many simultaneous connections the HttpClient can do to the same endpoint. This way we help ensure that all of the operations are being acted on. The operations that are waiting do not have their timeouts started, which means that they will not be spuriously cancelled. The user can still cancel the operations as all of the started operations will use the same `CancellationToken`. The `OperationsQueue` calls will also check this `CancellationToken` and will not start new operations if the user requested cancellation.

The `OperationsQueue` class works by keeping track of many operations are started and ensuring that no more than the limit are started at any given point. This is all done in an async loop, waiting for operations to complete and adding new operations as new slots are made available. This class could be more general, but at the moment only GCS operations require this treatment.


GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
```diff
@@ -0,0 +1,164 @@
...
+                        var completed = await Task.WhenAny(_operationsInFlight);
+                        _operationsInFlight.Remove(completed);
+                    }
+
+                    // Starts the next operation in the queue.
+                    Debug.WriteLine("Queueing next operation.");
+                    var nextOperation = _pendingOperations.Dequeue();
+                    _operationsInFlight.Add(nextOperation.ExecuteOperationAsync(_cancellationToken));
+                }
+                Debug.WriteLine("Ending operations scheduler.");
```
ILMTitan: I am concerned that a thread could stop here, then another thread could queue up some operations but never get to the while loop (because _scheduleActive is still true). Then this thread would come and exit without running the new tasks. The user would see this simply as the tasks hanging.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// Queue the given list of <paramref name="operations"/> but does not start them yet.
+        /// </summary>
+        /// <param name="operations">The operations to start.</param>
+        /// <param name="startOperationAction">The action to use to start the operaitons.</param>
+        public void EnqueueOperations(
+            IEnumerable<GcsFileOperation> operations,
+            Action<GcsFileOperation, CancellationToken> startOperationAction)
+        {
+            _operations.AddRange(operations);
+            foreach (var operation in operations)
```
ILMTitan: Resharper: Multiple enumeration of IEnumerable.
We can either append to _operations in the loop, or store `operations.ToList()` to a local variable and use that.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
```diff
@@ -0,0 +1,164 @@
...
+        private readonly List<Task> _operationsInFlight = new List<Task>();
+        private bool _schedulerActive = false;
+
+        /// <summary>
+        /// Returns the list of operations contained in the queue.
+        /// </summary>
+        public IReadOnlyList<GcsFileOperation> Operations => _operations;
+
+        public OperationsQueue(CancellationToken cancellationToken)
+            : this(cancellationToken, ServicePointManager.DefaultConnectionLimit)
```
benwulfe: what happens if defaultconnectionlimit changes after the operationsqueue is created?  is there an opportunity to reaadjust at some point

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
```diff
@@ -0,0 +1,164 @@
...
+
+namespace GoogleCloudExtension.GcsUtils
+{
+    /// <summary>
+    /// This class represents a queue of operaations that are to be completed. This class will
+    /// ensure that that the right concurrency level is maintained and no more than the specified
+    /// number of operations are started at any given time. If the overall operation is cancelled any
+    /// non started operations will be cancelled as well.
+    /// </summary>
+    public class OperationsQueue
```
benwulfe: this seems very GCS specific but is really a task management pattern.  
feels like there should be a solution for queuing tasks with the GCS operation management a layer above just flipping cancellationtokens.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -181,7 +181,7 @@ public void StartDroppedFilesUpload(IEnumerable<string> files)
                 caption: Resources.GcsFileBrowserUploadingProgressCaption,
                 message: Resources.GcsFileBrowserUploadingProgressMessage,
                 progressMessage: Resources.GcsFileBrowserUploadingOverallProgressMessage,
-                operations: uploadOperations,
+                operations: uploadOperationsQueue.Operations,
```
benwulfe: Is this a modal dialog you are showing to the user?  This is going to tick them off if they intended to upload a lot of files in the background.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/GcsFileOperation.cs
```diff
@@ -163,5 +164,16 @@ void IGcsFileOperationCallback.Error(DataSourceException ex)
...
         #endregion
+
+        /// <summary>
+        /// Returns a task that will completed once the operation is completed. The error state will not be
+        /// sent to this task.
+        /// </summary>
+        internal Task AwaitOperationAsync()
+        {
+            var taskCompletion = new TaskCompletionSource<int>();
+            this.Completed += (o, e) => taskCompletion.SetResult(0);
```
ivannaranjo: There's no point, all we care about is what `Task` has to offer, whether is completed or not.

ivannaranjo: True, I will run the code formatter over this to remove these.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        /// <summary>
+        /// Queue the given list of <paramref name="operations"/> but does not start them yet.
+        /// </summary>
+        /// <param name="operations">The operations to start.</param>
+        /// <param name="startOperationAction">The action to use to start the operaitons.</param>
+        public void EnqueueOperations(
+            IEnumerable<GcsFileOperation> operations,
+            Action<GcsFileOperation, CancellationToken> startOperationAction)
```
ivannaranjo: That was my initial implementation, I decoupled the operations to allow queuing operations from multiple sources.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
ivannaranjo: Nope, there are no multiple threads operating here. I will add a comment to the class, but this class is meant to run **only** on the UI thread.

ivannaranjo: Quick reminder that the VS extension is **not** multi threaded. We use background threads yet, but all of them send their result back to the UI thread via `Task` instances.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                _schedulerActive = false;
+            }
+        }
+
+        private void CancelPendingOperations()
+        {
+            foreach (var entry in _pendingOperations)
+            {
+                var callback = entry.Operation as IGcsFileOperationCallback;
```
ivannaranjo: No, `GcsFileOperation` main role is to serve as the model for the UI. It implements `IGcsFileOperationCallback` as an implementation detail of how it fetches its events. Therefore I do not want `IGcsFileOperationCallback` members polluting the API surface for `GcsFileOperation`.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
ivannaranjo: Nope, it remains fixed. Same as the built-in concurrence limits in .NET. Concurrence limits are seldom going to change beyond the initialization  phase of the process.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
ivannaranjo: Agreed, but there isn't. And this goes for all of the GCP APIs. We rely entirely on the `HttpClient` to queue the async tasks.

GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
ivannaranjo: There's no background copy functionality. But will think about that for v2.

ivannaranjo: Opened #625 to track the background feature.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
ivannaranjo: Done, true that multiple iterations of the `IEnumerable<T>` can be dangerious.

ivannaranjo:
PTAL

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/OperationsQueue.cs
ivannaranjo: Started a conversation with @jskeet and @chrisdunelm about whether this functionality would make sense at the API layer.

GoogleCloudExtension/GoogleCloudExtension.GcsUtils/GcsFileOperation.cs
ILMTitan: You are right. YAGNI.

PR #629, opened 2017-05-22
Save GCE instance default user to app data 

Deren-Liao:
Also added unit test cases for newly added VS solution options classes.


Deren-Liao:
Thank you Jim.

Issue #576, opened 2017-04-19
Templates should avoid spamming logs.


SurferJeffAtGoogle:
When an application is deployed to app engine, app engine checks /_ah/health multiple times per second.  The result is that the logs are so flooded with /_ah/health requests that it's a PITA to find any useful information in the logs.  Let's stop that.

One solution:
Provide appsettings.Staging.json and appsettings.Production.json with logging flags set to something less verbose like:
```json
  "Logging": {
    "IncludeScopes": false,
    "LogLevel": {
      "Default": "Information",
      "System": "Warning",
      "Microsoft": "Warning"
    }
  },
```

SurferJeffAtGoogle:
/_ah/health requests also flood traces.

iantalarico:
We fixed the trace issue in the client library (https://github.com/GoogleCloudPlatform/google-cloud-dotnet/issues/847).  We should be doing a new release this week and will just need to update the versions in the templates.

ivannaranjo:
Seems like the spam in the logs happens regardless of whether we handle the request or not. Us handling the `/_ah/health` route doesn't change the logging behavior, it just changes the HTTP error code from 404 to 200.

PR #631, opened 2017-05-23
Update NuGet packages on project templates.

ILMTitan:
This fixes issue #576.


GoogleCloudExtension/GoogleCloudExtension/Utils/StringUtils.cs
```diff
@@ -174,5 +175,23 @@ public static int LastNonSpaceIndex(string text)
...
+
+        /// <summary>
+        /// Adds a trailing slash if the directory is missing it.
+        /// </summary>
+        /// <param name="directoryString">The directory path string.</param>
+        /// <returns>The path to the directory with a trailing separator character.</returns>
+        public static string EnsureEndSeparator(string directoryString)
+        {
+            directoryString.ThrowIfNull(nameof(directoryString));
+            if (!directoryString.EndsWith(Path.DirectorySeparatorChar.ToString()) || !directoryString.EndsWith("/"))
```
Deren-Liao: Are you sure  "\"  character works too?

GoogleCloudExtension/GoogleCloudExtension/Utils/StringUtils.cs
```diff
@@ -174,5 +175,23 @@ public static int LastNonSpaceIndex(string text)
...
+            {
+                return directoryString + "/";
```
ivannaranjo: You should be using "\\" actually, since this is Windows.

GoogleCloudExtension/GoogleCloudExtension/source.extension.vsixmanifest
```diff
@@ -6,7 +6,7 @@
     accurate metrics.
     -->
     <Identity Id="GoogleAppEngine.Google.d3d3eeb8-3710-4bd9-97ba-1401bf2acd22" Version="1.2.3.0" Language="en-US" Publisher="Google Inc." />
-    <DisplayName>ct-vs</DisplayName>
+    <DisplayName>Google Cloud Tools for Visual Studio</DisplayName>
```
ivannaranjo: Thank you.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/project.json
```diff
@@ -1,6 +1,6 @@
 ﻿{
   "dependencies": {
-    "Google.Cloud.Diagnostics.AspNetCore": "1.0.0-alpha04",
+    "Google.Cloud.Diagnostics.AspNetCore": "1.0.0-beta10",
```
ivannaranjo: I believe this also requires changes on the template themselves, there were breaking changes between alpha and beta.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringUtils.cs
ILMTitan: Yes, because I am throwing it into a `new Uri()` immediately.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringUtils.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/source.extension.vsixmanifest
ILMTitan: Please note the ShortenVsixDeploymentPath target I added to the bottom of the GoogleCloudExtension.csproj msbuild file.

ILMTitan:
PTAL

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -58,7 +59,14 @@ public class GoogleProjectTemplateWizard : IGoogleProjectTemplateWizard
         ///<inheritdoc />
         public bool ShouldAddProjectItem(string filePath)
         {
-            return true;
+            if ("global.json".Equals(Path.GetFileName(filePath), StringComparison.OrdinalIgnoreCase))
```
ivannaranjo: Actually the file should always be in lowercase, even in Windows, and specially in our templates. This is because on Unix systems (case sensitive) the tooling will look for "global.json" as is. 

Also, consider having a constant for this name.

GoogleCloudExtension/GoogleCloudExtension/Utils/StringUtils.cs
```diff
@@ -186,7 +186,7 @@ public static string EnsureEndSeparator(string directoryString)
             directoryString.ThrowIfNull(nameof(directoryString));
             if (!directoryString.EndsWith(Path.DirectorySeparatorChar.ToString()) || !directoryString.EndsWith("/"))
```
ivannaranjo: You should also use the `Path.DirectorySeparatorChar` here for consistency.

GoogleCloudExtension/GoogleCloudExtension/VsVersion/VsVersionUtils.cs
```diff
@@ -22,7 +22,7 @@ namespace GoogleCloudExtension.VsVersion
     /// </summary>
     internal static class VsVersionUtils
     {
-        private const string VisualStudio2015Version = "14.0";
+        public const string VisualStudio2015Version = "14.0";
         private const string VisualStudio2017Version = "15.0";
```
ivannaranjo: Consider making both constants public, since they are now needed outside of the class.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/GoogleAspnetWebApi.csproj
```diff
@@ -2,11 +2,11 @@
...
   <PropertyGroup>
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+    <MinimumVisualStudioVersion Condition="$(VisualStudioVersion) == '15.0'">15.0</MinimumVisualStudioVersion>
+    <OldToolsVersion Condition="$(VisualStudioVersion) == '15.0'">14.0</OldToolsVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
-    <FileUpgradeFlags>
-    </FileUpgradeFlags>
-    <OldToolsVersion>14.0</OldToolsVersion>
-    <UpgradeBackupLocation />
+    <RestorePackages>false</RestorePackages>
```
ivannaranjo: Why are we not allowing the packages to be restored during the build? Are we now including the NuGet packages in the templates themselves?

GoogleCloudExtension/GoogleCloudExtension/Utils/StringUtils.cs
ILMTitan: Where should I use `Path.DirectorySeparatorChar`? I am using it everywhere except one "/", which is in addition to `Path.DirectorySeparatorChar`.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/GoogleAspnetWebApi.csproj
ILMTitan: We are not including the NuGet packages in the template zip file. We want to resolve them when the template is created. This is the project file for building the template zip file. The project file that is deployed does not include this tag.

ivannaranjo:
Please remove the changes to allow the extension to build on VS 2017, we will stay building only on VS 2015, the official build machine is going to be only VS 2015. This will ensure that we will keep the build consistent, single binary to target both VS 2015 and VS 2017.

This is going to cause a bit of discomfort for us as we keep supporting both VS 2015 and VS 2017, we will move to VS 2017 once we no longer have to support VS 2015.

ILMTitan:
I would really rather not remove those changes.

The official build machine should be on VS 2015. Most of the dev work should be on VS 2015. These changes don't impact that at all. All those changes do is make it possible to quickly iterate with a shorter, automated build->install->debug loop in VS 2017 when that is necessary, such as when testing .Net Core Templates in VS 2017.

ivannaranjo:
The main point is that this changes are not required for the development of the product, they are at best a small convenience. Having them in risks breaking the main build, having to maintain this non-standard build scripts for now on. I do not want to maintain non-standard build scripts from now on. I want to be able to automatically (mostly) update the project when the time comes to move to VS 2017. 

So please, remove these non-essential changes. Next time, please let's discuss _before_ the work starts so we don't have to throw work away.

ILMTitan:
I have disabled building in VS 2017. PTAL.

PR #637, opened 2017-05-24
Close remote PowerShell session at solution close time

Deren-Liao:
A remote PowerShell session is opened to start remote tools. 
The session keeps open so that if user want's to attach debugger again, it just connects.
The session closes in 10 minutes if no debugger attaches. 
Or if solution closes, the session should be closed as well.


GoogleCloudExtension/GoogleCloudExtension.PowerShellUtils/RemoteToolSession.cs
```diff
@@ -45,19 +46,45 @@ public class RemoteToolSession
...
+        /// </summary>
+        public void Stop()
+        {
+            if (!IsStopped)
+            {
+                Debug.WriteLine($"_cancelTokenSource.Cancel() ");
+                _cancelTokenSource.Cancel();
+            }
+            Debug.WriteLine($"_powerShellTask.Wait() ");
+            _powerShellTask.Wait();
```
ILMTitan: Do you actually need to wait for the task to end, or is cancelling it enough?

Deren-Liao:
Thanks!

GoogleCloudExtension/GoogleCloudExtension.PowerShellUtils/RemoteToolSession.cs
ILMTitan: Why do you care if the solution is exits quickly? Don't you only care if there is a zombie process? Or are you waiting on a dialog (in which case why are you not doing the handling at the view model level)?

Deren-Liao: Wait for the PowerShell object is closed.  When Powershell is closed, it closes the remote PowerShell session. Otherwise, it leave the remote PS session open while VS is closed. 
There is no guarantee it always shut remote PS successfully, the VS can get crash etc. 
But with this code, in most of the normal cases, as I tested, it can always does the job.

PR #653, opened 2017-06-01
Reverting out to ref.

ivannaranjo:
The parameter to the `SetValueAndRaise()` method is a **ref** parameter. We are passing the _pointer_ to the variable. The method does **not** produce a value into its parameter. We need to use the right semantics for C# here.

I have reverted the change, and I am ashamed to say that I didn't see this on the review. I will be more careful in the future.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Utils/Model.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Utils/Model.cs
@@ -33,7 +33,7 @@ namespace GoogleCloudExtension.Utils
         /// <param name="storage">Typically a ref to a field where the value is stored.</param>
         /// <param name="value">The new value.</param>
         /// <param name="propertyName">The name of the property that is changing, do not specify let the compiler determine it.</param>
-        protected void SetValueAndRaise<T>(out T storage, T value, [CallerMemberName] string propertyName = "")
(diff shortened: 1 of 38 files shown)


ILMTitan:
When this parameter is a ref, resharper gives the warning [Redundant Assignment](https://www.jetbrains.com/help/resharper/2017.1/RedundantAssignment.html).

Because the initial value is never read, it is in essence producing a value. `out` restricts the implementation; the implementation must the parameter as uninitialized and requires an assignment before the method returns. Our method meets both of those criteria. In return `out` provides extra functionality to callers (sending uninitialized varaibles etc.).

Why would we forego the extra functionality when our implementation meets the criteria?

ivannaranjo:
Because `out` is used for methods that _return_ something on them. This is not what the `SetValueAndRaise` does, it does not return anything. The method is given the storage where to store the value, that's it.

I don't care what Resharper is saying, `out` is not the right annotation for this.

ivannaranjo:
I suggest you turn off that warning on Resharper for the `SetValueAndRaise()` method, since it clearly is not getting it right.

ivannaranjo:
Updated to fix the conflicts. PTAL.

ivannaranjo:
Thank you.

PR #667, opened 2017-06-06
M2 gcs browser context menu

### LINKED ISSUE #661: Fix context menu handling on GCS browser.
Currently the context menu handling is too generic. The context menu is set at the `DataGrid` level. We need to set it at the _item_ level, that way we can have specific menu items for each item. 

We should also have a global context menu set in the `DataGrid` but only as a fallback. If no other context menu is found that is the one that should be used. This will contain things like creating new directories, refresh, etc... And will be used when the user right-clicks on the empty space in the `DataGrid`.

ivannaranjo:
This PR contains new behaviors that make the use of context menu in the GCS browser more useable. 

Now you can:
* Right click on an item and if it is not selected it will become the selected item.
* Right click on an already selected item, this will preserve the selection and show the pertinent context menu.
* Right click on the background of the data grid, which allows you to create new directories and select all items.

Fixes #661


GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsFileBrowserWindowControl.xaml
```diff
@@ -96,19 +96,27 @@
...
-
         <!-- The main area for the data. -->
         <DataGrid Grid.Row="2"
                   Grid.ColumnSpan="2"
+                  x:Name="_dataGrid"
                   IsReadOnly="True"
                   AutoGenerateColumns="False"
                   HeadersVisibility="Column"
                   utils:DataGridBehaviors.HasCustomSort="True"
                   ContextMenu="{Binding ContextMenu}"
```
ILMTitan: This binding doesn't exist anymore..

ivannaranjo:
Thank you for for the review.

ivannaranjo:
Thank you for the review.

PR #668, opened 2017-06-06
 M2 gcs browser toolbar buttons

ivannaranjo:
This PR adds a toolbar button to create new directories and some small tweaks into the name prompt window.


ivannaranjo:
Thank you for the review.

ivannaranjo:
Thank you for the review.

PR #677, opened 2017-06-08
Install git prompt

Deren-Liao:
This is to add a prompt asking user to install Git for Windows if git is not found.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.GitUtils/VSGitData.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.GitUtils/VSGitData.cs
@@ -62,7 +62,7 @@ namespace GoogleCloudExtension.GitUtils
             {
                 using (var key = OpenGitKey(gitKeyPath, "Repositories"))
                 {
-                    return key.GetSubKeyNames()
+                    return key?.GetSubKeyNames()
                         .Select(x =>
                         {
                             using (var subkey = key.OpenSubKey(x))
--- a/GoogleCloudExtension/GoogleCloudExtension/Resources.resx
+++ b/GoogleCloudExtension/GoogleCloudExtension/Resources.resx
@@ -2151,4 +2151,16 @@
     <value>Version name</value>
     <comment>Field display name for the version name textbox. Should match PublishDialogFlexVersionNameCaption.</comment>
   </data>
+  <data name="GitUtilsGitInstallLinkCaption" xml:space="preserve">
+    <value>Install Git for Windows</value>
+    <comment>Caption for the installation link</comment>
+  </data>
+  <data name="GitUtilsMissingGitErrorMessage" xml:space="preserve">
+    <value>Please follow the link to install Git for Windows</value>
+    <comment>The error message that asks to install Git for Windows</comment>
+  </data>
+  <data name="GitUtilsMissingGitErrorTitle" xml:space="preserve">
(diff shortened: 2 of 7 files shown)


GoogleCloudExtension/GoogleCloudExtension.GitUtils/VSGitData.cs
```diff
@@ -62,7 +62,7 @@ private static IEnumerable<string> RepositoryList(string gitKeyPath)
             {
                 using (var key = OpenGitKey(gitKeyPath, "Repositories"))
                 {
-                    return key.GetSubKeyNames()
+                    return key?.GetSubKeyNames()
```
ILMTitan: Perhaps coalesce with an empty list?
```
.ToList() ?? new List<...>();
```

GoogleCloudExtension/GoogleCloudExtension/Utils/ValidateGitDependencyHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+    public static class ValidateGitDependencyHelper
+    {
+        /// <summary>
+        /// Asking to install Git for Windows if git.exe is not found
+        /// </summary>
+        /// <returns>
+        /// True: It is installed properly.
+        /// Otherwise, returns false.
+        /// </returns>
+        public static bool ValidateGitForWindowsInstalled()
```
ILMTitan: Does this need to be its own file? It is a single method used in a single place. It could be moved to that class, or even inlined entirely.

Deren-Liao: The CSR feature will need to test git installation too. This is for sharing the code with CSR feature.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/VSGitData.cs
Deren-Liao: Don't bother to do so. In the bottom, it is already returning null.

Deren-Liao:
Thanks for the comment, please take a look again.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2151,4 +2151,16 @@
     <value>Version name</value>
     <comment>Field display name for the version name textbox. Should match PublishDialogFlexVersionNameCaption.</comment>
   </data>
+  <data name="GitUtilsGitInstallLinkCaption" xml:space="preserve">
+    <value>Install Git for Windows</value>
+    <comment>Caption for the installation link</comment>
+  </data>
+  <data name="GitUtilsMissingGitErrorMessage" xml:space="preserve">
+    <value>Please follow the link to install Git for Windows</value>
```
ivannaranjo: I think that we need a bit more of information on how to install git, perhaps point to a documentation page? The problem is that the user can install Git for windows but not make git available to CMD.exe, which means it will not available in the PATH. They need to do that or your feature won't work.

Of course this is only for VS 2015, since in VS 2017 we require Git for Windows installer, but..

Deren-Liao: Let me think through and see what's the best way.  For now, I'd check in this code for a short term fix.

Issue #483, opened 2017-03-01
New Google projects with logging don't actually log anything.


SurferJeffAtGoogle:
Although logging is configured, no code actually seems to log anything.  Logging home page visits would give nice instant gratification.

PR #678, opened 2017-06-08
Add logging of home page visits.

ILMTitan:
Fix for issue #483.

### DIFF
--- a/GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
+++ b/GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
@@ -17,7 +17,10 @@ namespace $safeprojectname$.Controllers
 
          public IActionResult Index()
         {
-            _logger.LogDebug("Home page hit!");
+            // Sends a message to configured loggers, including the Stackdriver logger.
+            // The Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker logger will log all controller actions with
+            // log level information. This log is for additional information.
+            _logger.LogInfo("Home page hit!");
             return View();
         }
 
@@ -35,7 +38,8 @@ namespace $safeprojectname$.Controllers
 
         public IActionResult Error()
         {
-            _logger.LogWarning("Error page hit!");
+            // Log messages with different log levels.
+            _logger.LogError("Error page hit!");
             return View();
         }
     }
--- a/GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
+++ b/GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
@@ -8,27 +8,34 @@ namespace $safeprojectname$.Controllers
 {
     public class HomeController : Controller
     {
-        public IActionResult Index()
+        private readonly ILogger _logger;
+
+        public HomeController(ILogger<HomeController> logger)
         {
+            _logger = logger;
+        }
+
+         public IActionResult Index()
+        {
+            _logger.LogDebug("Home page hit!");
             return View();
         }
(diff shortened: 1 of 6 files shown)


GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
```diff
@@ -8,27 +8,34 @@ namespace $safeprojectname$.Controllers
 {
     public class HomeController : Controller
     {
-        public IActionResult Index()
+        private readonly ILogger _logger;
+
+        public HomeController(ILogger<HomeController> logger)
```
Deren-Liao: I don't see where this constructor is called? 
The last two classes, you use readonly private member.    What's the reason you choose different ways ?

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
```diff
@@ -8,27 +8,34 @@ namespace $safeprojectname$.Controllers
...
         public IActionResult Contact()
         {
             ViewData["Message"] = "Your contact page.";
-
             return View();
         }
 
         public IActionResult Error()
         {
+            _logger.LogWarning("Error page hit!");
```
ivannaranjo: Shouldn't this be LogError?

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
```diff
@@ -8,27 +8,34 @@ namespace $safeprojectname$.Controllers
...
+        private readonly ILogger _logger;
+
+        public HomeController(ILogger<HomeController> logger)
+        {
+            _logger = logger;
+        }
+
+         public IActionResult Index()
         {
+            _logger.LogDebug("Home page hit!");
```
ivannaranjo: Maybe a bit more specific, the class and method? A comment to show the user that they can log and that this log will go to Stackdriver will also be good.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
ILMTitan: ASP.NET Core is injecting this type automatically. The two different ways are because ASP.NET uses Log4Net, while ASP.NET Core uses Microsoft​.Extensions​.Logging.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
ILMTitan: This is someone hitting the error page, which does not necessarily indicate a problem with the service. The user may have just added an invalid url. Thus log level error does not make sense. Many of these, however, may indicate a problem such as a broken link. Therefore log level warning is appropriate.

...

I should add a comment about all of that.

ivannaranjo: Still and error though. Is not a crash or anything, that would go through error reporting. But yes, please add comments about how to use the logging, mark it so users can remove it if they don't want it, etc...

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Controllers/HomeController.cs
ILMTitan: This is mostly being used as an example to show "Hey, you can log things!". The Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker logger already does most of the rest.

Issue #681, opened 2017-06-09
Add option to re-start file operations that were stopped.


ivannaranjo:
Currently once the operations are stopped in the file progress dialog there's no way to continue or restart the pending operations.

Provide a way for the user to continue the operations instead of having to restart the operations from the beginning.

PR #682, opened 2017-06-09
M2 gcs browser ux review

ivannaranjo:
This PR fixes the comments provided by the UX review:
* Provide more commands on the _command bar_, which now incorporated commands for:
  + Uploading files.
  + Deleting files.
  + Creating folder
* Rename the button from _Cancel_ to _Stop_. It makes it clearer that the operation has not be cancelled but stopped.
  + We don't provider, yet, any function to re-start the operations, see #681
* Remove the cancelled indication, it was deemed to be redundant. It is very clear from the state of the dialog that the operations are cancelled. It is also very clear which operations are completed and which ones aren't.


GoogleCloudExtension/GoogleCloudExtension/GcsFileBrowser/GcsBrowserViewModel.cs
```diff
@@ -349,6 +338,49 @@ private void OnDoubleClickCommand(GcsRow row)
             }
         }
 
+        private void OnUploadFilesCommand()
+        {
+            var dialog = new System.Windows.Forms.OpenFileDialog();
+            dialog.Multiselect = true;
```
ILMTitan: R#: This can be an object initializer.

PR #685, opened 2017-06-12
Analytics events for PubSub, Error viewer and Logs viewer

ivannaranjo:
This PR adds analytics events for the PubSub, Error viewer and Logs viewer features. For the Error viewer and Logs viewer it adds events for the data loading with timing, since those are going to potentially take a long time.


ivannaranjo:
I moved the event to the `LogViewerToolWindow` constructor instead. We should think about tracking _where_ the invocation is coming from.

PR #689, opened 2017-06-13
Add views

Deren-Liao:
Add the repository list view and the un-connect view.


Deren-Liao:
I stripped all business logic on loading the repository lists.  Only two view are added. And the unconnected view is small.  
I hope this is not "big" PR.   Is it?

GoogleCloudExtension/GoogleCloudExtension.TeamExplorerExtension/ISectionViewModel.cs
```diff
@@ -12,12 +12,14 @@
...
 
+using System;
+
 namespace GoogleCloudExtension.TeamExplorerExtension
 {
     /// <summary>
     /// Define interfance for Team Explorer section view model.
     /// </summary>
-    public interface ISectionViewModel
+    public interface ISectionViewModel : IDisposable
```
ivannaranjo: Are all section view models required to be disposable then? It seems to me that this should be in a case by case basis no?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrFontWeightConvert.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Windows;
+using System.Windows.Data;
+using System.Windows.Markup;
+
+namespace GoogleCloudExtension.CloudSourceRepositories
+{
+    /// <summary>
+    /// Converter that sets font to Bold if binding value is true.
+    /// </summary>
+    [ValueConversion(typeof(bool), typeof(System.Windows.FontWeight))]
```
ivannaranjo: We already have a boolean converter that can return two values depending on a boolean value. I would also argue that you should probably do this with styles instead of directly with font weights. You should be able to bind the style property as well.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
```diff
@@ -0,0 +1,170 @@
...
+                            StretchDirection="DownOnly"
+                            Width="16"
+                            Height="16"
+                            Source="{mp:ImageResource CloudSourceRepositories/Resources/cloud-source-repositories.png}" />
+                        
+                        <TextBlock       
+                            Grid.Column="1" 
+                            Text="{Binding Name}" 
+                            Margin="0,0,6,0"
+                            FontWeight="{Binding IsActiveRepo, Converter={local:CsrFontWeightConverter}}"
```
ivannaranjo: This is where you could be using the `BooleanConverter` instead.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
```diff
@@ -0,0 +1,170 @@
...
+                            Style="{StaticResource CommonTextStyle}" />
```
ivannaranjo: And this is where I would recommend having two different styles, instead of just changing the `FontWeight`.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
```diff
@@ -0,0 +1,170 @@
...
+                                        <ColumnDefinition Width="Auto" />
+                                        <ColumnDefinition Width="*" />
+                                    </Grid.ColumnDefinitions>
+                                    <Grid.RowDefinitions>
+                                        <RowDefinition Height="Auto" />
+                                        <RowDefinition Height="Auto" />
+                                    </Grid.RowDefinitions>
+
+                                    <TextBlock 
+                                        Grid.Row="0"
```
ivannaranjo: No need to specify the "0", is the default value.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -38,26 +53,113 @@ public ContentControl Content
...
+            }
+            if (CredentialsStore.Default.CurrentAccount == null)
+            {
+                ManageAccountsWindow.PromptUser();
+            }
+            if (CredentialsStore.Default.CurrentAccount != null)
+            {
+                Content = _reposContent;
+                s_isConnected = true;
+                (this as ISectionViewModel).Refresh();
```
ivannaranjo: Instead of this you could have a private method that both the interface method and this line call.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -38,26 +53,113 @@ public ContentControl Content
...
+            }
         }
 
         void ISectionViewModel.UpdateActiveRepo(string newRepoLocalPath)
         {
             Debug.WriteLine($"CsrSectionControlViewModel.UpdateActiveRepo {newRepoLocalPath}");
+            _reposViewModel.SetActiveRepo(newRepoLocalPath);
+        }
+
+        void IDisposable.Dispose()
```
ivannaranjo: This is really not what dispose is for actually. You should only use dispose if you have non-managed, non .NET, resources. Not to remove yourself from event handlers and such. This is really an anti-pattern.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -38,26 +53,113 @@ public ContentControl Content
...
 
         #endregion
+
+        private void OnAccountChanged()
+        {
+            if (s_isConnected && s_currentAccount == CredentialsStore.Default.CurrentAccount?.AccountName)
+            {
+                return;
+            }
+            (this as ISectionViewModel).Refresh();
```
ivannaranjo: Ditto, separate refresh into a common private method.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUnconnectedViewModel.cs
```diff
@@ -0,0 +1,42 @@
...
+    public class CsrUnconnectedViewModel : ViewModelBase
+    {
+        private CsrSectionControlViewModel _owner;
+
+        /// <summary>
+        /// Respond to Connect link button
+        /// </summary>
+        public ProtectedCommand ConnectCommand { get; }
+
+        public CsrUnconnectedViewModel(CsrSectionControlViewModel ownerControl)
```
ivannaranjo: A view model with a view model owner? That means that one of them is not named correctly or does not have the right role.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/RepoItem.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// Gets if the repository is currently active one.
+        /// </summary>
+        public bool IsActiveRepo
+        {
+            get { return _isActiveRepo; }
+            set { SetValueAndRaise(ref _isActiveRepo, value); }
+        }
+
+        public RepoItem()
```
ivannaranjo: No need for empty constructor.

GoogleCloudExtension/GoogleCloudExtension/packages.config
```diff
@@ -37,5 +37,6 @@
   <package id="Microsoft.VisualStudio.Validation" version="14.1.111" targetFramework="net452" />
   <package id="Newtonsoft.Json" version="10.0.2" targetFramework="net452" />
   <package id="System.Net.Http" version="4.3.1" targetFramework="net452" />
+  <package id="System.Windows.Interactivity.WPF" version="2.0.20525" targetFramework="net452" />
```
ivannaranjo: This is not an official Microsoft package, which is why I didn't use it before. We need to check the license for this before we can use it.

Deren-Liao:
I thought it was produced by Microsoft though 😠

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUnconnectedViewModel.cs
ILMTitan: It looks to me like this should be named "parent" rather than "owner".

GoogleCloudExtension/GoogleCloudExtension.TeamExplorerExtension/ISectionViewModel.cs
Deren-Liao: It makes sense to change the name to something like Cleanup etc. 
As Jim mentioned, if I don't unregister the handle from the event, the object won't be GCed. 

Currently, we only have one implementation to the interface.  Still it is good to add IDisposable or Cleanup etc here.   

Here is the flow.   Team Explorer decides when to "Dispose" the section.  I add this IDisposable so that when Team Explorer calls "Dispose", I add ISectionViewModel.Dispose() into it.   This gives a chance for the view model to do cleanup.   If in the future, other implementation don't need any cleanup. They can simply add empty implementation.  But if they do, this is the place to add any code for clean up.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrFontWeightConvert.cs
Deren-Liao: I did not know that.  Changed.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
Deren-Liao: To me, it is a lot easier to use a converter.   Do we need to use trigger or something to set the style?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
Deren-Liao: I'd change back to implicit for Refresh, to make the code simpler.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
Deren-Liao: I am changing the name to Cleanup().
It is called by the ITeamExplorerSection Dispose method, that's where we get the chance to unregister the handle.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUnconnectedViewModel.cs
Deren-Liao: Changed to parent.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/RepoItem.cs
Deren-Liao: Removed for this PR.  I'll add back with code

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrFontWeightConvert.cs
```diff
@@ -0,0 +1,47 @@
...
+                return (bool)value ? FontWeights.Bold : FontWeights.Normal;
+            }
+            return DependencyProperty.UnsetValue;
+        }
+
+        object IValueConverter.ConvertBack(
+            object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            FontWeight fontWt = (FontWeight)value;
+            return fontWt.ToString();
```
Deren-Liao: Removed

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
```diff
@@ -0,0 +1,170 @@
...
+                </i:EventTrigger>
+            </i:Interaction.Triggers>
+            <ListView.ItemTemplate>
+                <DataTemplate>
+                    <Grid>
+                        <Grid.ContextMenu>
+                            <ContextMenu>
+                                <MenuItem 
+                                    Header="{x:Static ext:Resources.CsrVisitUrlMenuHeader}" 
+                                    Command="{Binding VisitUrlCommand}" />
```
Deren-Liao: I'll add next PR

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
```diff
@@ -0,0 +1,170 @@
...
+                                        Grid.Row="0" 
+                                        Grid.Column="0"
+                                        Margin="0,0,6,0"
+                                        Text="{x:Static ext:Resources.CsrRepoNameLabel}"/>
+
+                                    <TextBlock 
+                                        Grid.Row="0"
+                                        Grid.Column="1"
+                                        HorizontalAlignment="Left"
+                                        Text="{Binding RepoFullName}" />
```
Deren-Liao: Added.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// <summary>
+        /// Sometimes, the view and view model is recreated by Team Explorer.
+        /// This is to preserve the list when a new user control is created.
+        /// Without doing so, user constantly sees the list of repos are loading without reasons.
+        /// </summary>
+        private static ObservableCollection<RepoItem> s_repoList;
+
+        private readonly ITeamExplorerUtils _teamExplorer;
+        private RepoItem _activeRepo;
+        private CsrSectionControlViewModel _owner;
```
Deren-Liao: Removed

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        /// <summary>
+        /// Responds to Clone command
+        /// </summary>
+        public ProtectedCommand CloneCommand { get; }
+
+        /// <summary>
+        /// Responds to Create command.
+        /// </summary>
+        public ProtectedCommand CreateCommand { get; }
```
Deren-Liao: I will add in next PR

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -27,7 +30,19 @@ namespace GoogleCloudExtension.CloudSourceRepositories
...
+        private static bool s_isConnected = false;
+        private static string s_currentAccount;
+
+        private readonly CsrReposContent _reposContent = new CsrReposContent();
+        private readonly CsrUnconnectedContent _unconnectedContent = new CsrUnconnectedContent();
+        private CsrReposViewModel _reposViewModel;
+        private CsrUnconnectedViewModel _unconnectedViewModel;
         private ContentControl _content;
+        private bool _inited;
+        private EventHandler _hander;
```
Deren-Liao: Updated.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -38,26 +53,113 @@ public ContentControl Content
...
+            teamExplorerService.ThrowIfNull(nameof(teamExplorerService));
+            _reposViewModel = new CsrReposViewModel(this, teamExplorerService);
+            _unconnectedViewModel = new CsrUnconnectedViewModel(this);
+            _reposContent.DataContext = _reposViewModel;
+            _unconnectedContent.DataContext = _unconnectedViewModel;
+
+            _hander = (sender, e) => OnAccountChanged();
+            CredentialsStore.Default.CurrentAccountChanged += _hander;
+            CredentialsStore.Default.Reset += _hander;
+            _inited = true;
```
Deren-Liao: Removed

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionStyles.xaml
```diff
@@ -0,0 +1,136 @@
+﻿<ResourceDictionary 
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:PresentationOptions="http://schemas.microsoft.com/winfx/2006/xaml/presentation/options">
```
Deren-Liao: Changed.

GoogleCloudExtension/GoogleCloudExtension/packages.config
Deren-Liao: Removed, changed to InputBindings. 
I thought it is Microsoft package.

GoogleCloudExtension/GoogleCloudExtension.TeamExplorerExtension/ISectionViewModel.cs
ivannaranjo: The way to do this is to use the weak event handler pattern, see [here](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/weak-event-patterns). I started using that in the VS extension but moved away because it was overkill. It seems that you now need it, we should add a WeakEvenHandler class back to the utils class so you can use it.

With this pattern clean up is automatic, by the GC, and no need to write cleanup code is necessary, as it should be in .NET.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposContent.xaml
ivannaranjo: I was saying use a `{Binding ...}` expression on the `Style` property, you can use the `BooleanConverter` to choose what style to use depending on the source property. Binding the style would give you more chances of customizing the item than going property by property.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+        public RepoItem SelectedRepository
+        {
+            get { return _selectedRepo; }
+            set { SetValueAndRaise(ref _selectedRepo, value); }
+        }
+
+        /// <summary>
+        /// Responds to Clone command
+        /// </summary>
+        public ProtectedCommand CloneCommand { get; }
```
ivannaranjo: Just a small nit, typically I only use the full type, `ProtectedCommand`, instead of the interface, `ICommand`, if I intend to enable/disable the command later on dynamically. Since you are not doing that I would recommend declaring the command as `ICommand`.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// </summary>
+        /// <param name="localPath">The repository local path</param>
+        public void SetActiveRepo(string localPath)
+        {
+            var repoItem = Repositories?.FirstOrDefault(
+                x => String.Compare(x.LocalPath, localPath, StringComparison.OrdinalIgnoreCase) == 0);
+            SetActiveRepo(repoItem);
+        }
+
+        private void SetActiveRepo(RepoItem newActiveRepo)
```
ivannaranjo: Wondering why this is not the setter for the ActiveRepo property?

ivannaranjo: Which also brings the question of why you don't have an `ActiveRepo` property that encapsulates all of this logic?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+            IsReady = false;
+            Loading = true;
+            Repositories = new ObservableCollection<RepoItem>();
+            try
+            {
+                // Replace it with code that list repositories in next PR.
+                await Task.Delay(10);   
+                SetActiveRepo(_teamExplorer.GetActiveRepository());
+            }
+            finally
```
ivannaranjo: What happens on error?

ivannaranjo: We don't show error UI or anything?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -38,26 +51,116 @@ public ContentControl Content
             private set { SetValueAndRaise(ref _content, value); }
         }
 
+        [ImportingConstructor]
         public CsrSectionControlViewModel()
-        { }
+        {
+            Debug.WriteLine("new CsrSectionControlViewModel");
```
ivannaranjo: Debugging only code?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -38,26 +51,116 @@ public ContentControl Content
...
+        /// <summary>
+        /// Switch to connected view
+        /// </summary>
+        public void Connect()
+        {
+            if (s_isConnected)
+            {
+                return;
+            }
+            if (CredentialsStore.Default.CurrentAccount == null)
```
ivannaranjo: Thinking about this a bit more this flow seems wrong. If there's no user selected you should direct the user to choose one, or login into Google, using the dialog. What you have right now doesn't give the user any instructions on what to do. The user presses connect and is given the `ManagedAccountsWindow`, which doesn't have any instructions on what they are supposed to do.

You should have an _empty state_ UI to tell the user what to do. We do this on the Cloud Explorer as well, if no user is selected the user is presented with text that instructs them on what to do.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/RepoItem.cs
```diff
@@ -0,0 +1,55 @@
...
+// limitations under the License.
+
+using GoogleCloudExtension.Utils;
+
+namespace GoogleCloudExtension.CloudSourceRepositories
+{
+    /// <summary>
+    /// A repository object binding to list view item
+    /// </summary>
+    public class RepoItem : Model
```
ivannaranjo: This class seems incomplete, how are the properties populated?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/RepoItem.cs
```diff
@@ -0,0 +1,55 @@
...
+        public bool IsActiveRepo
+        {
+            get { return _isActiveRepo; }
+            set { SetValueAndRaise(ref _isActiveRepo, value); }
+        }
+
+        /// <summary>
+        /// The command that opens repository url.
+        /// </summary>
+        public ProtectedCommand VisitUrlCommand { get; }
```
ivannaranjo: Models are supposed to be pure, no commands. Commands only live on the view models.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/RepoItem.cs
Deren-Liao: Yes, I'll add in next PR.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/RepoItem.cs
Deren-Liao: I renamed it to RepoItemViewModel

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
Deren-Liao: I'll think about it in follow up PR. 
I'll change button text to "sign in" if there is no account signed in.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
Deren-Liao: Yes, this is for debugging the MEF creation sequence etc.  I'd keep it because the MEF composition is really tricky. Having a debug message is helpful in future diagnostics.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
Deren-Liao: We do, the method is called by 
ErrorHandlerUtils.HandleAsyncExceptions(ListRepositoryAsync)

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
Deren-Liao: Added ActiveRepo property

Deren-Liao:
I am merging the code.   The only thing left is to add a new UI state where user is not signed into Google Cloud .   I am considering to change the connect... button to sign in ...  or something else.   It will be in follow up PR.   
Ivan, if you have other concerns, let me know.

PR #691, opened 2017-06-13
M2 gcs browser analytics

ivannaranjo:
This PR instruments the GCS browser code to send analytics events.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/GcsFileBrowserDirectoryLoadedEvent.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/GcsFileBrowserDirectoryLoadedEvent.cs
@@ -1,4 +1,18 @@
-﻿using System;
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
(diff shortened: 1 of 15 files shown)


GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/GcsFileBrowserDirectoryLoadedEvent.cs
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace GoogleCloudExtension.Analytics.Events
+{
+    internal static class GcsFileBrowserDirectoryLoadedEvent
+    {
+        private const string GcsFileBrowserDirectoryLoadedEventName = "gcsFileBrowserDirectoryLoaded";
```
ILMTitan: Do we have a way to know that none of these constants are a dupe?

ivannaranjo: Not yet, but that is a good point, we should have validation for these things. Added #696 to target this in M3.

PR #695, opened 2017-06-14
Merge debugger feature to master branch

Deren-Liao:


ivannaranjo:
Is there anything in particular that we should be looking at? Is this just a merge, were there any particular areas of hard conflict that we should be looking at again?

If it is a clean (ish) merge then there's not much for us to look at.

Deren-Liao:
Just merge.  I looked , there is no conflict.

Issue #708, opened 2017-06-16
Template Wizard: Unhelpful error message when no user selected.


ILMTitan:
When there is no user active google account, the select project id dialog gives the unhelpful error message 
"Error loading Projects" with More Information of "Object reference not set to an instance of an object."

PR #711, opened 2017-06-16
Fix issue #708. 

ILMTitan:
Add a check for existing account before trying to get the datasource.


GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
```diff
@@ -22,36 +21,42 @@
...
+        /// <summary>
+        /// If true, null or whitespace values will return <see cref="Visibility.Visible"/>
+        /// instead of <see cref="Visibility.Collapsed"/>.
+        /// </summary>
+        public bool Invert { private get; set; }
+
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            if (value == null)
+
```
Deren-Liao: Extra line

GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
```diff
@@ -22,36 +21,42 @@
...
+    /// or if it is string type and is null or empty or whitespace,
     /// set the visibility to <seealso cref="Visibility.Collapsed"/>.
     /// Otherwise, set the visibility as <seealso cref="Visibility.Visible"/>.
     /// Note: Only Convert is implemented, so this is not a bidirectional converter, do not use on TwoWay bindings.
     /// </summary>
     public class NullEmptyInvisibleConverter : MarkupExtension, IValueConverter
     {
-        #region Implements interface IValueConverter.
+        /// <summary>
+        /// If true, null or whitespace values will return <see cref="Visibility.Visible"/>
```
Deren-Liao: This reminds me, if the value to be converted is value type, it becomes invisible. 
I would add comment to discourage the usage of value type. But if it is value type, the result is invisible.
And add a few tests with value type input?

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2615,4 +2615,16 @@
     <value>Or create a new account</value>
     <comment>The caption to create a new account or mange accounts link</comment>
   </data>
-</root>
+  <data name="WizardPickProjectIdNoAccountMessageAction" xml:space="preserve">
+    <value>select an account</value>
```
Deren-Liao: Capitalize first character?   Select an account

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
```diff
@@ -65,21 +65,33 @@
                         <Setter Property="Padding" Value="2"/>
                     </Style>
                 </Grid.Resources>
-                <Border Visibility="{Binding LoadTask.IsPending, Converter={utils:VisibilityConverter}}">
+                <Border Visibility="{Binding LoadTask, Converter={utils:NullEmptyInvisibleConverter Invert=True}}">
```
Deren-Liao: I would hope WPF supports something like
   switch(...) 
     case ... 

Can you consider to use technics like this one is doing ? 
https://stackoverflow.com/questions/10190887/different-views-data-template-based-on-member-variable

GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
```diff
@@ -22,36 +21,42 @@
 namespace GoogleCloudExtension.Utils
 {
     /// <summary>
-    /// If the object is null, 
-    /// or if it is string type and is null or empty or whitespace, 
+    /// If the object is null,
+    /// or if it is string type and is null or empty or whitespace,
```
ivannaranjo: You will never get a null string, so the comment should read: _If the object is a string and it is empty or whitespace or if the value is null then it will return Visibility.Collapsed_.

GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
```diff
@@ -22,36 +21,42 @@
...
     /// Note: Only Convert is implemented, so this is not a bidirectional converter, do not use on TwoWay bindings.
     /// </summary>
     public class NullEmptyInvisibleConverter : MarkupExtension, IValueConverter
     {
-        #region Implements interface IValueConverter.
+        /// <summary>
+        /// If true, null or whitespace values will return <see cref="Visibility.Visible"/>
+        /// instead of <see cref="Visibility.Collapsed"/>.
+        /// </summary>
+        public bool Invert { private get; set; }
```
ivannaranjo: Making the `get` private is really kind of weird, since you are already making the set public. I would just make both public and live with it.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
ivannaranjo: Shouldn't it be "Select and account" ?

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2615,4 +2615,16 @@
...
+    <comment>The clickable part of the message shown when there is no current user on the template wizard pick project window.</comment>
+  </data>
+  <data name="WizardPickProjectIdNoAccountMessageLeft" xml:space="preserve">
+    <value>To get a list of projects,</value>
+    <comment>The left part of the message shown when there is no current user on the template wizard pick project window.</comment>
+  </data>
+  <data name="WizardPickProjectIdNoAccountMessageRight" xml:space="preserve">
+    <value>.</value>
+    <comment>The right part of the message shown when there is no current user on the template wizard pick project window.</comment>
```
ivannaranjo: I see what you are doing here, you should probably not do this in parts. This will make the code impossible to localize in the future. We don't really have a good way to have links in between the text and making it clickable. I would just change the text and add a button or a link after it, not in the middle of the text.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
ivannaranjo: Using a template selector is definitely something that we should consider at some point, please open a bug for this.

GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
ILMTitan: The way it was written, value types would unset the visibility property. I rewrote it (to enable non-null object types), and now value types will always be visible (unless inverted).

I also don't think adding tests for value types would be useful, as we don't know how they should be handled.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindowContent.xaml
ILMTitan: Opened bug #713.

GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.Utils/NullEmptyInvisibleConverter.cs
ILMTitan: Changed it back and added a comment to disable the R# suggestion.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
ILMTitan: No. The whole message reads:
> To get a list of projects, select an account.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
ILMTitan: Done.

PR #718, opened 2017-06-20
Add clone window

Deren-Liao:
Clone repository
devenv_2017-06-19_20-12-42


GoogleCloudExtension/GoogleCloudExtension.Utils/PathUtils.cs
```diff
@@ -49,5 +50,8 @@ public static string GetCommandPathFromPATH(string commandName)
 
             return null;
         }
+
+        public static bool IsPathEmpth(string filePath) =>
```
ivannaranjo: Add comment, also a better name would be `IsDirectoryEmpty()`.

GoogleCloudExtension/GoogleCloudExtension.Utils/WindowsCredentialManager.cs
```diff
@@ -73,13 +78,25 @@ public static bool Write(string targetName, string username, string password)
...
 
-        private enum CredentialPersistence : uint
+        public enum CredentialPersistence : uint
         {
             Session = 1,
             LocalMachine,
             Enterprise
         }
 
+        public enum CredentialType
```
ivannaranjo: Comments on this would be good. This values are coming from somewhere, we should document where they come from.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="../Theming/CommonResources.xaml" />
+                <ResourceDictionary Source="../Controls/DotsProgressIndicatorResources.xaml" />
+            </ResourceDictionary.MergedDictionaries>
+
+        </ResourceDictionary>
+    </UserControl.Resources>
+
+    <UserControl.Style>
+        <Binding Source="{StaticResource CommonDialogStyleBase}" />
```
ivannaranjo: Should use one of the fixed styles to get the same size as the rest of the dialogs. Or if a dynamic size is required use the dynamic style.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+
+    <theming:CommonDialogWindowBaseContent>
+        <theming:CommonDialogWindowBaseContent.Buttons>
+            <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiOkButtonCaption}"
+                                      Command="{Binding OkCommand}"
+                                      IsDefault="True" />
+            <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiCancelButtonCaption}" IsCancel="True" />
+        </theming:CommonDialogWindowBaseContent.Buttons>
+
+
```
ivannaranjo: Extra line.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+        <StackPanel IsEnabled="{Binding IsReady}">
+
```
ivannaranjo: Extra lines.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+            <Label 
+                Content="{x:Static ext:Resources.CsrProjectSelectionLabel}"
+                Target="{Binding ElementName=ProjectsComboBox}"
```
ivannaranjo: Prefer to use the same convention as members in a class for element names, in this case `_projectsComboBox`. These names will appear as members in the code behind.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+                Style="{StaticResource CommonLabelStyle}" />
+
+            <ComboBox
```
ivannaranjo: User common style for ComboBox.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+                SelectedItem="{Binding SelectedRepository, Mode=TwoWay}"
+                IsSynchronizedWithCurrentItem="True"
+                DisplayMemberPath="Name" 
+                IsTextSearchEnabled="True" />
+
+            <Label 
+                Content="{x:Static ext:Resources.CsrCloneLocalPathLabel}"
+                Target="{Binding ElementName=LocalPathTextBox}"
+                Style="{StaticResource CommonLabelStyle}" 
+                Margin="0,6,0,0" />
```
ivannaranjo: Other dialogs use 5 as the margin, why the difference?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+                    Command="{Binding PickFolderCommand}"
+                    Style="{StaticResource CommonButtonDynamicStyle}" Height="26" Margin="0,2"/>
+
+            </Grid>
+
+            <ctrl:DotsProgressIndicator 
+                Margin="0,12,0,0" 
+                Width="60" 
+                HorizontalAlignment="Left" 
+                Visibility="{Binding IsReady, Converter={utils:BooleanConverter TrueValue='Hidden',FalseValue='Visible'}}"/>
```
ivannaranjo: Use the `VisiblityConverter`, you can negate it if necessary.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+                Margin="0,6,0,0" />
+
+            <Grid>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="17*"/>
+                    <ColumnDefinition Width="48*"/>
+                    <ColumnDefinition Width="Auto"/>
+                </Grid.ColumnDefinitions>
+
+                <TextBox
```
ivannaranjo: You only have two elements but 3 columns, you could just define two columns, the first one with "*" width and the second one "auto" and you would be set.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+                    <ColumnDefinition Width="Auto"/>
+                </Grid.ColumnDefinitions>
+
+                <TextBox 
+                    x:Name="LocalPathTextBox"
+                    Text="{Binding LocalPath, UpdateSourceTrigger=PropertyChanged, ValidatesOnNotifyDataErrors=True, Mode=TwoWay}"
+                    Style="{StaticResource CommonTextBoxStyle}"
+                    Margin="0,0,6,0" Grid.ColumnSpan="2"/>
+
+                <Button
```
ivannaranjo: I think you need a tooltip for this button, since the caption is only ...

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+
+namespace GoogleCloudExtension.CloudSourceRepositories
+{
+    /// <summary>
+    /// View model for user control CsrCloneWindowContent.xaml.
+    /// </summary>
+    public class CsrCloneWindowViewModel : ValidatingViewModelBase
+    {
+        private CsrCloneWindow _owner;
+
```
ivannaranjo: Extra line.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+            _owner = owner.ThrowIfNull(nameof(owner));
+            PickFolderCommand = new ProtectedCommand(PickFoloder);
+            OkCommand = new ProtectedCommand(taskHandler: () => ExecuteAsync(CloneAsync), canExecuteCommand: false);
+            ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(InitializeAsync));
+        }
+
+        private async Task CloneAsync()
+        {
+            // If OkCommand is enabled, SelectedRepository and LocalPath is valid
+            string destPath = Path.Combine(LocalPath, SelectedRepository.GetRepoName());
```
ivannaranjo: This looks like the user only chooses the parent directory where the repo is going to be cloned, not the directory for the repo itself. This is potentially confusing since it is not what git does. You should make sure that this is in the tooltip and/or help icon.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+            {
+                IsReady = true;
+            }
+        }
+
+        private async Task ListRepoAsync()
+        {
+            Debug.WriteLine("ListRepoAsync");
+
+            Repositories = null;
```
ivannaranjo: What is the empty state for the UI? You should plan on showing something to the user so they know that there are not repos in their project, maybe even have a "<New repo>" item that the user can use to create a new repo or navigate to the cloud console.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+
+            Projects = await resourceManager.GetSortedActiveProjectsAsync();
+            if (Projects.Any())
+            {
+                SelectedProject = Projects.FirstOrDefault();
+                await ListRepoAsync();
+            }
+            else
+            {
+                UserPromptUtils.ErrorPrompt(
```
ivannaranjo: Not having a project is not an error, the UI should show the relevant information or even better, the dialog should not have been opened in the first place.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+        {
+            SetValidationResults(
+                ValidateLocalPath(LocalPath, Resources.CsrCloneLocalPathFieldName), 
+                nameof(LocalPath));
+            OkCommand.CanExecuteCommand = SelectedRepository != null && !HasErrors;
+        }
+
+        private IEnumerable<ValidationResult> ValidateLocalPath(string path, string fieldName)
+        {
+            if (String.IsNullOrWhiteSpace(path))
```
ivannaranjo: Technically speaking a path named "   " is valid, you should check for `IsNullOrEmpty()`. You should also check for invalid chars in directory names, such as "?*" etc...

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -281,5 +282,14 @@ private void SetCurrentRepo(string localPath)
             }
             return localRepos;
         }
+
+        private void Clone()
+        {
+            var repoItem = CsrCloneWindow.PromptUser();
+            if (repoItem != null)
+            {
+                (Repositories ?? new ObservableCollection<RepoItemViewModel>()).Add(repoItem);
```
ivannaranjo: Hmm... Not sure why you are doing this, if `Repositories` is `null` then you are just creating a temporary `ObservableCollection<T>` adding the new `repoItem` and throwing it away. I think that you are meant to assign the new collection to `Repositories`.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -70,8 +73,12 @@ public void Disconnect()
         /// <summary>
         /// Switch to connected view
         /// </summary>
-        public void Connect()
+        public async Task Connect()
         {
+            if (!(await InitializeGit()))
```
ivannaranjo: General advice, is ok to have temporary variables to make the code more readable.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -156,11 +166,51 @@ void ISectionViewModel.Cleanup()
...
+                }
+            }
             if (s_isConnected && s_currentAccount == CredentialsStore.Default.CurrentAccount?.AccountName)
             {
                 return;
             }
             Refresh();
         }
+
+        private async Task<bool> InitializeGit()
```
ivannaranjo: Since you are initializing a global variable, I would recommend using `Lazy<Task<bool>>` and get the protection that `Lazy<T>` gives you.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -156,11 +166,51 @@ void ISectionViewModel.Cleanup()
...
+            s_gitInited = SetGitCredential();
+            return s_gitInited;
+        }
+
+        private bool SetGitCredential()
+        {
+            if (CredentialsStore.Default.CurrentAccount != null)
+            {
+                if (!CsrGitUtils.StoreCredential(
+                    "https://source.developers.google.com",
```
ivannaranjo: Hardcoded string, consider having in a central location.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUnconnectedViewModel.cs
```diff
@@ -26,17 +27,12 @@ public class CsrUnconnectedViewModel : ViewModelBase
...
         public CsrUnconnectedViewModel(CsrSectionControlViewModel parent)
         {
             _parent = parent;
-            ConnectCommand = new ProtectedCommand(Connect);
-        }
-
-        private void Connect()
-        {
-            _parent.Connect();
+            ConnectCommand = new ProtectedCommand(taskHandler: parent.Connect);
```
ivannaranjo: No need to specify `taskHandler:` since it is the first parameter.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUtils.cs
```diff
@@ -29,6 +30,15 @@ namespace GoogleCloudExtension.CloudSourceRepositories
     public static class CsrUtils
     {
         /// <summary>
+        /// Gets repository name 
+        /// The Cloud Repository name has format $"{ProjectResourceName}/repos/{repoName}" };
+        /// Returns the last part which is the repo name.
+        /// </summary>
+        /// <param name="cloudRepo">Repository object</param>
+        /// <returns>Repository name</returns>
+        public static string GetRepoName(this Repo cloudRepo) => cloudRepo?.Name?.Split('/').LastOrDefault();
```
ivannaranjo: Why is `cloudRepo` allowed to be `null`?

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2651,5 +2651,48 @@
...
+  <data name="CsrRepositorySelectionLabel" xml:space="preserve">
+    <value>_Repository:</value>
+    <comment>Repository selection label</comment>
+  </data>
+  <data name="CsrCloneLocalPathFieldName" xml:space="preserve">
+    <value>Local path</value>
+    <comment>The field name of the input local path text box</comment>
+  </data>
+  <data name="CsrClonePathExistNotEmptyMessageFormat" xml:space="preserve">
+    <value>The destination path {0} is not empty</value>
```
ivannaranjo: Missing period at end.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2651,5 +2651,48 @@
...
+    <comment>Error message when destination path is not empty. {0} is the destination path where repository is copied to.</comment>
+  </data>
+  <data name="CsrClonePathNotExistMessage" xml:space="preserve">
+    <value>The path does not exist</value>
```
ivannaranjo: Missing period at end.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2651,5 +2651,48 @@
...
+  <data name="CsrCloneLocalPathLabel" xml:space="preserve">
+    <value>_Clone to:</value>
+    <comment>Label to the input local path text box</comment>
+  </data>
+  <data name="CsrCloneNoProject" xml:space="preserve">
+    <value>There is no project</value>
+    <comment>Error message if the user account has no any project</comment>
+  </data>
+  <data name="CsrClonePathBrowseButtonCaption" xml:space="preserve">
+    <value>...</value>
```
ivannaranjo: Why not use "Browse", that way you can also add an accelerator to the button.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2651,5 +2651,48 @@
...
+  <data name="CsrCloneFailedToSetCredentialMessage" xml:space="preserve">
+    <value>Failed to set credential</value>
+    <comment>Error message when it fails to set git credential</comment>
+  </data>
+  <data name="CsrCloneLocalPathLabel" xml:space="preserve">
+    <value>_Clone to:</value>
+    <comment>Label to the input local path text box</comment>
+  </data>
+  <data name="CsrCloneNoProject" xml:space="preserve">
+    <value>There is no project</value>
```
ivannaranjo: Missing dot.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2651,5 +2651,48 @@
...
+  <data name="CsrCloneWindowTitle" xml:space="preserve">
+    <value>Clone Google Cloud Source Repository</value>
+    <comment>The title of the clone CSR window</comment>
+  </data>
+  <data name="CsrCloneFailedMessage" xml:space="preserve">
+    <value>Failed to clone the repository</value>
+    <comment>Error message when it fails to clone for some reason.</comment>
+  </data>
+  <data name="CsrCloneFailedToSetCredentialMessage" xml:space="preserve">
+    <value>Failed to set credential</value>
```
ivannaranjo: Missing dot at the end.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2651,5 +2651,48 @@
     <value>View the repository on web</value>
     <comment>The menu header for visiting souce repository on web page</comment>
   </data>
-
+  <data name="CsrCloneWindowTitle" xml:space="preserve">
+    <value>Clone Google Cloud Source Repository</value>
+    <comment>The title of the clone CSR window</comment>
+  </data>
+  <data name="CsrCloneFailedMessage" xml:space="preserve">
+    <value>Failed to clone the repository</value>
```
ivannaranjo: Missing dot at the end.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: I need it to be Hidden instead of collapsed.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Did not notice this.  The designer modified my controls.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: I'll check if the path exists.  ? * won't pass the check.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUnconnectedViewModel.cs
ILMTitan: It should be the first argument, but @Deren-Liao recently added a second constructor for `ProtectedCommand` that takes a `Func<Task>`. That taskHandler argument is the second arg because it was causing resolution errors with the constructor that takes an `Action`.
I would suggest you change the async protected command constructor to a factory method instead, and use a new internal constructor that takes an `IProtectedAction`.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
Deren-Liao: This may not apply to this code.  Later, I'll add a retry page.  If it fails, user has chance to retry.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUnconnectedViewModel.cs
Deren-Liao: I agree, this is odd.  I'll make a change later.  #723

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrUtils.cs
Deren-Liao: Changed

GoogleCloudExtension/GoogleCloudExtension.Utils/WindowsCredentialManager.cs
```diff
@@ -32,7 +32,12 @@ public static class WindowsCredentialManager
         /// </param>
         /// <param name="username">The credential username.</param>
         /// <param name="password">The credential password.</param>
-        public static bool Write(string targetName, string username, string password)
+        /// <param name="credentialType">Credentrial type</param>
+        /// <param name="persistenceType">Credential persistence type</param>
+        public static bool Write(
+            string targetName, string username, string password,
+            CredentialType credentialType,
```
Deren-Liao: Added exception.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+    <theming:CommonDialogWindowBaseContent>
+        <theming:CommonDialogWindowBaseContent.Buttons>
+            <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiOkButtonCaption}"
+                                      Command="{Binding OkCommand}"
+                                      IsDefault="True" />
+            <theming:DialogButtonInfo Caption="{x:Static ext:Resources.UiCancelButtonCaption}" IsCancel="True" />
+        </theming:CommonDialogWindowBaseContent.Buttons>
+
+
+        <StackPanel IsEnabled="{Binding IsReady}">
```
Deren-Liao: Then the user has a chance to choose different project and trigger another ListRepo call. That makes  the code flow complicated.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+        {
+            Debug.WriteLine("Init");
+
+            ResourceManagerDataSource resourceManager = DataSourceFactories.CreateResourceManagerDataSource();
+            if (resourceManager == null)
+            {
+                return;
+            }
+
+            Projects = await resourceManager.GetSortedActiveProjectsAsync();
```
Deren-Liao: I moved this code out

GoogleCloudExtension/GoogleCloudExtension.Utils/WindowsCredentialManager.cs
Deren-Liao: My intention was to make sure caller knows what exact value they set.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Added tooltip

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+                dialog.ShowNewFolderButton = true;
+                var result = dialog.ShowDialog();
+                if (result == System.Windows.Forms.DialogResult.OK)
+                {
+                    LocalPath = dialog.SelectedPath;
+                }
+            }
+        }
+
+        private async Task ExecuteAsync(Func<Task> task)
```
Deren-Liao: It used for command, initialization etc, not directly for property.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            if (Directory.Exists(localPath))
+            {
+                throw new ArgumentException($"{localPath} arleady exists.");
+            }
+
+            Directory.CreateDirectory(localPath);
+
+            // git clone https://host/myrepo/ c:\git\myrepo --config credential.helper=manager
+            string command = $"clone {url} {localPath} --config credential.helper=manager";
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,89 @@
...
+            }
+            return await GitRepository.GetGitCommandWrapperForPathAsync(localPath);
+        }
+
+        /// <summary>
+        /// Store credential using git-credential-manager
+        /// </summary>
+        /// <param name="url">The repository url.</param>
+        /// <param name="refreshToken">Google cloud credential refresh token.</param>
+        /// <param name="useHttpPath">Set for the path of for host</param>
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,89 @@
...
+            var uriPartial = useHttpPath ? UriPartial.Path : UriPartial.Authority;
+            return WindowsCredentialManager.Write(
+                $"git:{uri.GetLeftPart(uriPartial)}",
+                username: "VisualStudioUser",
+                password: refreshToken,
+                credentialType: WindowsCredentialManager.CredentialType.Generic,
+                persistenceType: WindowsCredentialManager.CredentialPersistence.LocalMachine);
+        }
+
+        public static async Task<bool> SetUseHttpPath() =>
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
```diff
@@ -119,7 +119,7 @@ public Task<List<string>> GetRevisionFileAsync(string sha, string relativePath)
         /// <summary>
         /// Run a git command and return the output or error output.
         /// </summary>
-        private static async Task<List<string>> RunGitCommandAsync(
+        public static async Task<List<string>> RunGitCommandAsync(
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/PathUtils.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/WindowsCredentialManager.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done, using DockPanel

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Added tooltip

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: I added InvisibleValue property

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Added tooltip

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: When empty, I show a message in repos combo box now.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Moved to show before dialog open.  And this actually should never happen. Just to be safe, I adds this error box.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
Deren-Liao: Changed

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done. No accelerator for now.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,89 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GoogleCloudExtension.Utils;
+using System;
+using System.Collections.Generic;
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,89 @@
...
+                throw new ArgumentException($"{localPath} arleady exists.");
+            }
+
+            Directory.CreateDirectory(localPath);
+
+            // git clone https://host/myrepo/ c:\git\myrepo --config credential.helper=manager
+            string command = $"clone {url} {localPath} --config credential.helper=manager";
+            var output = await GitRepository.RunGitCommandAsync(command, localPath);
+            Debug.WriteLine(output?.FirstOrDefault() ?? "");
+            if (output == null)
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,106 @@
...
+
+            <Label 
+                Content="{x:Static ext:Resources.CsrProjectSelectionLabel}"
+                Target="{Binding ElementName=ProjectsComboBox}"
+                Style="{StaticResource CommonLabelStyle}" />
+
+            <ComboBox 
+                x:Name="ProjectsComboBox"
+                ItemsSource="{Binding Projects}"
+                SelectedItem="{Binding SelectedProject, Mode=TwoWay}"
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+using System.Windows.Input;
+
+namespace GoogleCloudExtension.CloudSourceRepositories
+{
+    /// <summary>
+    /// View model for user control CsrCloneWindowContent.xaml.
+    /// </summary>
+    public class CsrCloneWindowViewModel : ValidatingViewModelBase
+    {
+        private CsrCloneWindow _owner;
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+
+        private void ValidateInputs()
+        {
+            SetValidationResults(
+                ValidateLocalPath(LocalPath, Resources.CsrCloneLocalPathFieldName), 
+                nameof(LocalPath));
+            OkCommand.CanExecuteCommand = SelectedRepository != null && !HasErrors;
+        }
+
+        private IEnumerable<ValidationResult> ValidateLocalPath(string path, string fieldName)
```
Deren-Liao: Removed all input parameters

Deren-Liao:
Updated, please take a look.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,121 @@
...
+                IsTextSearchEnabled="True"
+                Style="{StaticResource CommonComboBoxStyle}"/>
+
+            <Label 
+                Content="{x:Static ext:Resources.CsrRepositorySelectionLabel}"
+                Target="{Binding ElementName=_repositoryComboBox}"
+                Style="{StaticResource CommonLabelStyle}"
+                Margin="0,5,0,0"/>
+
+            <Grid>
```
ILMTitan: This can be done with control templates rather than visibility. See the TemplateWizard PickProjectIdWindow.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -0,0 +1,121 @@
...
+
+                <ComboBox 
+                    Visibility="{Binding ProjectHasRepo, Converter={utils:VisibilityConverter}}"
+                    x:Name="_repositoryComboBox"
+                    ItemsSource="{Binding Repositories}"
+                    SelectedItem="{Binding SelectedRepository}"
+                    IsSynchronizedWithCurrentItem="True"
+                    DisplayMemberPath="Name" 
+                    IsTextSearchEnabled="True"
+                    Style="{StaticResource CommonComboBoxStyle}"/>
```
ILMTitan: You should also include a case where `Repositories` are loading.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
ILMTitan: It makes sense to disable the panel during the clone operation, but the `ListRepos` operation could be done as an `AsyncProperty`.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                await task();
+            }
+            finally
+            {
+                IsReady = true;
+            }
+        }
+
+        private async Task ListRepoAsync()
```
ILMTitan: With `Repositories` as an `AsyncProperty`, this method would looks something like this:

```
private async void StartListRepoTask()
{
    Debug.WriteLine(nameof(StartListRepoTask));
    AsyncProperty<IList<Repo>> repositories =
        AsyncPropertyUtils.CreateAsyncProperty(CsrUtils.GetCloudReposAsync(SelectedProject.ProjectId));
    Repositories = repositories;
    await repositories.ValueTask;
    if (Repositories == repositories)
    {
        SelectedRepository = Repositories.Value?.FirstOrDefault();
        ProjectHasRepo = Repositories.Value?.Any() ?? false;
    }
}
```

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -281,5 +282,47 @@ private void SetCurrentRepo(string localPath)
...
+
+        private async Task CloneAsync()
+        {
+            IsReady = false;
+            Loading = true;
+
+            ResourceManagerDataSource resourceManager = DataSourceFactories.CreateResourceManagerDataSource();
+            if (resourceManager == null)
+            {
+                return;
```
ILMTitan: If the resourceManager fails to be created, this view model will get stuck in the Loading state.

GoogleCloudExtension/GoogleCloudExtension/Utils/ProtectedCommand.cs
```diff
@@ -82,6 +79,23 @@ public bool CanExecuteCommand
 
     /// <summary>
     /// This class implements the <seealso cref="ICommand"/> interface and wraps the action for the command
+    /// in <seealso cref="ErrorHandlerUtils.HandleAsyncExceptions(Func{Task})(taskHandler)"/> 
+    /// to handle the exceptions that could escape.
+    /// </summary>
+    public class ProtectedAsyncCommand : ProtectedCommand
```
ILMTitan: Why does this need to be a separate class rather than just a factory method?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                UserPromptUtils.ErrorPrompt(
+                    message: Resources.CsrCloneFailedToSetCredentialMessage,
+                    title: Resources.UiDefaultPromptTitle);
+                return;
+            }
+
+            try
+            {
+                GitRepository localRepo = await CsrGitUtils.Clone(SelectedRepository.Url, destPath);
```
ILMTitan: Git outputs a bunch of progress while cloning. Do we want he user to be able to see that?

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// Clone a Google Cloud Source Repository locally.
+        /// </summary>
+        /// <param name="url">The repository remote URL.</param>
+        /// <param name="localPath">Local path to save the repository</param>
+        /// <returns>
+        /// A <seealso cref="GitRepository"/> object if clone is successful.
+        /// Or null if it fails for some reason.
+        /// </returns>
+        /// <exception cref="CsrGitCommandException">Throw when git command fails</exception>
+        public static async Task<GitRepository> Clone(string url, string localPath)
```
Deren-Liao: For now, only CSR uses it.  I don't see we need this be used elsewhere in foreseeable future.
The clone uses credential helper "manager" which is for CSR feature.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: I'll see if I can add later

GoogleCloudExtension/GoogleCloudExtension/Utils/ProtectedCommand.cs
Deren-Liao: Either way works well.
If making it factory method, it becomes odd that there is still a default constructor.  
By creating a new class, I avoid such inconsistent design.

GoogleCloudExtension/GoogleCloudExtension.Utils/WindowsCredentialManager.cs
```diff
@@ -78,13 +78,20 @@ public static class WindowsCredentialManager
         [DllImport("Advapi32.dll", EntryPoint = "CredWriteW", CharSet = CharSet.Unicode, SetLastError = true)]
         private static extern bool CredWrite([In] ref CREDENTIAL userCredential, [In] UInt32 flags);
 
+        /// <summary>
+        /// For more detail, use MSDN to check CREDENTIAL structure and
+        /// CRED_PERSIST_SESSION, CRED_PERSIST_LOCAL_MACHINE and CRED_PERSIST_ENTERPRISE.
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/WindowsCredentialManager.cs
```diff
@@ -73,13 +78,32 @@ public static bool Write(string targetName, string username, string password)
         [DllImport("Advapi32.dll", EntryPoint = "CredWriteW", CharSet = CharSet.Unicode, SetLastError = true)]
         private static extern bool CredWrite([In] ref CREDENTIAL userCredential, [In] UInt32 flags);
 
-        private enum CredentialPersistence : uint
+        /// <summary>
+        /// For more detail, use MSDN to check CREDENTIAL structure and
+        /// CRED_PERSIST_SESSION, CRED_PERSIST_LOCAL_MACHINE and CRED_PERSIST_ENTERPRISE.
+        /// </summary>
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/WindowsCredentialManager.cs
```diff
@@ -73,13 +78,32 @@ public static bool Write(string targetName, string username, string password)
...
+        public enum CredentialPersistence : uint
         {
             Session = 1,
             LocalMachine,
             Enterprise
         }
 
+        /// <summary>
+        /// For more detail, use MSDN to check CREDENTIAL structure Type field.
+        /// </summary>
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            GitRepository gitRepo = null;
+            Directory.CreateDirectory(localPath);
+
+            // git clone https://host/myrepo/ "c:\git\myrepo" --config credential.helper=manager
+            string command = $@"clone {url} ""{localPath}"" --config credential.helper=manager";
+            var output = await GitRepository.RunGitCommandAsync(command, localPath);
+            Debug.WriteLine(output?.FirstOrDefault() ?? "");
+            if (output == null ||
+                (gitRepo = await GitRepository.GetGitCommandWrapperForPathAsync(localPath)) == null)
```
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done. Slightly different, I use contentControl so that I don't have to define ControlTemplate wraps  ComboBox

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
Deren-Liao: Done

Deren-Liao:
Please take a look again.    Thanks.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,97 @@
...
+            Debug.WriteLine(output.FirstOrDefault() ?? "");
+            return await GitRepository.GetGitCommandWrapperForPathAsync(localPath);
+        }
+
+        /// <summary>
+        /// Store credential using git-credential-manager
+        /// </summary>
+        /// <param name="url">The repository url.</param>
+        /// <param name="refreshToken">Google cloud credential refresh token.</param>
+        /// <param name="useHttpPath">
```
ivannaranjo: Nit, a boolean value is not a good way of modeling this behavior. You are choosing between two completely unrelated things, the URL path or the URL host, and the name of the variable doesn't convey that meaning. I would use an enum, where you can specify the full choice.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// False: store credential for host.
+        /// </param>
+        /// <returns>
+        /// True: if credential is stored successfully.
+        /// Otherwise false.
+        /// </returns>
+        public static bool StoreCredential(string url, string refreshToken, bool useHttpPath = false)
+        {
+            url.ThrowIfNullOrEmpty(nameof(url));
+            refreshToken.ThrowIfNullOrEmpty(nameof(url));
```
ivannaranjo: Wrong variable name, should be `nameof(refreshToken)`.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            Uri uri = new Uri(url);
+            var uriPartial = useHttpPath ? UriPartial.Path : UriPartial.Authority;
+            return WindowsCredentialManager.Write(
+                $"git:{uri.GetLeftPart(uriPartial)}",
+                username: "VisualStudioUser",
```
ivannaranjo: You should promote this string to a constant with a descriptive comment of what it means.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
```diff
@@ -0,0 +1,97 @@
...
+                credentialType: WindowsCredentialManager.CredentialType.Generic,
+                persistenceType: WindowsCredentialManager.CredentialPersistence.LocalMachine);
+        }
+
+        /// <summary>
+        /// Set global git config useHttpPath for CSR host.
+        /// Refer to https://git-scm.com/docs/gitcredentials
+        /// </summary>
+        public static async Task SetUseHttpPathAsync() =>
+            await GitRepository.RunGitCommandAsync(
```
ivannaranjo: You can just return the task that `RunGitCommandAsync()` returns, no need to use `await` or make the method `async`.

GoogleCloudExtension/GoogleCloudExtension.Utils/VisibilityConverter.cs
```diff
@@ -28,6 +28,12 @@ namespace GoogleCloudExtension.Utils
     public class VisibilityConverter : MarkupExtension, IValueConverter
     {
         /// <summary>
+        /// By default, invisible is converted to Collapsed.
+        /// This property allows caller to override the false conversion value to Hidden.
+        /// </summary>
+        public Visibility InvisibleValue { get; set; } = Visibility.Collapsed;
```
ivannaranjo: Do not use Hidden, user Collapsed always. Hidden still uses space in the UI while Collapsed doesn't. You almost never want to use Hidden.

GoogleCloudExtension/GoogleCloudExtension.Utils/VisibilityConverter.cs
```diff
@@ -50,7 +56,7 @@ public object Convert(object value, Type targetType, object parameter, CultureIn
             if (value is bool)
             {
                 bool isVisible = IsNegated ^ (bool)value;
-                var result = isVisible ? Visibility.Visible : Visibility.Collapsed;
+                var result = isVisible ? Visibility.Visible : InvisibleValue;
```
ivannaranjo: Why do you need to use anything other than `Collapsed`?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,263 @@
...
+
+namespace GoogleCloudExtension.CloudSourceRepositories
+{
+    /// <summary>
+    /// View model for user control CsrCloneWindowContent.xaml.
+    /// </summary>
+    public class CsrCloneWindowViewModel : ValidatingViewModelBase
+    {
+        private readonly CsrCloneWindow _owner;
+
```
ivannaranjo: Extra line.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,263 @@
...
+            Debug.WriteLine(nameof(StartListRepoTaskAsync));
+            // Reset SelectedRepository before running the async task.
+            // This disables OkCommand
+            RepositoriesAsync = null;
+            ProjectHasRepo = false;
+            AsyncProperty<IList<Repo>> repos =
+                AsyncPropertyUtils.CreateAsyncProperty(CsrUtils.GetCloudReposAsync(SelectedProject.ProjectId));
+            RepositoriesAsync = repos;
+            await repos.ValueTask;
+            if (RepositoriesAsync == repos)
```
ivannaranjo: Instead of this you should consider having a single class with all of the necessary properties, expose it as single atomic value.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -0,0 +1,263 @@
...
+        private void ValidateInputs()
+        {
+            SetValidationResults(ValidateLocalPath(), nameof(LocalPath));
+            OkCommand.CanExecuteCommand = SelectedRepository != null && !HasErrors;
+        }
+
+        private IEnumerable<ValidationResult> ValidateLocalPath()
+        {
+            string fieldName = Resources.CsrCloneLocalPathFieldName;
+            if (String.IsNullOrEmpty(LocalPath))
```
ivannaranjo: You should validate `LocalPath` as well, ensure that the path is valid, does not contain forbidden chars, etc...

Deren-Liao: We don't need to validate local path.  This is validated automatically by checking the path exists.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.GitUtils/CsrGitUtils.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/VisibilityConverter.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension.Utils/VisibilityConverter.cs
Deren-Liao: Changed back

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Done

Deren-Liao:
Updated.  Added new class AsyncRepositories.  With the new class, the code looks better. 
Hi Jim,  please take a look again.  Thanks.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
```diff
@@ -114,14 +116,28 @@ public class GitRepository
...
+        /// <returns>
+        /// The command execution output.
+        /// If output is empty, returns empty list. 
+        /// If return value is null, it indicates there are some errors.
+        /// </returns>
         internal static async Task<List<string>> RunGitCommandAsync(
             string command,
-            string gitLocalRoot)
+            string gitLocalRoot,
+            bool throwOnError = true)
```
ILMTitan: This doesn't actually throw?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
```diff
@@ -57,27 +57,52 @@
...
+                                <Setter Property="Content" Value="{StaticResource EmptyList}" />
+                            </DataTrigger>
+                            <DataTrigger Binding="{Binding RepositoriesAsync.DisplayState}" Value="HasItems" >
+                                <Setter Property="Content" Value="{StaticResource NormalList}" />
+                            </DataTrigger>
+                        </Style.Triggers>
+                    </Style>
+                </ContentControl.Style>
+               
+            </ContentControl>
```
ILMTitan: This is way better than what I did on the on the PickProjectId wizard dialog.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -72,21 +59,18 @@ public Project SelectedProject
             {
                 var oldValue = _selectedProject;
                 SetValueAndRaise(ref _selectedProject, value);
-                if (_selectedProject != null && _isReady && oldValue != _selectedProject)
+                if (_selectedProject != null && IsReady && oldValue != _selectedProject)
                 {
-                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(ListRepoAsync));
+                    ErrorHandlerUtils.HandleAsyncExceptions(() =>
+                        RepositoriesAsync.StartListRepoTaskAsync(_selectedProject.ProjectId));
```
ILMTitan: I am pretty sure this task will never throw an exception.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace GoogleCloudExtension.CloudSourceRepositories
+{
+    /// <summary>
+    /// Wrapper on top of AsyncProperty.
+    /// Retrieves a list of <seealso cref="Repo"/> objects by calling async task.
+    /// 
+    /// The main goal is to allow user to choose different project while waiting for the list to be populated.
+    /// Anytime, user can change project and then it calls StartListRepoTaskAsync to update the repo list.
+    /// </summary>
+    public class AsyncRepositories : Model
```
ILMTitan: Because you want to reset the selected repository when the list load completes, you probably want to move it this class.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// Start an async task to get the list of repos.
+        /// Note: In the case of reentrancy, _latest is reset. Prior task values are abandoned.
+        /// </summary>
+        /// <param name="projectId">GCP project id.</param>
+        public async Task StartListRepoTaskAsync(string projectId)
+        {
+            Debug.WriteLine(nameof(StartListRepoTaskAsync));
+            _latest = AsyncPropertyUtils.CreateAsyncProperty(CsrUtils.GetCloudReposAsync(projectId));
+            RaiseAllPropertyChanged();
+            await _latest.ValueTask;
```
ILMTitan: After this await, you probably want to do something like
```
if(!Value.Contains(SelectedRepository)) {
    SelectedRepository = Value.FirstOrDefault();
}
```

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -0,0 +1,92 @@
...
+                return _latest.Value.Any() ? DisplayOptions.HasItems : DisplayOptions.NoItems;
+            }
+        }
+
+        /// <summary>
+        /// Start an async task to get the list of repos.
+        /// Note: In the case of reentrancy, _latest is reset. Prior task values are abandoned.
+        /// </summary>
+        /// <param name="projectId">GCP project id.</param>
+        public async Task StartListRepoTaskAsync(string projectId)
```
ILMTitan: Since this should never throw, and you are tracking the task with `_latest`, you can probably make this method `async void`.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
Deren-Liao: The selected repository is updated automatically when the ItemsSource is updated.
I added PropertyChanged event listener to make it explicit. 
Please note, I need to call ValidateInput when selected Repository changes. So, I keep the changed code in view model class.

GoogleCloudExtension/GoogleCloudExtension.GitUtils/GitRepository.cs
Deren-Liao: No, part of the code is missing. Add back.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowContent.xaml
Deren-Liao: Thanks

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: I thought the original await would catch exception.  
I changed the code not to use AsyncProperties so as to make sure exception is caught.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
Deren-Liao: I want it throw so as to show error dialog.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
Deren-Liao: I added PropertyChanged event handler to achieve the same result.

Deren-Liao:
PTAL

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepositories/AsyncRepositoriesTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            Assert.AreEqual(AsyncRepositories.DisplayOptions.Pending, _testObject.DisplayState);
+            Assert.IsNull(_testObject.Value);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(TestException))]
+        public void TestCatchException()
+        {
+            AsyncRepositories.GetCloudReposAsync = async (projectId) =>
```
ILMTitan: You are going to be setting this function in nearly every.  I suggest you set it up in `TestInitialize` or even `ClassInitialize`. I also suggest you move the `TaskCreationSource` to a field.

```
_tcs = new TaskCompletionSource<IList<Repo>>();
_getCloudReposMock = new Mock<string, Func<Task<IList<Repo>>>>()
_getCloudReposMock.Setup(f => f()).Returns(() => {
    var task = _tcs.Task;
    // Reset the tcs to make it easier to call and set multiple times.
    _tcs = new TaskCompletionSource<IList<Repo>>();
    return task;
);
AsyncRepositories.GetCloudReposAsync = _getCloudReposMock.Object;
```
With that in the setup, this test becomes simply
```
[TestMethod]
[ExpectedException(typeof(TestException))]
public async Task TestCatchException()
{
    _tcs.SetException(new TestException());
    try {
        await _testObject.StartListRepoTaskAsync("projectId"));
    } finally {
       Assert.IsNotNull(task.Result);
    }
}
```

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepositories/AsyncRepositoriesTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            };
+
+            var task = AsyncWait(() => _testObject.StartListRepoTaskAsync("projectid"));
+            task.Wait();
+            Assert.IsNotNull(task.Result);
+            throw task.Result;          
+        }
+
+        [TestMethod]
+        public void TestPendingAndCompletion()
```
ILMTitan: This would become:
```
task = _testObject.StartListRepoTaskAsync("projectid");
Assert.AreEqual(AsyncRepositories.DisplayOptions.Pending, _testObject.DisplayState);
Assert.IsNull(_testObject.Value);

_tcs.SetResult(_testRepos);
await task;

Assert.AreEqual(AsyncRepositories.DisplayOptions.HasItems, _testObject.DisplayState);
Assert.AreSame(_testRepos, _testObject.Value);
```

ILMTitan: I would separate this into one test for pending and one test for completion. If you do more setup in the `[TestInitialize]`, the the pending setup will be one line (starting the task), and the completion setup will be two (if you move the `SetResult` above the `StartTask`).

Deren-Liao:
I am moving the unit tests to another PR. 
Your input on tests are good. I'll make changes later.

Deren-Liao:
Thanks a lot!

PR #719, opened 2017-06-20
Properly enumerating projects in solution folders.

### LINKED ISSUE #593: When Web application is in a solution folder Then Google Cloud tools context menu item is disabled
If you put your web application inside a solution folder it always shows up as disabled. Changing target framework does not affect it.

image

ivannaranjo:
The problem that was described was caused because the extension was only enumerating those projects that were placed directly in the root of the logical solution tree. The new code will recurse through `SolutionFolders` to find **all** of the projects that are part of the solution. Then from those projects we will select the right one.

Fixes #593

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/SolutionUtils/SolutionHelper.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/SolutionUtils/SolutionHelper.cs
@@ -35,6 +35,9 @@ namespace GoogleCloudExtension.SolutionUtils
     {
         private const string ProjectJsonName = "project.json";
 
+        // This is the GUID for solution folder items.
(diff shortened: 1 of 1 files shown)


GoogleCloudExtension/GoogleCloudExtension/SolutionUtils/SolutionHelper.cs
```diff
@@ -114,6 +118,61 @@ private IParsedProject GetSelectedProject()
...
+        private IList<Project> GetSolutionProjects()
+        {
+            List<Project> result = new List<Project>();
+
+            var rootProjects = _solution.Projects;
+            foreach (Project item in rootProjects)
+            {
+                if (item.Kind == SolutionFolderKind)
+                {
+                    result.AddRange(GetSolutionFolderProjects(item));
```
ILMTitan: I wish C# had `yield foreach`. This is a perfect candidate for it.

Issue #722, opened 2017-06-20
Code Coverage reporting braces as uncovered lines.


ILMTitan:
The code coverage report is counting some braces as uncovered lines. According to my research, this happens when you test code coverage on a debug dll. We should change the tests to only run the release configuration dlls.

Additionally, we should filter out `GoogleCloudExtension.Resources`, as it is a generated file with usage tested by *ensure_no_unused_strings.sh*.

PR #724, opened 2017-06-20
Fix invalid code coverage.

ILMTitan:
Issue #722.
Only run tests on release builds.
Filter out GoogleCloudExtension.Resources.


ivannaranjo:
LGTM

PR #728, opened 2017-06-21
Master fixing 560

### LINKED ISSUE #560: Clear corrupted Windows Credential
{"Key not valid for use in specified state.\r\n"}

   at System.Security.Cryptography.ProtectedData.Unprotect(Byte[] encryptedData, Byte[] optionalEntropy, DataProtectionScope scope)
   at GoogleCloudExtension.Accounts.WindowsCredentialsStore.LoadEncryptedCredentials(String path) 

This error goes on and one and there is no way to correct it.

ivannaranjo:
This PR fixes #560 by adding a handler for the `CryptographicException` exception thrown when the encrypted user's credentials cannot be loaded. This typically happens when the user resets the password, since the key used for the encryption process is derived from the password. In that case the user will lose all of the credentials, but they are easy to re-create if necessary.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
@@ -17,6 +17,7 @@ using GoogleCloudExtension.DataSources;
 using GoogleCloudExtension.GCloud;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
@@ -75,6 +76,7 @@ namespace GoogleCloudExtension.Accounts
                 result = Directory.EnumerateFiles(instanceStoragePath)
                     .Where(x => Path.GetExtension(x) == PasswordFileExtension)
                     .Select(x => LoadEncryptedCredentials(x))
+                    .Where(x => x != null)
                     .OrderBy(x => x.User);
             }
             _credentialsForInstance[instancePath] = result;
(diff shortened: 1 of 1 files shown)


GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -125,13 +127,43 @@ public string GetStoragePathForInstance(Instance instance)
...
+                return new WindowsInstanceCredentials { User = userName, Password = Encoding.UTF8.GetString(passwordBytes) };
+            }
+            catch (CryptographicException)
+            {
+                Debug.WriteLine($"Failed to decrypt credentials from: {path}");
+                try
+                {
+                    File.Delete(path);
+                }
+                catch (IOException)
```
ILMTitan: Do we want to fail silently on `IOException`? That feels like the type of thing a user can fix, and therefore should be informed of.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -125,13 +127,43 @@ public string GetStoragePathForInstance(Instance instance)
...
+                {
+                    Debug.WriteLine($"Failed cleaning corrupted credentials {path}");
+                }
+                return null;
+            }
+            catch (IOException)
```
ILMTitan: Here too.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ivannaranjo: If the code failed reading the file with an `IOException` then that means that something at the IO level went wrong. We attempt to delete the file as a cleanup step as last ditch effort but really there's nothing else to do. This is probably either a bad hard drive or the permissions getting really really screw up. The user cannot really do anything while this process is happening.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ivannaranjo: Same as above, the file failed to load, either permissions screw up or harddrive failure, either way we give up.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ILMTitan: I'm not saying the user can fix it while the process is happening, but they are likely able to fix it and try again. Replacing a hard drive and changing permissions on a folder a both things a user can do (or at least get tech support to do for them). Telling the user that there is a problem that they might be able to solve is the first step of solving the problem.

ivannaranjo: Added #732 to track this work. I do not want to add new UI or a new error string at this point. This fix is sufficient for Q2, we can add more complex behavior in Q3.

ivannaranjo:
PTAL p

PR #730, opened 2017-06-21
Enhance error message 

Deren-Liao:
When Remote PowerShell fails to connect due to credential error etc, currently, it shows error with internal PowerShell commands etc.  It confuses user. 
This PR is to show some generic error message.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerFirewallPort.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerFirewallPort.cs
@@ -102,35 +102,32 @@ namespace GoogleCloudExtension.AttachDebuggerDialog
 
         /// <summary>
         /// Check if GCE firewall rules include a rule that enables the port to target GCE VM.
-        /// A firewall rule contains tag, 
-        /// if the GCE instance also has the tag, the rule is applied to the GCE instance.
+        /// If a firewall rule contains tag, 
+        /// and the GCE instance also has the tag, the rule is applied to the GCE instance.
         /// </summary>
         public async Task<bool> IsPortEnabled()
         {
             string portTag = PortInfo.GetTag(_gceInstance);
 
             // If the instance does not contain the tag, the firewall rule is not set.
-            if (_gceInstance.Tags?.Items?.Contains(portTag) ?? false)
+            if (_gceInstance.Tags?.Items?.Contains(portTag) != true)
             {
-                var rules = await _dataSource.GetFirewallListAsync();
-                foreach (var rule in rules)
-                {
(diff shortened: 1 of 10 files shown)


GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerSettingsStore.cs
```diff
@@ -35,20 +35,21 @@ internal static class AttachDebuggerSettingsStore
...
         {
+            List<InstanceDefaultUser> results = null;
             if (File.Exists(DefaultUsersSettingsFullPath))
             {
                 string jsonText = File.ReadAllText(DefaultUsersSettingsFullPath);
                 if (!string.IsNullOrWhiteSpace(jsonText))
                 {
                     try
                     {
-                        return JsonConvert.DeserializeObject<List<InstanceDefaultUser>>(jsonText);
```
ILMTitan: It is purely a matter of opinion, but I prefer the way the function was written originally. The inner return was the focus of the method, and would not be missed when the function is read in the future.

GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerFirewallPort.cs
```diff
@@ -116,13 +116,14 @@ public int WaitForFirewallRuleTimeInSeconds()
...
-                    if (rule.TargetTags?.Contains(portTag) ?? false)
+                    if (rule?.TargetTags?.Contains(portTag) ?? false)
                     {
                         continue;   // Skip, rules does not contain the tag.
                     }
                     foreach (var allowed in rule.Allowed)
                     {
-                        if (allowed.IPProtocol == "tcp" && allowed.Ports.Any(y => y == PortInfo.Port.ToString()))
+                        if (allowed?.IPProtocol == "tcp" && 
+                            (allowed?.Ports?.Any(y => y == PortInfo.Port.ToString()) ?? false))
```
ILMTitan: Is this fixing an error you discovered?

Deren-Liao: No, I reviewed code and see this potentially throws null reference error.

GoogleCloudExtension/GoogleCloudExtension/AttachDebuggerDialog/AttachDebuggerSettingsStore.cs
Deren-Liao: I am worried DeserializeObject may return null in some extreme cases.

Deren-Liao:
PTAL.

PR #731, opened 2017-06-21
Move GoogleCloudExtension start parameters to the .csproj file.

ILMTitan:
This PR moves the default "Run Visual Studio Experimental Instance" properties from the *GoogleCloudExtension.csproj.user* file to the *GoogleCloudExtension.csproj* file. This lets us remove and ignore the *csproj.user* file from git.

Running still works in this configuration, and changing the location to run still saves to the *csproj.user* file. The *csproj.user* properties override the *.csproj* properties, so users with a different VS install dir will open the run configuration, select their *devenv.exe* path, and be able to debug.


ILMTitan:
I argue this is less hostile towards not installing VS in exactly the same spot.

The flow for both cases is:
 * A user notices the default launch parameter is bad
 * They go to GoogleCloudExtension => Properties => Debug
 * They set the Start external program path to their VS install
 * The new path is saved to the *GoogleCloudExtension.csproj.user*
     * Note: The properties in *GoogleCloudExtension.csproj.user* override those in *GoogleCloudExtension.csproj*

The difference is now git sees it.

Right now, if a user has VS installed in a different spot, their *csproj.user* is always going to show up as changed with their path. They they have to do something special to cause git to ignore that change.

With this change, the user never sees the changed *csproj.user* because git ignores it, as it should.

ivannaranjo:
It doesn't matter, this is how projects and solutions are structured in Visual Studio. We will be fighting with Visual Studio every single time. And I do not want to require VS to be installed in a particular location.

I would rather live in a world where we remove the .user.csproj (add it to .gitignore) and we all have our own copy of it with the settings for launching the debugger, and others. So if you want to do that I would accept it.

ILMTitan:
Closing this PR until after the M2 release.

PR #739, opened 2017-06-26
Fix null reference pointer error on account reset event

Deren-Liao:


ivannaranjo:
In general a small description of the bug and how the fix fixes the bug would be appreciated. In this case is trivial, but noting why the value would be `null` is good for future reference as well as to help the reviewers.

Deren-Liao:
Sure thing.  I'll add more description in the future.
Thanks.

Issue #736, opened 2017-06-26
Cannot create ASP.NET Core projects using Google templates in VS 2017


ivannaranjo:
When attempting to create new ASP.NET Core projects in VS 2017 using the Google templates the process fails with the following error message:
```
---------------------------
Microsoft Visual Studio
---------------------------
Operation could not be completed. The project file 'c:\users\ivann\documents\visual studio 2017\Projects\GoogleAspNetCoreMvc1\GoogleAspNetCoreMvc1\GoogleAspNetCoreMvc1.xproj' cannot be migrated. See the migration report for details.
---------------------------
OK   
---------------------------
```

The error message cites a migration report but this migration report cannot be, possibly because the template expansion never finished.

I have not seen this issue before as the template expansion was working correctly before.

My version of VS is 15.2 (26430.4) and verified it reproes as well with (26430.12).

This bug is a shipblocker.

ivannaranjo:
Attempting to expand the template on VS 2015 yields this error:
---------------------------
Microsoft Visual Studio
---------------------------
The imported project "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DotNet\Microsoft.DotNet.Props" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.  c:\users\ivann\documents\visual studio 2015\Projects\TestProjectMvc\TestProjectMvc\TestProjectMvc.xproj
---------------------------
OK   
---------------------------

ivannaranjo:
Seems to only fail on Windows Server 2016, and possibly Windows 10.

ILMTitan:
Does the versions of VS 2015 it is failing on have Microsoft .NET Core Tools? That is required for the .NET Core projects.

ILMTitan:
I just ran VS 2017 as a debug process under VS 2015, and the template worked. I simultaneously ran a VS 2017 as a standalone process and the templates failed.

ILMTitan:
It looks like this fails if Visual Studio 2017 is not run as an administrator.

ivannaranjo:
I have .NET Core tools installed in all versions of VS, so I can test the templates, deployments etc... Any idea why the .props file would not be found?

ILMTitan:
The only reason I can think of is if the *.props* file is not there.
I have not changed the *.xproj* files at all. Looking at them, they seem to be doing the correct things (using variables to find the files).

```
  <PropertyGroup>
[...]

PR #743, opened 2017-06-27
Fix for the VS 2017 templates.

ILMTitan:
Avoid the Access denied exception that was breaking the templates in VS 2017.

### DIFF
--- a/GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/out.project.json
+++ b/GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/out.project.json
@@ -5,22 +5,22 @@
       "version": "1.0.3",
       "type": "platform"
     },
-    "Microsoft.AspNetCore.Diagnostics": "1.0.0",
-    "Microsoft.AspNetCore.Mvc": "1.0.4",
+    "Microsoft.AspNetCore.Diagnostics": "1.0.2",
+    "Microsoft.AspNetCore.Mvc": "1.1.2",
     "Microsoft.AspNetCore.Razor.Tools": {
       "version": "1.0.0-preview4-final",
       "type": "build"
     },
-    "Microsoft.AspNetCore.Server.IISIntegration": "1.0.3",
-    "Microsoft.AspNetCore.Server.Kestrel": "1.0.4",
-    "Microsoft.AspNetCore.StaticFiles": "1.0.3",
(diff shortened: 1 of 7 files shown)


ILMTitan:
Fix for the first part of Issue #736.

ivannaranjo:
See my comment in the bug. We need to make sure that the templates work in VS 2017 without VS 2015, or the .NET Tools for VS 2015, installed in the system. That is the root cause of the issue. Once I installed the tools in my machines everything worked. But we cannot ask the users to do the same, as it requires having VS 2015 in the system as well.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/out.project.json
```diff
@@ -5,22 +5,22 @@
...
-    "Microsoft.AspNetCore.Server.Kestrel": "1.0.4",
-    "Microsoft.AspNetCore.StaticFiles": "1.0.3",
-    "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.2",
+    "Microsoft.AspNetCore.Server.IISIntegration": "1.1.1",
+    "Microsoft.AspNetCore.Server.Kestrel": "1.1.1",
+    "Microsoft.AspNetCore.StaticFiles": "1.1.1",
+    "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.1.1",
     "Microsoft.Extensions.Configuration.Json": "1.0.2",
-    "Microsoft.Extensions.Logging": "1.0.2",
+    "Microsoft.Extensions.Logging": "1.1.1",
```
ivannaranjo: These seem suspect for a 1.0 project, do they work well with .NET Core 1.0? Have you tried this?

ILMTitan: I did and it worked fine.

ivannaranjo: Ok, but is there a reason to update them? Does a comparable 1.0 template from Microsoft do the same thing ?

ILMTitan: When I tested it, there were some errors regarding out of date dependencies. Not sure how I missed it when I broke them in the first place.

PR #752, opened 2017-07-06
Master gcs browser ui feedback

ivannaranjo:
Updated the UI to better mach the feedback given recently about grids. This PR brings the `DataGrid` used in the GCS browser more inline with how the `DataGrid` is used in other parts of the UI.

In particular this PR adds alternating columns, removes borders and updates the selection (active and inactive) to better match the UI.

screen shot 2017-07-06 at 2 00 13 pm

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Theming/CommonResources.xaml
+++ b/GoogleCloudExtension/GoogleCloudExtension/Theming/CommonResources.xaml
@@ -12,7 +12,6 @@
 
     <!-- Text style. -->
     <Style x:Key="CommonTextStyle" BasedOn="{StaticResource CommonTextStyleBase}" TargetType="{x:Type TextBlock}">
-        <Setter Property="Foreground" Value="Black" />
         <Setter Property="TextWrapping" Value="Wrap" />
     </Style>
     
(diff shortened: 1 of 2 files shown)


GoogleCloudExtension/GoogleCloudExtension/Theming/CommonResources.xaml
```diff
@@ -12,7 +12,6 @@
 
     <!-- Text style. -->
     <Style x:Key="CommonTextStyle" BasedOn="{StaticResource CommonTextStyleBase}" TargetType="{x:Type TextBlock}">
-        <Setter Property="Foreground" Value="Black" />
```
Deren-Liao: Why do you remove this property?  I am afraid some UIs that uses this style may become wrong.  You can create another style for your use case.

ivannaranjo: Because the default is black and it prevents any change in colors for text that derives from it. Without this change I cannot change the color of the text during selection.

ivannaranjo: I'll see if I can make the more conservative change, with a separate style.

ivannaranjo:
Added a new style, used it from the data grid. PTAL

PR #758, opened 2017-07-10
Template tests

ILMTitan:
Add integration tests to compile the templates.


GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -44,9 +45,16 @@ public class GoogleProjectTemplateWizard : IGoogleProjectTemplateWizard
             WizardRunKind runKind,
             object[] customParams)
         {
-            string projectId = PromptPickProjectId();
+            var dte = (DTE)automationObject;
+            bool isEmbedded = dte.CommandLineArguments.Contains("-Embedding");
```
Deren-Liao: Do you think it's safer to use   dte?.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
```diff
@@ -44,9 +45,16 @@ public class GoogleProjectTemplateWizard : IGoogleProjectTemplateWizard
...
+
+            string projectId = isEmbedded ?
```
Deren-Liao: My understanding, when it is "isEmbedded"  it is driven under tests ?    May you add a comment to be clear?

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <summary>
+        /// Get the COM object from the running object table that matches the predicate.
+        /// </summary>
+        /// <param name="predicate">The test to check the monikers against.</param>
+        public static object GetRunningObject(Func<IMoniker, IBindCtx, bool> predicate)
+        {
+            IBindCtx bindCtx;
+            Marshal.ThrowExceptionForHR(CreateBindCtx(0, out bindCtx));
+            IRunningObjectTable rot;
+            bindCtx.GetRunningObjectTable(out rot);
```
Deren-Liao: Check return value

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+            else
+            {
+                return null;
+            }
+        }
+
+        private static IEnumerable<IMoniker> GetRunningMonikers(this IRunningObjectTable rot)
+        {
+            IEnumMoniker monikersEnum;
+            rot.EnumRunning(out monikersEnum);
```
Deren-Liao: Check HRESULT.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
ILMTitan: Yes. Every example I have seen, including the NuGet wizard written by Microsoft, does this.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+            return monikersEnum.ToEnumerable();
+        }
+
+        private static IEnumerable<IMoniker> ToEnumerable(this IEnumMoniker monikersEnum)
+        {
+            int fetchSize = 1;
+            var monikers = new IMoniker[fetchSize];
+            IntPtr fetchedCount = IntPtr.Zero;
+            while (monikersEnum.Next(fetchSize, monikers, fetchedCount) == 0)
```
Deren-Liao: The annoying part working with COM (c++) code is that you have to check return value everywhere. But that's how it is.  
Please check the HRESULT here too.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
ILMTitan: Yes, it is a command line flag that prevents the primary window from popping up.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
Deren-Liao: define S_OK.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
ILMTitan: This one does not have a return value.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
ILMTitan: This one does not have a return value either.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComMessageFilter.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace ProjectTemplate.Tests
+{
+    /// <summary>
+    /// A COM IMessageFilter that prevents call rejection.
+    /// <see href="https://msdn.microsoft.com/en-us/library/windows/desktop/ms693740"/>
+    /// </summary>
+    /// <seealso href="https://www.viva64.com/en/b/0169/"/>
+    public class ComMessageFilter : MarshalByRefObject, IDisposable, IMessageFilter
+    {
+
```
Deren-Liao: extra empty line

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComMessageFilter.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Management;
+using System.Runtime.ExceptionServices;
+
+namespace ProjectTemplate.Tests
+{
+    public static class ProcessHelper
```
Deren-Liao: Add comment

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+            var killedProcesses = new List<Process>();
+            foreach (ManagementBaseObject mo in searcher.Get())
+            {
+                try
+                {
+                    int processId = Convert.ToInt32(mo["ProcessID"]);
+                    var childProcess = Process.GetProcessById(processId);
+                    killedProcesses.AddRange(KillProcessTree(childProcess));
+                }
+                catch (AggregateException e)
```
Deren-Liao: I feel it's odd here because InnerException can itself be AggregateException too..    Maybe you just catch all Exception?

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                process.Kill();
+                killedProcesses.Add(process);
+            }
+            catch (Exception e)
+            {
+                exceptions.Add(e);
+            }
+
+            RethrowAnyExceptions(exceptions);
```
Deren-Liao: Throw AggregateException or other custom exception type here makes sense.  Keep exception context actually confuses  caller. Suppose you have 5 processes, 1 failed and throws an exception.  By throw the exception with context, it looks like one stopped execution immediately.   In fact you continue to kill other processes.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Reflection;
```
Deren-Liao: Just follow existing pattern, we add license disclaimer  at AssemblyInfo.cs too

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        private static IEnumerable<ErrorItem> ErrorItems =>
+            ((IEnumerable)Dte.ToolWindows.ErrorList.ErrorItems).Cast<ErrorItem>();
+        private static DTE2 Dte => s_visualStudio.Dte;
+        [SuppressMessage("ReSharper", "SuspiciousTypeConversion.Global")]
+        private static Solution2 Solution => (Solution2)Dte.Solution;
+        private static string SolutionFolderPath => Path.Combine(Path.GetTempPath(), SolutionName);
+
+        [SuppressMessage("ReSharper", "UnusedAutoPropertyAccessor.Global")]
+        [SuppressMessage("ReSharper", "MemberCanBePrivate.Global")]
+        public TestContext TestContext { get; set; }
```
Deren-Liao: Move the public down under private properties.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            Process[] msBuildProcesses = Process.GetProcessesByName("MSBuild");
+            foreach (Process killedProcess in msBuildProcesses.SelectMany(p => p.KillProcessTree()).ToList())
+            {
+                killedProcess.WaitForExit();
+            }
+        }
+
+        [ClassCleanup]
+        public static void CleanupClass()
+        {
```
Deren-Liao: Nit: Try to delete the created solution and project directory?

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                    Directory.Delete(SolutionFolderPath, true);
+                }
+                catch (UnauthorizedAccessException)
+                {
+                    // MSBuild sometimes persists and holds a handle on a file in the solution package directory.
+                    KillMsBuild();
+                    Directory.Delete(SolutionFolderPath, true);
+                }
+            }
+            while (Directory.Exists(SolutionFolderPath))
```
Deren-Liao: Set a timeout.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+            Directory.CreateDirectory(SolutionFolderPath);
+        }
+
+        private static void KillMsBuild()
+        {
+            Process[] msBuildProcesses = Process.GetProcessesByName("MSBuild");
+            foreach (Process killedProcess in msBuildProcesses.SelectMany(p => p.KillProcessTree()).ToList())
+            {
+                killedProcess.WaitForExit();
```
Deren-Liao: Use the timeout version and set a timeout

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            TestContext.WriteLine($"Testing template {TemplateName}");
+
+            CreateProjectFromTemplate();
+            Solution.SolutionBuild.Build(true);
+
+            Assert.AreEqual(vsBuildState.vsBuildStateDone, Solution.SolutionBuild.BuildState, TemplateName);
+            string descriptions = string.Join("\n", ErrorItems.Select(e => e.Description));
+            Assert.AreEqual(0, ErrorItems.Count(), $"{TemplateName} error descriptions: {descriptions}");
+            Assert.AreEqual(0, Solution.SolutionBuild.LastBuildInfo, TemplateName);
+        }
```
Deren-Liao: Maybe add a check that the built assembly file is there?

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/Properties/AssemblyInfo.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: I would rather keep them around so I can use them to debug any issues.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/VisualStudioWrapper.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            try
```
Deren-Liao: add  if ( disposing ) ..     I don't think it makes sense to let GC to kill the DTE process.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: I think public should be more visible, and therefore should be above private. Our style guide only says fields should be above properties.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: Done, Changed the `while` to an `if`.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
Deren-Liao: Is this the one you use?  https://msdn.microsoft.com/en-us/library/windows/desktop/ms684004(v=vs.85).aspx
It says it returns E_UNEXPECTED  in some cases.
This function can return the standard return values E_UNEXPECTED and S_OK.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
Deren-Liao: https://msdn.microsoft.com/en-us/library/windows/desktop/ms678491(v=vs.85).aspx
This method can return the standard return values E_OUTOFMEMORY and S_OK.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/VisualStudioWrapper.cs
ILMTitan: The process is outside of the CLR, and can persist after the CLR dies, which is bad. The point of this class is to own the process and clean it up under every circumstance that it can.  This includes finalization.

The point of the dispose pattern is to only dispose *managed* resources during disposal, but to dispose *unmanaged* resources (like external processes) during both disposal and finalization.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
ILMTitan: Changed it to always throw `AggregateException`, even when there is just one exception.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
ILMTitan: I would rather avoid nesting `AggregateException`s.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
ILMTitan: Done.

ILMTitan:
PTAL.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/VisualStudioWrapper.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                Dte.Quit();
```
Deren-Liao: Is it safe to Quit() twice then?

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
Deren-Liao: I meant to change it to dte?.  so as to check if the input is null

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
Deren-Liao: The comment moved location.
For GetRunningObjectTable, MSDN says it can return E_UNEXPECTED .
https://msdn.microsoft.com/en-us/library/windows/desktop/ms684004(v=vs.85).aspx 
This function can return the standard return values E_UNEXPECTED and S_OK.

Deren-Liao:
* Can you check again on my comment about check HRESULT?  MSDN says they do return values other than S_OK. 
* I think it might be time to check in your changes to M3 instead of M2 .

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
ILMTitan: The managed wrapper `System.Runtime.InteropServices.ComTypes.IBindCtx` appears to convert that return to exceptions. `IBindCtx.GetRunningObjectTable(out IRunningObjectTable rot)` is `void`.

https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.comtypes.ibindctx.getrunningobjecttable(v=vs.110).aspx

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.cs
ILMTitan: Right, sorry.
No, the automationObject should never be null.

ILMTitan:
I am sure about the HRESULTs. The .NET methods for those two are void.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+            rot.EnumRunning(out monikersEnum);
+            return monikersEnum.ToEnumerable();
+        }
+
+        private static IEnumerable<IMoniker> ToEnumerable(this IEnumMoniker monikersEnum)
+        {
+            int fetchSize = 1;
+            var monikers = new IMoniker[fetchSize];
+            IntPtr fetchedCount = IntPtr.Zero;
+            while (monikersEnum.Next(fetchSize, monikers, fetchedCount) == HRESULT.S_OK)
```
ivannaranjo: Should you throw in case of an unexpected error?

ivannaranjo: If you don't check the `fetchedCount` how do you know how many items were actually added to the `monikers` array?

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+                    exceptions.Add(e);
+                }
+            }
+
+            try
+            {
+                process.Kill();
+                killedProcesses.Add(process);
+            }
+            catch (Exception e)
```
ivannaranjo: You should check if it is a critical exception as well, and let the process die. Even if it is testing code this could be hiding bad bugs in your test code.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace ProjectTemplate.Tests
+{
+    /// <summary>
+    /// The methods of this class help to manage processes.
+    /// </summary>
+    public static class ProcessHelper
+    {
+        /// <summary>
+        /// Kills the given process and all decenant processes.
```
ivannaranjo: decenant => descendant

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                // Allow Directory.Delete to finish before creating.
+                Thread.Sleep(200);
+            }
+            Directory.CreateDirectory(SolutionFolderPath);
+        }
+
+        private static void KillMsBuild()
+        {
+            Process[] msBuildProcesses = Process.GetProcessesByName("MSBuild");
+            foreach (Process killedProcess in msBuildProcesses.SelectMany(p => p.KillProcessTree()).ToList())
```
ivannaranjo: Shouldn't `KillProcessTree()` do the waiting (optionally) for you?

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                {
+                    // MSBuild sometimes persists and holds a handle on a file in the solution package directory.
+                    KillMsBuild();
+                    Directory.Delete(SolutionFolderPath, true);
+                }
+            }
+            if (Directory.Exists(SolutionFolderPath))
+            {
+                // Allow Directory.Delete to finish before creating.
+                Thread.Sleep(200);
```
ivannaranjo: This can make the test flaky, do we have a way to synchronously wait for the directory to be deleted?

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                Solution.Remove(project);
+            }
+        }
+
+        [TestMethod]
+        [DataSource(
+            "Microsoft.VisualStudio.TestTools.DataSource.CSV",
+            @"|DataDirectory|\Templates.csv",
+            "Templates#csv",
+            DataAccessMethod.Random)]
```
ivannaranjo: Does this mean that the order changes every time? Wouldn't this make reproducing the test failure difficult? I know that we don't want to depend on the order of the entries on the .csv file but we should be able to reproduce the test failure reliably.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/Templates.csv
```diff
@@ -0,0 +1,5 @@
+TemplateName
+GoogleAspnetMvc
+GoogleAspNetCoreMvc
+GoogleAspnetWebApi
+GoogleAspNetCoreWebApi
```
ivannaranjo: Nit, have a \n at the end of the last line.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/VisualStudioWrapper.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// If that fails, it tries to get the running Visual Studio 2015 instance and create a new experimental
+        /// instance from that path.
+        /// </remarks>
+        /// <returns>A new <see cref="VisualStudioWrapper"/>.</returns>
+        public static VisualStudioWrapper CreateExperimentalInstance()
+        {
+            try
+            {
+                // Try the devenv on the path first.
+                return new VisualStudioWrapper("devenv", "/rootSuffix Exp");
```
ivannaranjo: Idea, you could specify your own suffix, install the extension there and then at the end of the tests dispose of that cleanly. That way you could have a completely separate "instance" of VS to run the tests.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/VisualStudioWrapper.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        private bool IsMatchingMonikerName(IMoniker moniker, IBindCtx bindCtx)
+        {
+            string name;
+            try
+            {
+                moniker.GetDisplayName(bindCtx, null, out name);
+            }
+            catch
```
ivannaranjo: You should check for critical exceptions here, you don't want a corrupted environment to run your tests.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
ILMTitan: Because this is cleanup code, I want to try to clean up any external resources (processes) we are using, even if the exception is a StackOverflow or AccessViolation.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: Added `FileSystemWatcher.WaitForChanged`.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
ILMTitan: I was waffling on that, but I think you are right. Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/Templates.csv
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/VisualStudioWrapper.cs
ILMTitan: That would be nice, but I am currently relying on the *GoogleCloudExtension.csproj* to deploy the extension for me. To not do that would require me to find the *GoogleCloudExtension.vsix* within the project output, something that is more complicated that it should be.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
ILMTitan: When `fetchSize` is 1, `fetchedCount` may be `null` even when 1 item was actually fetched. The return code gives all the necessary information for a `fetchSize` of 1.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProcessHelper.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ComHelper.cs
ILMTitan: Done.

GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/VisualStudioWrapper.cs
ILMTitan: Changed it to only catch the documented exceptions. The underlying C method can return `E_NOTIMPL` and `MK_E_EXCEEDEDDEADLINE`, which map to `NotImplementedException` and the generic `COMException` as described [here](https://docs.microsoft.com/en-us/dotnet/framework/interop/how-to-map-hresults-and-exceptions).

ILMTitan:
PTAL

Deren-Liao:
Please check the build failure.

PR #769, opened 2017-07-26
Fix template tests

ILMTitan:
These are the same changes as the previously reverted Template Tests PR.
The only change a Write-Host call in the *TestScript.ps1*.
The previous changes were hanging the build, but these do not for some reason.

### DIFF
--- a/GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
+++ b/GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
@@ -1,0 +1,184 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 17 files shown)


GoogleCloudExtension/ProjectTemplates/ProjectTemplate.Tests/ProjectTemplatesTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        [ClassCleanup]
+        public static void CleanupClass()
+        {
+            s_visualStudio.Dispose();
+        }
+
+        [TestInitialize]
+        public void BeforeEach()
```
Deren-Liao: Did not notice this before.  Is it safe just delete the method?

TestScript.ps1
```diff
@@ -1,10 +1,16 @@
+Param([string]$Configuration)
+
+if(-not $Configuration) {
```
Deren-Liao: nit: Add space in between if and (

ivannaranjo:
LGTM

PR #778, opened 2017-08-16
Set template framework to latest

ILMTitan:
Update the template wizard to set template projects to the latest supported target framework version.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModelBase.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModelBase.cs
@@ -61,7 +61,11 @@ namespace GoogleCloudExtension.CloudExplorerSources.PubSub
 
         public ITopicItem Item { get; }
 
-        public event EventHandler ItemChanged;
+        event EventHandler ICloudExplorerItemSource.ItemChanged
+        {
+            add { }
+            remove { }
+        }
 
         protected TopicViewModelBase(
             IPubsubSourceRootViewModel owner,
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModelBase.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModelBase.cs
@@ -61,11 +61,7 @@ namespace GoogleCloudExtension.CloudExplorerSources.PubSub
 
         public ITopicItem Item { get; }
 
-        event EventHandler ICloudExplorerItemSource.ItemChanged
-        {
-            add { }
-            remove { }
-        }
+        public event EventHandler ItemChanged;
 
         protected TopicViewModelBase(
             IPubsubSourceRootViewModel owner,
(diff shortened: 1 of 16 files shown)


GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/TopicViewModelBase.cs
```diff
@@ -61,7 +61,11 @@ internal abstract class TopicViewModelBase : TreeHierarchy, ITopicViewModelBase
 
         public ITopicItem Item { get; }
 
-        public event EventHandler ItemChanged;
+        event EventHandler ICloudExplorerItemSource.ItemChanged
```
ivannaranjo: This change seems unrelated to the rest of the PR.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetWebApi/GoogleAspnetWebApi.csproj
```diff
@@ -88,7 +88,6 @@
     <None Include="Controllers\HomeController.cs" />
     <None Include="Controllers\ValuesController.cs" />
     <None Include="Global.asax.cs" />
-    <None Include="Global.asax" />
```
ivannaranjo: This seems suspect, why removing this file?

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/GoogleAspnetMvc.csproj
```diff
@@ -67,7 +67,6 @@
     <None Include="Controllers\HomeController.cs" />
     <None Include="Controllers\ManageController.cs" />
     <None Include="Global.asax.cs" />
-    <None Include="Global.asax" />
```
ivannaranjo: This seems suspect, why removing this file?

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/Project.xproj
```diff
@@ -10,7 +10,6 @@
     <RootNamespace>$safeprojectname$</RootNamespace>
     <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">.\obj</BaseIntermediateOutputPath>
     <OutputPath Condition="'$(OutputPath)'=='' ">.\bin\</OutputPath>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
```
ivannaranjo: This seems suspect, you remove it but not replacement for this line is offered? Was this line wrong before?

ivannaranjo:
Looked at the change, and while it looks good, the problem with this approach is that you never actually update the dependencies in the project.json nor generate an updated <projec>.csproj. Meaning that yes, you change the target framework version, but the project is still using ASP.NET Core 1.0, when it should be using the version of ASP.NET that matches the target version.

We really need to have separate 1.1, and now 2.0, project templates.

GoogleCloudExtension/ProjectTemplates/GoogleAspnetMvc/GoogleAspnetMvc.csproj
ILMTitan: This is a duplicate in the project. Removing this entry makes the file only show up once in the solution explorer.

ILMTitan:
I am going to close this, as the framework chooser dialog is a better solution.

PR #779, opened 2017-08-17
Change Template source directory layout.

ILMTitan:
Change template source directory layout to reduce character count below the limit when deploying to the experimental instance. Remove dependency on MSBuilder.

Note: this does not change the layout of the Template when it is created. The files deleted from GoogleAspNetCoreMvc/wwwroot/lib are parts of bower packages that are restored when a package restore is run. The files moved in GoogleAspNetWebApi are placed correctly because the TargetFolderName of the vstemplate file is changed to a path to the correct location. The names of the vstemplate files themselves only change the names of folders but don't have any other visible impact.


ivannaranjo:
Do the files get restored in all scenarios? What packages are we talking about here?

ILMTitan:
The packages are defined in *bower.json* and are **bootstrap, jquery, jquery-validation and jquery-validation-unobtrusive**.

A package restore occurs immediately after the template is loaded, and before every build.

GoogleCloudExtension/ProjectTemplates/GoogleAspNetCoreMvc/GoogleAspNetCoreMvc.csproj
```diff
@@ -74,23 +74,6 @@
     <None Include="wwwroot\images\banner4.svg" />
     <None Include="wwwroot\js\site.js" />
     <None Include="wwwroot\js\site.min.js" />
-    <None Include="wwwroot\lib\bootstrap\dist\css\bootstrap-theme.css" />
```
ILMTitan: They are not brought back by `dotnet restore`, but are brought back by `dotnet publish`, which is the command that copies the *wwwroot* folder.

Issue #782, opened 2017-08-17
Update .Net Core templates to use the GA Diagnostics libraries.


ILMTitan:

iantalarico:
@ILMTitan FYI: These are now GA.

Issue #781, opened 2017-08-17
Update templates to use Asp.Net Core 2.0


ILMTitan:

ivannaranjo:
Actually, we should allow both 1.1 and 2.0. The built-in templates's wizard has a drop down to select what version of the runtime to target. We should follow that as well.

PR #792, opened 2017-09-05
Change templates to be a single template.

ILMTitan:
Create a single template with a version and type selection dialog.
This is a fix for issues #781 and #782.


ivannaranjo:
I'll go through the massive changelist to perform some spot checks, any guidance on what to expect would be great.

Also, is the feature complete? Should this be in its own branch until the templates are fully functional?

GoogleCloudExtension/GoogleCloudExtension.TemplateWizards/DelegatingTemplateWizard.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.VisualStudio.OLE.Interop;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.TemplateWizard;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.ComponentModel.Composition.Hosting;
+
+namespace GoogleCloudExtension.TemplateWizards
+{
+
```
ivannaranjo: Empty line.

GoogleCloudExtension/GoogleCloudExtension.TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+namespace GoogleCloudExtension.TemplateWizards
+{
+    /// <summary>
+    /// Wizard for a project template. Delegates to GoogleProjecTemplateSelectorWizard
+    /// in GoogleCloudExtension.dll. That class can not be used directly because
```
ivannaranjo: Could you ellaborate a bit about the issues? Open an issue in the github repo so we can track this issue and point to it.

GoogleCloudExtension/GoogleCloudExtension.Utils/PathUtils.cs
```diff
@@ -57,7 +57,22 @@ public static string GetCommandPathFromPATH(string commandName)
...
+        /// <param name="fromDirectory">The directory that is the start location of the relative path.</param>
+        /// <param name="toDirectory">The directory that is the end location of the relative path.</param>
+        /// <returns>
+        /// The relative path from <paramref name="fromDirectory"/> to <paramref name="toDirectory"/>
+        /// </returns>
+        public static string GetRelativePath(string fromDirectory, string toDirectory)
+        {
+            var fromUri = new Uri(fromDirectory.EnsureEndSeparator().Replace('\\', '/'));
+            var toUri = new Uri(toDirectory.EnsureEndSeparator().Replace('\\', '/'));
+            return fromUri.MakeRelativeUri(toUri).ToString().Replace('/', Path.DirectorySeparatorChar);
```
ivannaranjo: You are only using `Path.DirectorySeparatorChar` here, but not above, you should use it everywhere in this method to be consistent.

The documentation is not exactly clear on what the _relative path from one directory to another_ means, I would suggest adding an example.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+
+            return scannedTokens;
+        }
+
+        /// <summary>
+        /// Check if input is digits only.
+        /// return false if it is null.
+        /// Empty string is valid so it returns true.
+        /// </summary>
+        public static bool IsDigitsOnly(string text) => text != null && text.All(char.IsDigit);
```
ivannaranjo: Why is an empty string considered to be _all digits_? It should return `false` since by definition it does not contain digits.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+        ///         (b) right side of a double quotes pair.
+        ///         (c) the very last alone double quote.
+        ///         (e) the escaped double quote.
+        ///     If it is a single space, it could be
+        ///         (a) A space inside a double quotes pair.  Add it to currentToken.
+        ///         (b) else, it is a separator, add the currentToken to
+        /// </summary>
+        public static IEnumerable<string> SplitStringBySpaceOrQuote(string source)
+        {
+            if (source == null)
```
ivannaranjo: If the source is `null` or empty then you should probably return an empty enumerable no?

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+            bool hasLeadingBackslash = false;
+            List<string> scannedTokens = new List<string>();
+            foreach (char chr in source)
+            {
+                if (hasLeadingBackslash)
+                {
+                    hasLeadingBackslash = false;
+                    if (chr == '"')
+                    {
+                        currentToken.Append("\\\"");
```
ivannaranjo: This will add the string with two caracters \ and ". I would assume that you want to add only the char ", since the character \ is just the scape?

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+                        continue;
+                    }
+                    else
+                    {
+                        currentToken.Append("\\\\");
```
ivannaranjo: Same here, you shoulkd be adding a single \, not a double one.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+                    }
+                }
+
+                switch (chr)
+                {
+                    case '"':
+                        if (inDelimitedString)
+                        {
+                            if (currentToken.Length > 0)
```
ivannaranjo: You don't allow empty strings in the input? What if the input is: A B "" C?

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Are you assuming that the user will only escape the " char? What if they escape others, like spaces?

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+
+                    default:
+                        currentToken.Append(chr);
+                        break;
+                }
+            }
+
+            if (currentToken.Length > 0)
+            {
+                var splits = currentToken.ToString().Split(
```
ivannaranjo: How can this happen given the rules implemented above? The `currentToken` should not contain spaces, unless it is a delimited string, in which case you want to keep them.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+        public static int FirstNonSpaceIndex(string text)
+        {
+            if (text == null)
+            {
+                return -1;
+            }
+
+            for (int i = 0; i < text.Length; ++i)
+            {
+                if (text[i] != ' ')
```
ivannaranjo: Prefer using `Char.IsWhiteSpace()` since that will be a more comprehensive list of white space chars. Unless it has to specifically be only the ' ' char.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+        public static int LastNonSpaceIndex(string text)
+        {
+            if (text == null)
+            {
+                return -1;
+            }
+
+            for (int i = text.Length - 1; i >= 0; --i)
+            {
+                if (text[i] != ' ')
```
ivannaranjo: Same as above, prefer using the `Char.IsWhiteSpace()` method for this.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        /// <summary>
+        /// Adds a trailing slash if the directory is missing it.
+        /// </summary>
+        /// <param name="directoryString">The directory path string.</param>
+        /// <returns>The path to the directory with a trailing separator character.</returns>
+        public static string EnsureEndSeparator(this string directoryString)
+        {
+            directoryString.ThrowIfNull(nameof(directoryString));
+            if (!directoryString.EndsWith(Path.DirectorySeparatorChar.ToString()) && !directoryString.EndsWith("/"))
```
ivannaranjo: Why would the directory, on Windows, end with the '/' char?

ivannaranjo: This is not really a string util, this is more of a path util method however.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        /// <summary>
+        /// The preview version of ASP.NET Core used by VS 2015.
+        /// </summary>
+        public static readonly AspNetVersion AspNetCore1Preview = new AspNetVersion("1.0-preview");
+
+        /// <summary>
+        /// ASP.NET Core 1.0.
+        /// </summary>
+        public static readonly AspNetVersion AspNetCore1 = new AspNetVersion("1.0");
```
ivannaranjo: What happened to "1.1"?

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// ASP.NET Core 2.0.
+        /// </summary>
+        public static readonly AspNetVersion AspNetCore2 = new AspNetVersion("2.0");
+
+        /// <summary>
+        /// ASP.NET 4.
+        /// </summary>
+        public static readonly AspNetVersion AspNet4 = new AspNetVersion("4", false);
```
ivannaranjo: The version really is "4.5.x" or "4.6.x" depending on the version of VS and the installed version of the .NET Framework.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            return new List<AspNetVersion>
+            {
+                AspNetCore1,
+                AspNetCore2
+            };
+        }
+
+        /// <summary>
+        /// The version number of ASP.NET. This corrisponds to the version of .NET Core used as well.
```
ivannaranjo: corrsponds => corresponds

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// Human readable version of this object.
+        /// </summary>
+        /// <returns>A localized formatted version of this ASP.NET version</returns>
+        public override string ToString()
+        {
+            string nameFormat = IsCore ? Resources.AspNetCoreVersionedName : Resources.AspNetVersionedName;
+            return string.Format(nameFormat, Version);
+        }
+
+        public static IList<AspNetVersion> GetAvailableVersions(string vsVersion, FrameworkType framework)
```
ivannaranjo: Method comment missing.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                case VsVersionUtils.VisualStudio2017Version:
+                    IList<AspNetVersion> versions = GetVs2017AspNetCoreVersions();
+                    if (framework == FrameworkType.NetFramework)
+                    {
+                        versions.Add(AspNet4);
+                    }
+                    return versions;
+                default:
+                    throw new InvalidOperationException(
```
ivannaranjo: We should limit throwing in case of unknown VS, this can cause us to crash when Microsoft releases VS 2019 (or whatever). We need to re-think how we are _forward_ compatible.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/FrameworkType.cs
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using GoogleCloudExtension.Utils;
+
+namespace GoogleCloudExtension.TemplateWizards.Dialogs.TemplateChooserDialog
+{
+
```
ivannaranjo: Extra line.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/FrameworkType.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// An object that describes the type of framework.
+    /// </summary>
+    public enum FrameworkType
+    {
+        /// <summary>
+        /// The enum value for .NET Core.
+        /// </summary>
+        [LocalizedDisplayName(nameof(Resources.FrameworkTypeDisplayNameNetCore))]
```
ivannaranjo: Why would an enum have a localized display name? We should never just convert enum values to string, we should choose the string for them ourselves. This is an implementation details that should never emerge to the user.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModelResult.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <summary>
+        /// The type of application to create.
+        /// </summary>
+        public string AppType
+        {
+            get
+            {
+                if (_isMvc)
+                {
+                    return "MVC";
```
ivannaranjo: These should be localized strings.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModelResult.cs
```diff
@@ -0,0 +1,75 @@
...
+                }
+                else
+                {
+                    throw new InvalidOperationException("Result should be either MVC or WebAPI.");
+                }
+            }
+        }
+
+        /// <param name="templateChooserViewModel">The view model this results object will pull its data from.</param>
+        public TemplateChooserViewModelResult(TemplateChooserViewModel templateChooserViewModel)
```
ivannaranjo: Method comment seems suspect.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModelResult.cs
```diff
@@ -0,0 +1,75 @@
...
+        private readonly bool _isWebApi;
+
+        /// <summary>
+        /// The type of application to create.
+        /// </summary>
+        public string AppType
+        {
+            get
+            {
+                if (_isMvc)
```
ivannaranjo: Seems like this should be an enum instead of booleans, since the type of template is mutually exclusive.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserWindowContent.xaml
```diff
@@ -0,0 +1,83 @@
...
+             mc:Ignorable="d"
+             d:DataContext="{d:DesignInstance local:TemplateChooserViewModel}">
+    <UserControl.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="../../../Theming/DefaultStylesResource.xaml"/>
+                <ResourceDictionary Source="../../../Theming/ThemingResources.xaml"/>
+            </ResourceDictionary.MergedDictionaries>
+            <DataTemplate DataType="{x:Type local:FrameworkType}">
+                <TextBlock Text="{Binding Converter={utils:EnumDisplayNameConverter}}"/>
```
ivannaranjo: Too much magic, just have this in the view model, which is the class responsible for talking to the view. The enums should always be view agnostic, including not having the localized string attached to them.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
```diff
@@ -0,0 +1,169 @@
...
+                }
+                if (result == null)
+                {
+                    throw new WizardBackoutException();
+                }
+                string thisTemplateDirectory = Path.GetDirectoryName((string)customParams[0]);
+
+                string version = result.SelectedVersion.Version;
+                string templatePath = Path.Combine(
+                    thisTemplateDirectory, "..", "..", "..", result.AppType, "1033", version, $"{version}.vstemplate");
```
ivannaranjo: I applaud the effort, but the code is Windows specific. You can specify a format string that will create the right path without using this. Is ok to hard code the '\' for file paths on Windows specific code.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <inheritdoc/>
+        public void BeforeOpeningFile(ProjectItem projectItem)
+        {
+            throw new NotImplementedException("This wizard should delegate to another template and cancel itself.");
+        }
+
+        /// <inheritdoc/>
+        public void RunFinished()
+        {
+            throw new NotImplementedException("This wizard should delegate to another wizard and cancel itself.");
```
ivannaranjo: This tells me that this PR should go against a separate feature branch, not the master branch, as we should not be committing unfinished features there.

GoogleCloudExtension/GoogleCloudExtension/Utils/EnumDisplayNameConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+using System;
+using System.ComponentModel;
+using System.Globalization;
+using System.Reflection;
+using System.Windows.Data;
+using System.Windows.Markup;
+
+namespace GoogleCloudExtension.Utils
+{
+    public class EnumDisplayNameConverter : MarkupExtension, IValueConverter
```
ivannaranjo: Please don't use this, enums should not have a localized display string associated with them.

GoogleCloudExtension/GoogleCloudExtension.Utils/PathUtils.cs
ILMTitan: I don't use it above, because I want to convert both types of directory separators to '/' so `Uri` will treat them as expected. I use it here because I am converting back to the native format.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: This is not a function I wrote. This is a file being move into a different assembly.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: Also not my code.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: Also not my code.

ILMTitan: Also not my code.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: Also not my code.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: Also not my code.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: Also not my code.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: Also not my code.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ILMTitan: Also not my code.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AspNetVersion.cs
ILMTitan: Changing default case to match the VS2017 case.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/FrameworkType.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.TemplateWizards/DelegatingTemplateWizard.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
ILMTitan: When I first tried to reference a class in GoogleCloudExtension.dll from a vstemplate file, I got an error popup about being unable to reference a class from an MEF assembly (which GoogleCloudExtension.dll, containing the package, is).

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModelResult.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModelResult.cs
ILMTitan: This actually relates to the directory of the target template and is never presented to the user. Regardless, it is not encoded as an Enum.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
ILMTitan: Changes this to use `Path.DirectoryName` multiple times.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/FrameworkType.cs
ILMTitan: I agree a user should never see an enum. Are you saying that a ViewModel should never have an enum? That we need to manually convert from strings to enums and back in the ViewModel? That seems like a lot of fragile code for not a lot of gain.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
ILMTitan: This is not unfinished. This method should never be called.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModelResult.cs
ILMTitan: Tried to make this comment more clear by adding to the class comment.

GoogleCloudExtension/GoogleCloudExtension.TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
ivannaranjo: Please on an issue so we can come back to this again later and fix it, if you think that this can be eventually fixed.

GoogleCloudExtension/GoogleCloudExtension.Utils/PathUtils.cs
ivannaranjo: Hmm... Perhaps a comment on what you are doing would make this code clearer to read. Also, our code for VS is mostly Windows specific (i.e. not portable) so we can always assume that \ will be the path separator.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Please open an issue so we can track fixing this code then.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Please open an issue so we can track fixing this code then.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Please open an issue so we can track fixing this code then.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Please open an issue so we can track fixing this code then.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Please open an issue so we can track fixing this code then.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Please open an issue so we can track fixing this code then.

GoogleCloudExtension/GoogleCloudExtension.Utils/StringUtils.cs
ivannaranjo: Please open an issue so we can track fixing this code then.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModelResult.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        /// <summary>
+        /// The Google Cloud Project ID.
+        /// </summary>
+        public string GcpProjectId { get; }
+
+        /// <summary>
+        /// The selected framework type.
+        /// </summary>
+        public FrameworkType SelectedFramework { get; }
```
ivannaranjo: I'm now even more convinced that we should not have localized strings associated with the enums like this. We should have a small class that associates the enum with the localized string, including one with the default value if necessary.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModel.cs
```diff
@@ -0,0 +1,160 @@
...
+
+namespace GoogleCloudExtension.TemplateWizards.Dialogs.TemplateChooserDialog
+{
+    /// <summary>
+    /// View Model for the Template Chooser dialog.
+    /// </summary>
+    public class TemplateChooserViewModel : ViewModelBase
+    {
+        private string _gcpProjectId;
+        private FrameworkType _selectedFramework = (FrameworkType)(-1);
```
ivannaranjo: You should instead define an enum value of `None` (which should be set to 0 btw) to indicate that there's no framework selected. This is one of the reasons why having magic to associate strings with enums does not work.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/AppType.cs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+namespace GoogleCloudExtension.TemplateWizards.Dialogs.TemplateChooserDialog
+{
+    /// <summary>
+    /// The types of applications we have templates for.
+    /// </summary>
+    public enum AppType
+    {
+        None,
```
ivannaranjo: Nit, have a small comment for each option, helps when using it in intellisense. They are self-explanatory I know, but it will help in the long run.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserViewModel.cs
```diff
@@ -25,12 +25,12 @@ namespace GoogleCloudExtension.TemplateWizards.Dialogs.TemplateChooserDialog
     /// </summary>
     public class TemplateChooserViewModel : ViewModelBase
     {
+
         private string _gcpProjectId;
-        private FrameworkType _selectedFramework = (FrameworkType)(-1);
+        private FrameworkType _selectedFramework = FrameworkType.None;
```
ivannaranjo: Great, thank you! :)

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserWindowContent.xaml
```diff
@@ -71,11 +79,11 @@
                     <RadioButton
                         GroupName="AppType"
                         Content="{x:Static ext:Resources.WizardTemplateChooserMvcLabel}"
-                        IsChecked="{Binding IsMvc}"/>
+                        IsChecked="{Binding AppType, Converter={utils:RadioEnumConverter Target={x:Static local:AppType.Mvc}}}"/>
```
ivannaranjo: I have to ask, what is the advantage of this vs. what you had before? Since the set of choices is well known, and limited, why having the extra complexity that your solution adds? If we were to have 10s of choices I would agree, but we only have 2 choices, and I don't think that we will ever have more than 4. Having it hardcoded in the UI is perfectly valid.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserWindowContent.xaml
```diff
@@ -16,7 +16,15 @@
                 <ResourceDictionary Source="../../../Theming/ThemingResources.xaml"/>
             </ResourceDictionary.MergedDictionaries>
             <DataTemplate DataType="{x:Type local:FrameworkType}">
-                <TextBlock Text="{Binding Converter={utils:EnumDisplayNameConverter}}"/>
+                <TextBlock x:Name="TextDisplay"/>
```
ivannaranjo: Wouldn't it be simpler to just have two entries under the `ComboBox` with the two options? I'm sorry but this solution is waaay more complex than what it really needs to be. You can even use the [`Tag`](https://msdn.microsoft.com/en-us/library/system.windows.frameworkelement.tag(v=vs.110).aspx) property in the TextBlock to contain the value of the enum that you want and just add two `TextBlock` instances.

GoogleCloudExtension/ProjectTemplates/GcpProjectTemplate/Core/WebAPI/2.0.csproj
```diff
@@ -16,7 +16,10 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Cloud.Diagnostics.AspNetCore" Version="1.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0" />
```
ivannaranjo: Looking at a 2.0 project generated by dotnet, it is much much simpler, it collapses all of these packages into:
```xml
<ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
  </ItemGroup>
```

We should make sure that the generated .csproj files are idiomatic.

GoogleCloudExtension/GoogleCloudExtension.Utils/PathUtils.cs
ILMTitan: Added comments.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserWindowContent.xaml
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/TemplateChooserDialog/TemplateChooserWindowContent.xaml
ILMTitan: Changed this back to have IsMvc/IsWebApi properties on the ViewModel.

GoogleCloudExtension/ProjectTemplates/GcpProjectTemplate/Core/WebAPI/2.0.csproj
ILMTitan: Done.

PR #799, opened 2017-09-15
Remove the creation of a complex cloudbuild.yaml when a simpler Dockerfile suffices.

### LINKED ISSUE #591: Remove the code that generate cloud builder build files for app containers.
Currently the extension generates a very simple build file to build the app's container. This is not necessary as the gcloud CLI actually contains a command to build Docker images directly on the cloud builder when no complex configuration is required. The extension should call it like so:
```bash
gcloud container builds submit --tag={tag to use} {path to too of app's files}
```

We should only be using a build file if we intend to do more complex transformations during the build.

ivannaranjo:
When we build a Docker image to be deployed to GKE we create a `cloudbuild.yaml` file that only builds the image by invoking `docker build` on the existing `Dockerfile`. As it turns out `gcloud` already supports this directly. Instead of generating the `cloudbuild.yaml` we can invoke `gcloud` like this:
```bash
gcloud container builds submit --tag={name of the tag} {path to the published app}
```
This will accomplish the same, building the Docker image for the app, with less complexity.

Fixes #591

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Deployment/GkeDeployment.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Deployment/GkeDeployment.cs
@@ -127,14 +127,17 @@ namespace GoogleCloudExtension.Deployment
                 }
 
                 NetCoreAppUtils.CopyOrCreateDockerfile(project, appRootPath);
-                var image = CloudBuilderUtils.CreateBuildFile(
+                var imageTag = CloudBuilderUtils.GetImageTag(
                     project: options.GCloudContext.ProjectId,
                     imageName: options.DeploymentName,
-                    imageVersion: options.DeploymentVersion,
-                    buildFilePath: buildFilePath);
+                    imageVersion: options.DeploymentVersion);
 
                 if (!await ProgressHelper.UpdateProgress(
-                    GCloudWrapper.BuildContainerAsync(buildFilePath, appRootPath, outputAction, options.GCloudContext),
+                    GCloudWrapper.BuildContainerAsync(
+                        imageTag: imageTag,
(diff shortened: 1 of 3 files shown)


GoogleCloudExtension/GoogleCloudExtension.Deployment/GkeDeployment.cs
```diff
@@ -114,27 +114,27 @@ public class DeploymentOptions
...
 
-                NetCoreAppUtils.CopyOrCreateDockerfile(project, appRootPath);
-                var image = CloudBuilderUtils.CreateBuildFile(
+                NetCoreAppUtils.CopyOrCreateDockerfile(project, stageDirectory);
+                var imageTag = CloudBuilderUtils.GetImageTag(
                     project: options.GCloudContext.ProjectId,
                     imageName: options.DeploymentName,
-                    imageVersion: options.DeploymentVersion,
-                    buildFilePath: buildFilePath);
+                    imageVersion: options.DeploymentVersion);
```
ILMTitan: Since every parameter to `GetImageTag()` is found in `options`, would it be better to move the function into class `DeploymentOptions` itself?

    string imageTag = options.ImageTag;

or 
   
    string imageTag = options.GetImageTag();

That would also let you get rid of class `CloudBuilderUtils` entirely.

ivannaranjo: Actually I want to keep it separate because that way creating a Docker image is not only related to GKE. I am in fact working on a private change at the moment where I need this code shared as it is, because I am just building docker images out of the project.

ivannaranjo:
PTAL

PR #801, opened 2017-09-15
Modify the templates so they can be built in place.

### LINKED ISSUE #796: Ensure that we don't send diagnostics information during development by default in templates.
Curent .NET Core templates do setup logging and add the stackdriver logger provider for all environments, we should ensure that this only happens for production environments. The development environment should not be sending diagnostics by default.

ILMTitan:
This allows us to edit them in Visual Studio.
This PR also fixes #796.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
@@ -49,7 +49,6 @@ namespace GoogleCloudExtension.TemplateWizards
         {
             try
             {
-
                 string projectName = replacements[ReplacementsKeys.ProjectNameKey];
 
                 TemplateChooserViewModelResult result;
@@ -72,6 +71,7 @@ namespace GoogleCloudExtension.TemplateWizards
                         Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName((string)customParams[0]))));
 
 
+                var serviceProvider = automationObject as IServiceProvider;
                 object[] newCustomParams = GetNewCustomParams(replacements, customParams.Skip(1), result);
 
                 string version = result.SelectedVersion.Version;
@@ -79,7 +79,6 @@ namespace GoogleCloudExtension.TemplateWizards
                     thisTemplateDirectory, result.AppType.ToString(), "1033", version, $"{version}.vstemplate");
                 string destinationFolder = replacements[ReplacementsKeys.DestinationDirectoryKey];
 
-                var serviceProvider = automationObject as IServiceProvider;
                 var vsSolution = (IVsSolution6)serviceProvider.QueryService<SVsSolution>();
                 IVsHierarchy newProject;
                 Marshal.ThrowExceptionForHR(
@@ -128,7 +127,7 @@ namespace GoogleCloudExtension.TemplateWizards
         }
 
         private static object[] GetNewCustomParams(
-            Dictionary<string, string> replacements,
+            IReadOnlyDictionary<string, string> replacements,
             IEnumerable customParams,
             TemplateChooserViewModelResult result)
         {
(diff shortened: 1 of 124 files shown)


GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
```diff
@@ -128,35 +127,43 @@ public void RunFinished()
...
+                switch (result.SelectedFramework)
+                {
+                    case FrameworkType.NetFramework:
+                        var targetFramework = "net461";
+                        additionalCustomParams.Add($"netcoreapp{result.SelectedVersion.Version}={targetFramework}");
+                        break;
+                    case FrameworkType.NetCore:
+                        // Do nothing.
+                        break;
+                    case FrameworkType.None:
```
ivannaranjo: These two cases, `FrameworkType.None` and `default`, seem like the same case.

GoogleCloudExtension/ProjectTemplates/GcpProjectTemplate/Core/MVC/1.0.csproj
```diff
@@ -1,11 +1,11 @@
...
 
   <PropertyGroup>
-    <TargetFramework>$targetframework$</TargetFramework>
+    <TargetFramework>netcoreapp1.0</TargetFramework>
     <PreserveCompilationContext>true</PreserveCompilationContext>
-    <AssemblyName>$safeprojectname$</AssemblyName>
+    <AssemblyName>SafeProjectName</AssemblyName>
     <OutputType>Exe</OutputType>
-    <PackageId>$safeprojectname$</PackageId>
+    <PackageId>SafeProjectName</PackageId>
```
ivannaranjo: Doesn't this hardcode the value of `<PackageId>` to `"SafeProjectName"`? Why removing the placeholders?

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateSelectorWizard.cs
ILMTitan: They are different enough that I wanted to provide two different error messages.

GoogleCloudExtension/ProjectTemplates/GcpProjectTemplate/Core/MVC/1.0.csproj
ILMTitan: `$safeprojectname$` is a default replacement key, and all default replacement keys are of the form `$keyname$`.  That does not *have* to be true, however. By adding a new replacement key `SafeProjectName` during the `RunStarted()` method of the wizard with the same value, the text `SafeProjectName` gets replaced the same way as `$safeprojectname$` does. The big benefit is that `SafeProjectName` is a legal namespace, allowing us to compile and test our templates in place, without needing to run an experimental instance of visual studio.

ILMTitan:
> How are you replacing the SafeProjectName with the value of $safeprojectid$ when instantiating the templates?

In GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/GoogleProjectTemplateWizard.

    replacements[ReplacementsKeys.EmbeddableSafeProjectNameKey] =
     replacements[ReplacementsKeys.SafeProjectNameKey];

where 

    public const string SafeProjectNameKey = "$safeprojectname$";
    public const string EmbeddableSafeProjectNameKey = "SafeProjectName";

a new replacement is added to the replacements dictionary. The key is `SafeProjectName`, and the value comes from the value of the key `$safeprojectname$`

ILMTitan:
PTAL

GoogleCloudExtension/ProjectTemplates/GcpProjectTemplate/Core/MVC/1.0.csproj
ivannaranjo: I see.. Doesn't it also mean that we loose track of which ones are placeholders vs. which symbols are final symbols that should appear on the final template? The mistake I made seems too dangerious vs. the advantage of building the template. I would rather expand the template and build it no?

ivannaranjo:
I'm questioning the basic principle, given the fact that it is very dangerous to have a typo, or even just look through the sources looking for the symbols that are going to be replaced. They look like every other symbol.

Is there away in which we could expand a template on our own while running the tests so we can verify that the templates build instead of this?

ivannaranjo:
Alternatively, if we use a fixed naming pattern that is easy to spot, but still valid C# I would go with that, like `_PlaceHolder_` or something along those lines?

ILMTitan:
I am going to close this for now. I am running into some problems running and building the 1.0 template because of extraneous files that are added when used in place.

ILMTitan:
Actually, I don't care if they compile and run; that is what the ProjectTemplates.Tests project is for. The big win for this is being able to edit files with IntelliSense support.

ILMTitan:
Agreed. Changed the replacement key to `_safe_project_name_`.

ILMTitan:
PTAL

PR #805, opened 2017-09-21
Changes suggested by the strings review.

ILMTitan:


GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdWindow.cs
```diff
@@ -32,12 +32,13 @@ private PickProjectIdWindow() : base(GoogleCloudExtension.Resources.WizardPickPr
         /// <summary>
         /// Initalizes the Pick Project Window and waits for it to finish.
         /// </summary>
+        /// <param name="projectName"></param>
```
ivannaranjo: What do you mean by `projectName`, add the comment for this. It is easy to confuse the name with the ID, so we need to be explicit.

ILMTitan: Fixed.

PR #811, opened 2017-09-29
Master removing beta dependency

ivannaranjo:
Removing the dependency on the `beta` component from `gcloud`. This PR also refactors a bit how the check for components is done, instead of using the string with the name we will now use an `enum` which will check that we're using a valid component name.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudComponent.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudComponent.cs
@@ -1,10 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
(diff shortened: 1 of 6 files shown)


GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudComponent.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Placeholder for no component.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// The beta component, contains the beta features for gcloud, only depend on this if
+        /// absolutely necessary as things change rapidly.
+        /// </summary>
+
```
ILMTitan: Extra line.

PR #815, opened 2017-10-02
Using the FQDN of a GCE instance instead of the IP address.

### LINKED ISSUE #790: Use FQDN for GCE VM instances using the public IP.
Use the FQDN instead of just an IP address when connecting to a GCE VM, this will mean generating the FQDN from the IP as follows:
For the IP address 1.2.3.4 the FQDN will be 4.3.2.1.bc.googleusercontent.com

This will help with connecting to the VMs from private networks where the use of the Remote Desktop client is restricted.

ivannaranjo:
This PR changes the way we refer to the machine when opening a terminal server session, by using the FQDN of the GCE instance instead of the bare IP address.

Fixes #790

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.DataSources/InstanceExtensions.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.DataSources/InstanceExtensions.cs
@@ -140,6 +140,19 @@ namespace GoogleCloudExtension.DataSources
             instance.NetworkInterfaces?.FirstOrDefault()?.AccessConfigs?.FirstOrDefault()?.NatIP;
 
         /// <summary>
+        /// Returns the FQDN for an instance in GCE.
+        /// </summary>
+        /// <param name="instance">The instance to inspect.</param>
+        /// <returns></returns>
+        public static string GetFullyQualifiedDomainName(this Instance instance)
+        {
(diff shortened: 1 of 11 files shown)


GoogleCloudExtension/GoogleCloudExtension.DataSources/InstanceExtensions.cs
```diff
@@ -140,6 +140,23 @@ public static WindowsInstanceInfo GetWindowsInstanceInfo(this Instance instance)
...
+        {
+            var ipAddress = instance.GetPublicIpAddress();
+            if (ipAddress == null)
+            {
+                return "";
+            }
+
+            var reverseIp = string.Join(".", ipAddress.Split('.').Reverse());
+            return $"{reverseIp}.bc.googleusercontent.com";
+        }
```
ILMTitan: Create unit tests for this function? You should really only need three cases (null instance, null ipAddress, proper ipAddress).

ivannaranjo: Ok, will see what I can do.

ivannaranjo:
Didn't add a case for the `null` instance because I don't support that case.

PTAL.

PR #818, opened 2017-10-02
Correctly check for .NET Core before creating config files.

### LINKED ISSUE #816: The option to create an app.yaml and Dockerfile is enabled for an ASP.NET 4.x project.
It seems that creating the project from the built-in template is fooling the project detection.

ivannaranjo:
This PR uses the right check to enable/disable the menu item for creating config files to ASP.NET Core apps. The check has been centralized into an extension method so it can be used anywhere we need to check for the ASP.NET Core framework, like the publishing wizard.

Fixed #816

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.Deployment/IParsedProjectExtensions.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.Deployment/IParsedProjectExtensions.cs
@@ -1,4 +1,18 @@
-﻿using System;
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


GoogleCloudExtension/GoogleCloudExtension.Deployment/IParsedProjectExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// Determines if the given project is a .NET Core project or not.
+        /// </summary>
+        /// <param name="project">The project to check.</param>
+        /// <returns>Returns true if the project is a .NET Core project, false otherwise.</returns>
+        public static bool IsAspNetCoreProject(this IParsedProject project)
+            => project.ProjectType == KnownProjectTypes.NetCoreWebApplication1_0 ||
+               project.ProjectType == KnownProjectTypes.NetCoreWebApplication1_1 ||
+               project.ProjectType == KnownProjectTypes.NetCoreWebApplication2_0;
```
ILMTitan: Create unit tests for this function? There should be 6 (null project, null ProjectType, invalid ProjectType, and the 3 success cases).

ivannaranjo: This function is trivial, the tests are really unnecessary since they will just duplicate the function itself. There's really no logic.

PR #820, opened 2017-10-03
Add GCP Project Id Text to publish dialogs.

### LINKED ISSUE #791: Publish to appengine flex does not show the name of the target project.
Created a new Google ASP.Net Core Web API project.
Selected my account and my test project (note that first project in the list is a production project).
Ensured that selected project ID also appears in "Enter project ID".
Ensured that appsettings.json contains the correct project ID
Right click project in solution explorer -> Publish to Google Cloud...
Select Appengine Flex
Leave settings unchanged, click Publish

Project compiles successfully

Observe Output window, notice that wrong target project is used.

descriptor:      [C:\Users\xxx\AppData\Local\Temp\c1uamfsa.zjr\app.yaml]
source:          [C:\Users\xxx\AppData\Local\Temp\c1uamfsa.zjr]
target project:  [REDACTED - WRONG project]
target service:  [default]
target version:  [20170904t184514]

Publish fails because I don't have write access to the production project.

ILMTitan:
Fixes #791.

Adds *Google Cloud Project ID* `TextBox` and *Select Project* `Button` to Publish dialogs.

Change project select dialog to change the current project (both for template dialogs and for publish dialogs).

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/PublishDialog/PublishDialogStepBase.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/PublishDialog/PublishDialogStepBase.cs
@@ -13,8 +13,6 @@
 // limitations under the License.
 
 using GoogleCloudExtension.Accounts;
-using GoogleCloudExtension.TemplateWizards.Dialogs.ProjectIdDialog;
-using GoogleCloudExtension.Utils;
 using GoogleCloudExtension.Utils.Validation;
 using System;
 using System.Windows;
@@ -29,8 +27,6 @@ namespace GoogleCloudExtension.PublishDialog
     {
         private bool _canGoNext;
         private bool _canPublish;
-        internal Func<string, string> PickProjectPrompt = PickProjectIdWindow.PromptUser;
-        public IPublishDialog PublishDialog { get; private set; }
 
         public bool CanGoNext
         {
@@ -60,17 +56,10 @@ namespace GoogleCloudExtension.PublishDialog
 
         public abstract FrameworkElement Content { get; }
         public string GcpProjectId => CredentialsStore.Default.CurrentProjectId;
-        public ProtectedCommand SelectProjectCommand { get; }
 
         public event EventHandler CanGoNextChanged;
 
         public event EventHandler CanPublishChanged;
-
-        /// <inheritdoc />
-        protected PublishDialogStepBase()
-        {
-            SelectProjectCommand = new ProtectedCommand(OnSelectProjectCommand);
-        }
 
         public virtual IPublishDialogStep Next()
         {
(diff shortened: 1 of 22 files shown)


ivannaranjo:
Would you be interested on parrying down this change to only have the display portions of the project where the deployment is going to happen? That is so we can get the bug fixed and we can have the full discussion on the "current project" model as a separate change?

ILMTitan:
Done. Please take another look.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/FlexStep/FlexStepContent.xaml
```diff
@@ -18,6 +18,18 @@
     </UserControl.Resources>
 
     <StackPanel>
+        <DockPanel HorizontalAlignment="Left">
```
ivannaranjo: Nit, since these elements are already inside of the `StackPanel` what is the advantage of the nested `DockPanel`?

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GceStep/GceStepContent.xaml
```diff
@@ -17,6 +17,18 @@
     </UserControl.Resources>
 
     <StackPanel>
+        <DockPanel HorizontalAlignment="Left">
```
ivannaranjo: Same, question.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/GkeStep/GkeStepContent.xaml
```diff
@@ -17,6 +17,18 @@
     </UserControl.Resources>
 
     <StackPanel>
+        <DockPanel HorizontalAlignment="Left">
```
ivannaranjo: Same question with the `DockPanel`.

GoogleCloudExtension/GoogleCloudExtension/TemplateWizards/Dialogs/ProjectIdDialog/PickProjectIdViewModel.cs
```diff
@@ -169,6 +169,7 @@ private void OnChangeUser()
 
         private void OnOk()
         {
+            CredentialsStore.Default.UpdateCurrentProject(SelectedProject);
```
ivannaranjo: This change doesn't seem relevant to just display the current project.

ivannaranjo: We should not be changing the current project in the selection of the project in the template, we need to find a coherent model for this.

GoogleCloudExtension/GoogleCloudExtension/PublishDialog/PublishDialogStepBase.cs
```diff
@@ -54,6 +55,7 @@ protected set
         }
 
         public abstract FrameworkElement Content { get; }
+        public string GcpProjectId => CredentialsStore.Default.CurrentProjectId;
```
ivannaranjo: Nit, add a comment for the new public property.

GoogleCloudExtension/GoogleCloudExtension/PublishDialogSteps/FlexStep/FlexStepContent.xaml
ILMTitan: It was an artifact from when there was also a button to the right of the text box. Now that the button is back, it is needed again.

PR #843, opened 2017-11-07
Enable csr api

Deren-Liao:
Let user enable Cloud Source Repositories API if not enabled yet.


GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -65,10 +66,26 @@ public Project SelectedProject
...
+                if (_selectedProject == null)
                 {
-                    ErrorHandlerUtils.HandleAsyncExceptions(() =>
-                        RepositoriesAsync.StartListRepoTaskAsync(_selectedProject.ProjectId));
+                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(
+                        () => RepositoriesAsync.StartListRepoTaskAsync(null)));
+                }
+                else if (oldValue != _selectedProject)
+                {
+                    List<string> requiredApis = new List<string>() { KnownApis.CloudSourceRepositoryApiName };
```
ILMTitan: This can be `var`. You can also get rid of the `()`.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -65,10 +66,26 @@ public Project SelectedProject
...
+                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(async () =>
```
ILMTitan: Remove the whitespace at the end of the line.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -65,10 +66,26 @@ public Project SelectedProject
...
+                }
+                else if (oldValue != _selectedProject)
+                {
+                    List<string> requiredApis = new List<string>() { KnownApis.CloudSourceRepositoryApiName };
+                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(async () => 
+                    {
+                        ApiManager apiManager = ApiManager.GetApiManager(_selectedProject.ProjectId);
+                        if (!await apiManager.EnsureAllServicesEnabledAsync(
+                                requiredApis, 
+                                String.Format(Resources.CsrEnableApiMessageFormat, _selectedProject.Name)))
```
ILMTitan: `String` => `string`

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -65,10 +66,26 @@ public Project SelectedProject
...
+                        () => RepositoriesAsync.StartListRepoTaskAsync(null)));
+                }
+                else if (oldValue != _selectedProject)
+                {
+                    List<string> requiredApis = new List<string>() { KnownApis.CloudSourceRepositoryApiName };
+                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(async () => 
+                    {
+                        ApiManager apiManager = ApiManager.GetApiManager(_selectedProject.ProjectId);
+                        if (!await apiManager.EnsureAllServicesEnabledAsync(
+                                requiredApis,
```
ILMTitan: Remove whitespace.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -65,10 +66,26 @@ public Project SelectedProject
...
+                    List<string> requiredApis = new List<string>() { KnownApis.CloudSourceRepositoryApiName };
+                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(async () => 
+                    {
+                        ApiManager apiManager = ApiManager.GetApiManager(_selectedProject.ProjectId);
+                        if (!await apiManager.EnsureAllServicesEnabledAsync(
+                                requiredApis, 
+                                String.Format(Resources.CsrEnableApiMessageFormat, _selectedProject.Name)))
+                        {
+                            SelectedProject = null;
+                            return;
```
ILMTitan: I tend to prefer `if(x) { y(); } else { z(); }` to `if(!x) { z(); return; } y();`, especially in cases like this where the `return` is returning a lambda expression. I tend to use the latter only when checking preconditions.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2903,4 +2903,8 @@
     <value>Select Google Cloud Project to Publish to - {0}</value>
     <comment>The title of the Select Google Cloud Project dialog created from the Publish dialogs. The parameter is the name of the Visual Studio project.</comment>
   </data>
+  <data name="CsrEnableApiMessageFormat" xml:space="preserve">
+    <value>To proceed, you need to enable Cloud Source Repositories API for the chosen project &lt;{0}&gt;. Otherwise, Click Cancel button and choose a different project. Do you want to enable the necessary services for project &lt;{0}&gt; now?</value>
```
ILMTitan: I am not sure about the angle brackets around the project id. I would either change them to double quotes or just remove them entirely.

I feel like this entire line is to verbose. The popup is preventing the user from proceeding, so the `To proceed` is unneeded. Users should know what a cancel button does. The project id is repeated, which shouldn't be necessary. The user just chose the project, so you don't need to tell them they chose it.

I would go with something like this:

    The required Cloud Source Repositories APIs for project {0} are not enabled. Do you want to enable them now?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -65,10 +66,26 @@ public Project SelectedProject
             {
                 var oldValue = _selectedProject;
                 SetValueAndRaise(ref _selectedProject, value);
-                if (_selectedProject != null && IsReady && oldValue != _selectedProject)
+                if (_selectedProject == null)
                 {
-                    ErrorHandlerUtils.HandleAsyncExceptions(() =>
-                        RepositoriesAsync.StartListRepoTaskAsync(_selectedProject.ProjectId));
+                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(
+                        () => RepositoriesAsync.StartListRepoTaskAsync(null)));
```
ILMTitan: Instead of calling `StartListRepoTaskAsync(null)` why not just create a new function `AsyncRepositories.ClearTask()`? A new method would also let you keep `StartListRepoTaskAsync()` simple.

GoogleCloudExtension/GoogleCloudExtension/ApiManagement/ApiManager.cs
```diff
@@ -38,12 +38,28 @@ public class ApiManager
         /// </summary>
         public static ApiManager Default => s_defaultManager.Value;
 
-        private Lazy<ServiceManagementDataSource> _dataSource = new Lazy<ServiceManagementDataSource>(CreateDataSource);
+        private Lazy<ServiceManagementDataSource> _dataSource;
 
         private ApiManager()
         {
             CredentialsStore.Default.CurrentAccountChanged += OnCurrentCredentialsChanged;
```
ivannaranjo: If the project Id is being specified then don't listen for changes on `CurrentAccountChanged` or `Reset`, since the manager is now temporary.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -62,9 +62,16 @@ public DisplayOptions DisplayState
         public async Task StartListRepoTaskAsync(string projectId)
         {
             Debug.WriteLine(nameof(StartListRepoTaskAsync));
-            _latestTask = GetCloudReposAsync(projectId);
-            RaiseAllPropertyChanged();
-            await _latestTask;
+            if (String.IsNullOrWhiteSpace(projectId))
```
ivannaranjo: Is there a good reason why `projectId` would be `null`?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
ivannaranjo: Agreed, don't change the `null` expectation.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
ivannaranjo: This should actually be a `static readonly` value in the class. This list of required APIs will not change at all.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
ivannaranjo: I would use the same string as the other dialogs, that way later on it should be possible to modify the strings easily and have a consistent experience.

GoogleCloudExtension/GoogleCloudExtension/ApiManagement/ApiManager.cs
ILMTitan: That is what the code is doing now. This constructor is for the static `ApiManager`, while the other is for the temporary one associated with a specific projectId. This one should still listen for changes, and the other does not.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
ILMTitan: Are you saying we should have a standardized message that takes the name of the feature (i.e. "Cloud Source Repositories") as a parameter? The name of the feature would also have to be a resource string.

Deren-Liao: For this message box, highlighting the project id is important.  Most other features of our extension binds to the selected project in Cloud Explorer.  Enabling an API means to enable it for the project.  CSR feature here the project can be different.  Without showing the project name here, it's easy to "confuse/mislead" the user.

GoogleCloudExtension/GoogleCloudExtension/ApiManagement/ApiManager.cs
Deren-Liao: As Jim mentioned, the ApiManager is used only by the Default ApiManager.

Deren-Liao:
Updated.  Please take a look again.

GoogleCloudExtension/GoogleCloudExtension/ApiManagement/ApiManager.cs
```diff
@@ -29,7 +29,7 @@ namespace GoogleCloudExtension.ApiManagement
     /// that manages its own <seealso cref="ServiceManagementDataSource"/> instance. This class will update itself when the user
     /// changes the current project/user.
     /// </summary>
-    public class ApiManager
+    public class ApiManager : IApiManager
```
ivannaranjo: This makes a lot of sense for testing, thank you.

GoogleCloudExtension/GoogleCloudExtension/ApiManagement/IApiManager.cs
```diff
@@ -0,0 +1,48 @@
...
+        Task<bool> AreServicesEnabledAsync(IList<string> serviceNames);
+
+        /// <summary>
+        /// This method will check that all given services are enabled and if not will prompt the user to enable the
+        /// necessary services.
+        /// </summary>
+        /// <param name="serviceNames">The services to check.</param>
+        /// <param name="prompt">The prompt to use in the prompt dialog to ask the user for permission to enable the services.</param>
+        /// <returns>A task that will be true if all services where enabled, false if the user cancelled or if the operation failed.</returns>
+        Task<bool> EnsureAllServicesEnabledAsync(IEnumerable<string> serviceNames, string prompt);
```
ivannaranjo: Note, in the work I am doing for the App Engine app management I am actually not using the `EnsureAllServicesEnabledAsync()` method anymore. I am showing the message inline in the dialog to avoid bombarding the users with dialogs. We should look at doing something similar in the UI for the CSR.

Ideally the UI will just have a placeholder to show the user that the API needs to be enabled, then the user can enable the API to list the clusters.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindow.cs
```diff
@@ -49,6 +49,11 @@ private CsrCloneWindow(IList<Project> projects) : base(StringResources.CsrCloneW
         }
 
         /// <summary>
+        /// Only for unit test, don't call it in production code.
+        /// </summary>
+        internal CsrCloneWindow() : base(StringResources.CsrCloneWindowTitle) { }
```
ivannaranjo: Interesting, normally you wouldn't need to test the Window. In my tests I just pass null for the owner of a view model, and change the view model to deal with a null one. Perhaps you should also allow the injection of the Window into the view model instead of instantiating the window during your tests.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        private TaskCompletionSource<IList<Repo>> _testTaskCompletionSource;
+        private Mock<Func<string, Task<IList<Repo>>>> _getCloudReposMock;
+        private Mock<IApiManager> _apiManagerMock;
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            _apiManagerMock = new Mock<IApiManager>();
+            _testProjectMock = Mock.Of<Project>();
+            _cloneWindowMock = new Mock<CsrCloneWindow>();
```
ivannaranjo: We should have an interface here I think, typically the Window interaction with the view model is very narrow.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -55,23 +55,25 @@ public DisplayOptions DisplayState
         }
 
         /// <summary>
+        /// Clear the list of Repositories.
+        /// </summary>
+        public void ClearList()
+        {
+            _latestTask = null;
+            RaiseAllPropertyChanged();
```
ivannaranjo: Should you just be able to set the right property value so only that property changes? You shouldn't need to reset the entire object's state.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -55,23 +55,25 @@ public DisplayOptions DisplayState
...
-                _latestTask = null;
-            }
-            else
-            {
-                _latestTask = GetCloudReposAsync(projectId);
-                RaiseAllPropertyChanged();
-                await _latestTask;
-            }
+            _latestTask = GetCloudReposAsync(projectId);
+            RaiseAllPropertyChanged();
```
ivannaranjo: Same thing here, it seems that the properties should notify of changes themselves rather than this. Probably more of a nit at this point, but definitely something that we should look at.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            Assert.AreEqual(AsyncRepositories.DisplayOptions.Pending, _testObject.DisplayState);
+            Assert.IsNull(_testObject.Value);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(TestException))]
+        public void TestCatchException()
+        {
+            _testTaskCompletionSource.SetException(new TestException());
+            var task = ExecuteAndHandleException(() => _testObject.StartListRepoTaskAsync("projectid"));
```
ILMTitan: I am a little confused what this `ExecuteAndHandleException` is.
Could you just do this?

    public async Task TestCatchException() ...

        await _testObject.StartListRepoTaskAsync("projectid");
        Assert.Fail("Should have thrown exception.");

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            Assert.IsNotNull(task.Result);
+            throw task.Result;
+
+        }
+
+        [TestMethod]
+        public void TestPendingAndCompletion()
+        {            
+            Assert.AreEqual(AsyncRepositories.DisplayOptions.Pending, _testObject.DisplayState);
+            Assert.IsNull(_testObject.Value);
```
ILMTitan: You already test these Asserts in `TestInitialState`. Don't assert the preconditions; only assert the final results/state.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            _testTaskCompletionSource.SetResult(_testRepos);
+            _testObject.StartListRepoTaskAsync("projectid").Wait();
```
ILMTitan: Instead of using `.Wait()`, you can make the test method itself `async`, and then `await` here.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
ILMTitan: Usually (including this case) the only thing the view model should do to the window is close it. The view model can then just take an `Action closeOwner` as a constructor parameter. That is even easier to mock.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+namespace GoogleCloudExtensionUnitTests.CloudSourceRepository
+{
+    [TestClass]
+    public class CsrCloneWindowViewModelTests
+    {
+        private const int BackGroundAsyncTaskWaitSeconds = 5;
+        private Mock<CsrCloneWindow> _cloneWindowMock;
+        private CsrCloneWindowViewModel _cloneWindowViewModel;
+        private Project _testProjectMock;
```
ILMTitan: My convention names this `_mockedTestProject` to distinguish that this is an object that is mocked, rather than a mock that has an object.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            _apiManagerMock.Verify();
+        }
+
+        /// <summary>
+        /// This method wait for a few seconds till <param name="cloneWindowViewModel"/> IsReady is true;
+        /// </summary>
+        private void WaitForBackgroundAsyncTask(CsrCloneWindowViewModel cloneWindowViewModel)
+        {
+            int waitedSeconds = 0;
+            Thread.Sleep(1000); // Wait for async task kick off
```
ILMTitan: Unit tests should usually take 10s of milliseconds. What is taking so long for the cloneWindowViewModel to set up, and why can't it be abstracted?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -65,10 +70,27 @@ public Project SelectedProject
...
-                if (_selectedProject != null && IsReady && oldValue != _selectedProject)
+                if (_selectedProject == null)
+                {
+                    RepositoriesAsync.ClearList();
+                }
+                else if (oldValue != _selectedProject)
                 {
-                    ErrorHandlerUtils.HandleAsyncExceptions(() =>
-                        RepositoriesAsync.StartListRepoTaskAsync(_selectedProject.ProjectId));
+                    ErrorHandlerUtils.HandleAsyncExceptions(() => ExecuteAsync(async () =>
```
ILMTitan: What is this `ExecuteAsync` method doing? Why is it necessary?

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            Assert.AreSame(_emptyList, _testObject.Value);
+
+        }
+
+        [TestMethod]
+        public void TestClearList()
+        {
+            _testTaskCompletionSource.SetResult(_testRepos);
+            _testObject.StartListRepoTaskAsync("projectid").Wait();
+            Assert.AreEqual(AsyncRepositories.DisplayOptions.HasItems, _testObject.DisplayState);
```
ILMTitan: Again, this is a precondition, already tested in `TestPendingAndCompletion`.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            _testObject.ClearList();
+            Assert.IsNull(_testObject.Value);
```
ILMTitan: Should `_testObject.DisplayState` also change? If so, you should add an assert for it.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        [TestMethod]
+        public void TestPendingAndCompletion()
+        {            
+            Assert.AreEqual(AsyncRepositories.DisplayOptions.Pending, _testObject.DisplayState);
+            Assert.IsNull(_testObject.Value);
+            _testTaskCompletionSource.SetResult(_testRepos);
+            _testObject.StartListRepoTaskAsync("projectid").Wait();
+            Assert.AreEqual(AsyncRepositories.DisplayOptions.HasItems, _testObject.DisplayState);
+            Assert.AreSame(_testRepos, _testObject.Value);
```
ILMTitan: You don't care if `AsyncRepositories` returns the exact same object. If it decides making a defensive copy is useful, this will fail. Instead, use `CollectionAssert.AreEqual()`.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            _testProjectMock.ProjectId = "test_project_id";
+            _testTaskCompletionSource.SetResult(new List<Repo>());
+            _apiManagerMock.Setup(
+                x => x.EnsureAllServicesEnabledAsync(It.IsAny<IEnumerable<string>>(), It.IsAny<string>()))
+                .Returns(Task.FromResult(false));
+            _cloneWindowViewModel.SelectedProject = _testProjectMock;
+            WaitForBackgroundAsyncTask(_cloneWindowViewModel);
+            Assert.AreEqual(null, _cloneWindowViewModel.SelectedRepository);
+            _apiManagerMock.Verify();
```
ILMTitan: This call does not verify the setup. The setup must be `.Verifiable()` for this to be a valid test.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        [TestMethod]
+        public void ChangeSelectedProjectTest()
+        {
+            _testProjectMock.ProjectId = "test_project_id";
+            Repo repoMock = Mock.Of<Repo>();
+            _testTaskCompletionSource.SetResult(new List<Repo> { repoMock, Mock.Of<Repo>() });
+            _apiManagerMock.Setup(
+                x => x.EnsureAllServicesEnabledAsync(It.IsAny<IEnumerable<string>>(), It.IsAny<string>()))
```
ILMTitan: You probably only need one test that verifies this call. You should probably also have a test for the view model state when this call's task is still pending. Those can probably be the same call. I would argue in favor of doing this setup once, and verifying a specific setup:

    _apiManagerMock.Verify(m => m.EnsureAllServiceEnabledAsync(CsrCloneWindowViewModel.s_requiredApis, "test_project_id"), Times.Once);

Deren-Liao:
I'll make more changes to unit tests. Please hold on for review

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
Deren-Liao: unit test does not work well with async method. As far as I tested, it always passes.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
Deren-Liao: I tried to modify the test to async, the test execution engine just run pass without waiting though.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
Deren-Liao: Great to learn!  
I vaguely recall the Python unit test (or c++) framework by default let you "Verify()".

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
Deren-Liao: It complains IList can't be converted to ICollection. 
As for now they are same and test passes. If later one changes the code and break the test, he/she can just modify the test.

GoogleCloudExtension/GoogleCloudExtension/ApiManagement/IApiManager.cs
Deren-Liao: Done.

Deren-Liao:
Please take a look again.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -97,5 +97,11 @@ public enum DisplayOptions
             /// </summary>
             HasItems
         }
+
+        private void UpdateValueAndDisplayState()
```
ivannaranjo: Still find this suspect, why is the property not raising the change itself? That is, why is not the `Value` and `DisplayState` setters raising the property changed event? It can be something to be dealt with in a different PR since it is not central to this PR.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
ivannaranjo: This doesn't seem right. Is the data source correclty mocked?

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
```diff
@@ -61,7 +61,6 @@ public void TestCatchException()
         {
             _testTaskCompletionSource.SetException(new TestException());
             var task = ExecuteAndHandleException(() => _testObject.StartListRepoTaskAsync("projectid"));
-            Assert.IsNotNull(task.Result);
             throw task.Result;
```
ivannaranjo: This is throwing the result, which would never happend (theoretically) since the mere call to `Result` should throw by itself and if it doesn't then you should fail the test with an assertion. 

In short you should be able to use `await task` here to get the exception thrown for you.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
ivannaranjo: No, async test methods are correctly supported, I have used them myself.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -41,9 +40,8 @@ public class CsrCloneWindowViewModelTests
         public void Initialize()
         {
             _apiManagerMock = new Mock<IApiManager>();
-            _testProjectMock = Mock.Of<Project>();
-            _cloneWindowMock = new Mock<CsrCloneWindow>();
-            _testProjects = new List<Project> { _testProjectMock, Mock.Of<Project>() };
+            _mockedTestProject = Mock.Of<Project>();
```
ivannaranjo: You mix the use of `Mock.Of()` with `new Mock()`. You should settle in a single style. I would prefer we all use the more C# friendly `new Mock()` style if possible, but I am flexible as long as we're consistent.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -56,36 +54,54 @@ public void Initialize()
             AsyncRepositories.GetCloudReposAsync = _getCloudReposMock.Object;
             CsrCloneWindowViewModel.s_getApiManagerFunc = (projectId) => _apiManagerMock.Object;
 
-            _cloneWindowViewModel = new CsrCloneWindowViewModel(_cloneWindowMock.Object, _testProjects);
+            Action testAction = () => { };
```
ivannaranjo: You might want to know if the action is called to verify that the window is actually "closed" by the view model at the right time.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -94,7 +110,7 @@ public void CsrApiNotEnabledTest()
         private void WaitForBackgroundAsyncTask(CsrCloneWindowViewModel cloneWindowViewModel)
         {
             int waitedSeconds = 0;
-            Thread.Sleep(1000); // Wait for async task kick off
+            Thread.Sleep(10); // Wait 10 milliseconds for async task kick off
```
ivannaranjo: This is not really going to work if the tasks are all assigned to the same synchronization context, that is, if everything runs in the UI pump. The test framework supports async tests correctly, you should be able to make your tests async and just await tasks to complete. I have done that myself and it just works. Is really the only way to tests async methods.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/AsyncRepositoreisTests.cs
Deren-Liao: Yes, it works now...  There must be mistake somewhere.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
Deren-Liao: This is not changing the DisplayState or Value directly.  It is changing the _latestTask and triggers Value and DisplayState value change.   I agree it is useful to "SetAndRaise" . I did not do that because we don't directly change the values.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
ILMTitan: `Mock.Of<T>()` returns a `T` not a `Mock<T>`. It is most useful when you don't care about setting up anything (although it would arguably be better to just get a `new T()`).

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/AsyncRepositories.cs
```diff
@@ -88,5 +97,11 @@ public enum DisplayOptions
...
+        {
+            RaisePropertyChanged(nameof(Value));
+            RaisePropertyChanged(nameof(DisplayState));
+        }
```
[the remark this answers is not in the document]
Deren-Liao: It is used by two places.  Just to make the code a bit shorter.  Unless you strongly against it, I'd just keep it as is.

Deren-Liao:
Please take a look again.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -20,28 +20,28 @@
...
         private Mock<IApiManager> _apiManagerMock;
+        private Mock<Action> _closeWindowActionMock;
 
         [TestInitialize]
         public void Initialize()
         {
             _apiManagerMock = new Mock<IApiManager>();
-            _mockedTestProject = Mock.Of<Project>();
-            _testProjects = new List<Project> { Mock.Of<Project>(), _mockedTestProject };
+            _mockedTestProject = new Mock<Project>().Object;
```
ivannaranjo: I would recommend keeping the Mock alive and using `_mockedTestProject.Object` when required.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -54,51 +54,61 @@ public void Initialize()
...
         }
 
         [TestMethod]
-        public void ChangeSelectedProjectTest()
+        public async Task ChangeSelectedProjectTest()
         {
             _mockedTestProject.ProjectId = "test_project_id";
-            Repo repoMock = Mock.Of<Repo>();
-            _testTaskCompletionSource.SetResult(new List<Repo> { repoMock, Mock.Of<Repo>() });
+            Repo repoMock = new Mock<Repo>().Object;
```
ivannaranjo: Same, keep the Mock object around.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
```diff
@@ -54,51 +54,61 @@ public void Initialize()
...
                 .Verifiable();
-            Repo repoMock = Mock.Of<Repo>();
-            _testTaskCompletionSource.SetResult(new List<Repo> { repoMock, Mock.Of<Repo>() });
+            Repo repoMock = new Mock<Repo>().Object;
+            _testTaskCompletionSource.SetResult(new List<Repo> { repoMock, new Mock<Repo>().Object });
             Assert.IsTrue(_cloneWindowViewModel.EnableApiCommand.CanExecuteCommand);
             _cloneWindowViewModel.EnableApiCommand.Execute(null);
-            WaitForBackgroundAsyncTask(_cloneWindowViewModel);
+            await WaitForBackgroundAsyncTask(_cloneWindowViewModel);
             _apiManagerMock.Verify();
```
ivannaranjo: What is is this verifying?

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudSourceRepository/CsrCloneWindowViewModelTests.cs
Deren-Liao: Wait for the "repo" list to be populated.

Deren-Liao:
Thank you for doing the review

PR #837, opened 2017-10-26
Create an App Engine app

ivannaranjo:
This PR adds the functionality to create a new App Engine app directly from Visual Studio when required. There are two entrypoints to the new functionality:
* A right click menu in the placeholder shown when no app is found in the Cloud Explorer.
* During a deployment to App Engine flexible, when the extension detects that there's no app.

This PR also merges the functionality of the `Polling<T>` class into the `OperationUtils` so we have a single way of polling for operations moving forward.

Fixes #634 Fixed #508

PR #851, opened 2017-11-28
Fix margins

### LINKED ISSUE #832: The spacing of the project selection controls is inconsistent with the rest of the UI
The `TextBox` is too close to the "Clusters" tag. The problem appears in the App Engine publish step and the GKE step. See below for the how the problem looks in the GKE step.

screen shot 2017-10-20 at 2 18 52 pm

ILMTitan:
Make dialog content spacing more in line with visual studio dialog layout examples.

Fixes #832


ILMTitan:
The issue this PR was meant to fix was fixed in PR #837.

PR #852, opened 2017-11-28
Change the publish dialogs to use content control with data triggers.

ILMTitan:
These changes make it easier to see what the dialog looks like at various stages in the designer.


ILMTitan:
It was a reason, but the main reason is `ContentControl` provides a more readable experience, creating separation between "how it looks" code and "what is shown" code, as well as signaling "only one is shown" better than naked visibility bindings.

ivannaranjo:
As I mentioned, look into using sample data instead of this approach. Is much tooling friendly. The reason to do visibility bindings is still clearer code than the _action as a distance_ that you are using here with style setters.

ivannaranjo:
Other than that I do not want us to use this pattern, makes things more complex than necessary (action at a distance) and only covers the case when the entire content control is to be replaced.

ILMTitan:
The fact this pattern only covers the case when the entire content control is to be replaced is part of the point. When you see this pattern, you know an entire section is being swapped out, while when you see a visibility binding, you expect that section to add to the rest of the UI in some cases.

ivannaranjo:
I'm rejecting this PR, please do not use this pattern. We can talk more about this. At this point I prefer us to use the visiblity and design time data to design and test the UI rather than this style.

PR #853, opened 2017-11-29
Change the "Features:" label on the templates to "Template type:".

### LINKED ISSUE #823: The text "Select a feature" seems very vague
Looking at the text for the template wizard it seems that the "Select a feature" label is not very specific. What the user has to choose is not a feature, is an application framework. The user has to choose between using MVC or WebAPI.

We should consider making these another dropdown instead. Features should be things like authentication, storage, etc... 

screen shot 2017-10-04 at 1 47 30 pm

### LINKED ISSUE #830: Templates should not ask for GCP Project ID
To avoid a runtime exception, we made the templates ask for a GCP Project ID and put it in the config. Instead, the template code itself should disable those features when the proper environment is not detected  (when it is not running is the google cloud) and the config is not set.

ILMTitan:
Fixes #823 

Remove the Google Cloud Project ID input from the templates.
Disable stackdriver logging when GCP Project ID is not set in the templates.
Fixes #830

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Theming/DefaultStylesResource.xaml
+++ b/GoogleCloudExtension/GoogleCloudExtension/Theming/DefaultStylesResource.xaml
@@ -17,13 +17,13 @@
     </Style>
 
     <Style TargetType="TextBox" BasedOn="{StaticResource CommonTextBoxStyle}">
-        <Setter Property="Margin" Value="5"/>
+        <Setter Property="Margin" Value="0,0,12,6"/>
         <Setter Property="VerticalAlignment" Value="Center"/>
         <Setter Property="VerticalContentAlignment" Value="Center"/>
     </Style>
     
     <Style TargetType="Button" BasedOn="{StaticResource CommonButtonStyle}">
-        <Setter Property="Margin" Value="2"/>
+        <Setter Property="Margin" Value="0,0,0,6"/>
         <Setter Property="Padding" Value="0"/>
         <Setter Property="VerticalAlignment" Value="Center"/>
         <Setter Property="Width" Value="Auto"/>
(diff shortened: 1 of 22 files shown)


GoogleCloudExtension/GoogleCloudExtension/Theming/DefaultStylesResource.xaml
```diff
@@ -17,35 +17,35 @@
...
-        <Setter Property="Margin" Value="2"/>
+        <Setter Property="Margin" Value="0,0,12,6"/>
         <Setter Property="Padding" Value="2"/>
         <Setter Property="VerticalAlignment" Value="Center"/>
         <Setter Property="VerticalContentAlignment" Value="Center"/>
     </Style>
     
     <Style TargetType="Label" BasedOn="{StaticResource CommonLabelStyle}">
-        <Setter Property="Margin" Value="5"/>
+        <Setter Property="Margin" Value="0,0,0,2"/>
```
ivannaranjo: I have to ask, where are these margins coming from?

GoogleCloudExtension/ProjectTemplates/GcpProjectTemplate/Core/MVC/Startup.cs
```diff
@@ -20,25 +27,25 @@ public Startup(IHostingEnvironment env)
...
-            // Sends Exceptions to Stackdriver Error Reporting.
-            services.AddGoogleExceptionLogging(
-                options =>
-                {
-                    options.ProjectId = projectId;
-                    options.ServiceName = GetServiceName();
-                    options.Version = GetVersion();
-                });
+
+            if (HasProjectId)
```
ivannaranjo: Consider logging a warning if `HasProjectId` is false, this way the user can see that logging was disabled and why.

GoogleCloudExtension/GoogleCloudExtension/Theming/DefaultStylesResource.xaml
ILMTitan: These margins are coming from [Layout for Visual Studio](https://docs.microsoft.com/en-us/visualstudio/extensibility/ux-guidelines/layout-for-visual-studio). The figure there indicates a distance between a label and its control of 5, but that is to the baseline of the label. I am guessing that defenders (y,g,j etc.) take 3 pixels of that.

GoogleCloudExtension/ProjectTemplates/GcpProjectTemplate/Core/MVC/Startup.cs
ILMTitan: Done.

ILMTitan:
I am not as concerned about missing ServiceName and Version configuration, as they both have reasonable, straightforward defaults.

PR #856, opened 2017-12-06
Changing the way the project is selected.

### LINKED ISSUE #794: If the list of projects available for the current user is long the cloud explorer appears to hang.
The issue comes from the fact that it might take a long time to fully load the list of projects available for the current user. There's currently no indication that a load operation is in flight, which makes the cloud explorer appear hung. We need to add an indicator to the UI (perhaps in the dropdown) that let's the user know what is going on.

ivannaranjo:
There's quite a bit of rework in this PR to change the way the concept of _current project_ is handled by the extension in the `CredentialsStore` class. The class will also keep a list of all of the projects accessible by the current user, this way changing between projects will not require a total reload of projects. The user can request a refresh of the list of projects at any time, 

There's a bit of work to add search to the dialog as well, so the projects are filtered as the user types. When the user has many many projects this is incredibly useful. We no longer sort the projects on the API, the user can choose any sort order they want/need on the `DataGrid` that is the in the UI instead.

The refactoring includes moving the pick project dialog out of the wizard and into the main structure of the extension. I am also changing the way the projects are loaded by dialog to make use of the common `Task` that contains the projects for the current user.

Some of the tests are failing at the moment, I am reworking how to do the tests for the `PickProjectIdViewModel` class. 

The 

Fixes #794


GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml
```diff
@@ -0,0 +1,167 @@
...
+                         Style="{StaticResource CommonTextBoxStyle}"
+                         IsEnabled="{Binding LoadTask.IsSuccess}"
+                         TextChanged="OnFilterTextChanged" />
+            </Grid>
+
+            <!-- Message shown when there's no account selected. -->
+            <StackPanel VerticalAlignment="Center"
+                        Grid.Row="2"
+                        Grid.ColumnSpan="2"
+                        Visibility="{Binding HasAccount, TargetNullValue=false,Converter={utils:VisibilityConverter IsNegated=True}}"
```
ILMTitan: I am not a fan of changing out the content control for simple visibility bindings.

`ContentControl` makes it obvious that exactly one of several elements will be visible. It also marks exactly which elements could be swapped; those in the `ContentControl.Resources` element. It also puts all of the logic to determine which one to show in a single location, the Triggers.

With the visibility bindings, to see that elements will be swapped out, you have to first notice a bunch of elements have the same gird coordinates, and have visibility bindings, and that the visibility bindings are mutually exclusive. To determine which elements might be shown, you have to read all of the visibility bindings, and determine which are mutually exclusive, and which are not. Even this may not be clear; it is non-obvious that `HasAccount` is related to `LoadTask`. Finally, the logic of which one is shown when is scattered in 4 different places over 50 lines.

Using `ContentControl` to mark a UI element with several possible layouts is much more readable.

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml
```diff
@@ -0,0 +1,167 @@
+﻿<!--
+    Change the d:DataContext attribute to {d:DesignInstance sample:<sample class>, IsDesignTimeCreatable=true}
+    classes to change the UI state as follows:
+    * Use sample:ErrorState to see the UI in the error state.
+    * Use sample:LoadingState to see the UI in the loading state.
+    * Use sample:LoadedState to see the UI in the loaded state, with projects in the DataGrid.
+    * Use sample:NeedsAccountState to see the UI in the needs account state.
```
ILMTitan: This was not working for me, but does seem like a better solution than setting a default value on a control.

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml
```diff
@@ -0,0 +1,167 @@
...
+
+    <theming:CommonDialogWindowBaseContent HasBanner="True">
+        <theming:CommonDialogWindowBaseContent.Buttons>
+            <theming:DialogButtonInfo
+                Caption="{x:Static ext:Resources.UiOkButtonCaption}"
+                Command="{Binding OkCommand}"
+                IsDefault="True"/>
+            <theming:DialogButtonInfo
+                Caption="{x:Static ext:Resources.UiRefreshButtonCaption}"
+                Command="{Binding RefreshCommand}" />
```
ILMTitan: From the [Layout for Visual Studio](https://msdn.microsoft.com/en-us/library/mt162317.aspx) guidelines:
> Command buttons operate on the dialog frame, not on the content.

Refresh operates on the content, while OK and Cancel operate on the dialog. Move the refresh button inside of the dialog, possibly inside of the `LoadTask.IsSuccess` swapped element.

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        private void OnFilterItemInCollectionView(object sender, System.Windows.Data.FilterEventArgs e)
+        {
+            var viewModel = DataContext as PickProjectIdViewModel;
+            e.Accepted = viewModel?.FilterItem(e.Item) ?? false;
+        }
+
+        private void OnFilterTextChanged(object sender, System.Windows.Controls.TextChangedEventArgs e)
+        {
+            var cvs = Resources["cvs"] as CollectionViewSource;
```
ILMTitan: If you move `csv` out of the resources and create it in `DataGrid.Items` directly:


                <DataGrid.Items>
                    <CollectionViewSource
                        x:Name="_csv"
                        Source="{Binding Projects}"
                        Filter="OnFilterItemInCollectionView"
                        IsLiveFilteringRequested="True" />
                </DataGrid.Items>

You can make this call simply with no casting or magic strings:

            _csv.View.Refresh();

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/CloudExplorerViewModel.cs
```diff
@@ -405,47 +398,32 @@ private void InvalidateEmptyState()
...
         {
+            // Clear the roots collection to clean the UI.
+            Roots = null;
             foreach (var source in _sources)
             {
                 source.Refresh();
             }
-            RaisePropertyChanged(nameof(Roots));
+
+            // Wait for a full cycle so the automation server has time to adapt to the chagnes in the UI.
```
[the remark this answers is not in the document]
ivannaranjo: The UI automation server reads the UI state and uses it to update screen readers and other accessibility clients. It can also be used to automate the UI triggering actions etc... In the case of the `TreeView` it seems that it holds on to old UI elements and we need to give it a cycle to update itself.

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml
ivannaranjo: Indeed, you can change this without affecting the runtime state of the app, and you still get to see the UI just as it would be for the data, including lists, button states, etc...

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml
ivannaranjo: Actually is not, your use of style setters makes the UI much more difficult to author and really difficult to read what is going on, since the rules that determine what UI is visible is not in the UI but in a separate section of the xaml.

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml.cs
ivannaranjo: There's no magic going on, that is how resources work.

ILMTitan: I meant magic string in the [magic constant anti-pattern](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants) sense.

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml
ILMTitan: I am arguing that the rules that determine which UI section is visible are all closely related and should be in a single section of the xaml. I also argue that what a part of the UI looks like and when it is visible are not closely related, and can be separated. The fact that the visibility of the **Loading** UI section is controlled by `LoadTask.IsPending` rather than some other property on the view model is irreverent to whether the **Loading** UI section should have a static block of text or an animated spinner.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/CloudExplorerToolWindowControl.xaml
```diff
@@ -321,13 +321,24 @@
...
 
         <!-- Project switcher. -->
-        <ComboBox ItemsSource="{Binding EffectiveProjects}"
-                  IsEnabled="{Binding IsNotEmptyState}"
-                  SelectedItem="{Binding CurrentProject}"
-                  SelectedIndex="0"
-                  Grid.Row="1"
-                  Style="{DynamicResource {x:Static vsshell:VsResourceKeys.ComboBoxStyleKey}}" />
-
+        <Grid  Grid.Row="1" IsEnabled="{Binding LoadingProject, Converter={utils:NotConverter}}">
```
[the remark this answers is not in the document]
ivannaranjo: I understand, but it actually complicates things more, this binding expression is now making it explicit what it is binding to, the negation of the `LoadingProject` property. I would agree with you if there was a double negative involved, but there isn't one.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrReposViewModel.cs
```diff
@@ -375,7 +375,7 @@ private void CreateEmptySolutionAtPath(string localPath)
             IsReady = false;
             try
             {
-                var projects = await resourceManager.GetSortedActiveProjectsAsync();
+                var projects = await resourceManager.GetProjectsListAsync();
```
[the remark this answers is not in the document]
ivannaranjo: No, the sorting will be done in the UI by the user.

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml.cs
ivannaranjo: ACK, but extracting a string when it is used in a single spot in code is a bit too much.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/CloudExplorerViewModel.cs
```diff
@@ -91,7 +93,23 @@ private set
...
+            private set { SetValueAndRaise(ref _roots, value); }
+        }
+
+        public string ProjectDisplayString
+        {
+            get { return _projectDisplayString; }
+            private set { SetValueAndRaise(ref _projectDisplayString, value); }
+        }
+
+        public bool LoadingProject
```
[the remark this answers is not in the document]
ivannaranjo: Actually this was redundant, the overall `Grid` is enabled/disabled during the load of the data which is a superset of this property. I am removing this.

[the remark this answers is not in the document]
ivannaranjo: The field will remain to avoid re-entrancy issues.

ivannaranjo:
PTAL

GoogleCloudExtension/GoogleCloudExtension/PickProjectDialog/PickProjectIdWindowContent.xaml
```diff
@@ -90,6 +89,18 @@
                          Style="{StaticResource CommonTextBoxStyle}"
                          IsEnabled="{Binding LoadTask.IsSuccess}"
                          TextChanged="OnFilterTextChanged" />
+
+                <Button Grid.Column="2"
+                        Margin="5,0,0,0"
+                        IsEnabled="{Binding LoadTask.IsSuccess}"
```
ILMTitan: Why is this disabled on an error? That seems like the most likely time a refresh would be required.

PR #857, opened 2017-12-07
Create an Options page for the Cloud Explorer.

### LINKED ISSUE #562: Add option page for Pub/Sub filtered topics.
Add an option page to Google Cloud Tools options in Visual Studio options so users can update the list of filtered Pub/Sub topics themselves.

ILMTitan:
Add the ability to select which topics are filtered from the cloud explorer.

Fixes #562

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
@@ -153,14 +153,20 @@ namespace GoogleCloudExtension.CloudExplorerSources.PubSub
         /// <summary>
         /// Gets the last part of the full name i.e. the leaf of the path.
         /// </summary>
-        public static string GetPathLeaf(string path)
+        /// <param name="fullTopicName">The full topic name (e.g. <code>"projects/project-id/topics/topic-name"</code>)</param>
+        public static string GetPathLeaf(string fullTopicName)
         {
-            return GetPathSections(path).LastOrDefault();
+            return GetPathSections(fullTopicName).LastOrDefault();
         }
 
-        public static string GetTopicProject(string path)
+        /// <summary>
+        /// Gets the project part of a full topic name.
+        /// </summary>
+        /// <param name="fullTopicName">The full topic name (e.g. <code>"projects/project-id/topics/topic-name"</code>)</param>
+        /// <returns>The project id part of the full topic name. This must be the second section.</returns>
+        public static string GetTopicProject(string fullTopicName)
         {
-            return GetPathSections(path).Skip(1).FirstOrDefault();
+            return GetPathSections(fullTopicName).Skip(1).FirstOrDefault();
         }
 
         private static string[] GetPathSections(string path)
--- a/GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
@@ -19,6 +19,7 @@ using Google.Apis.Pubsub.v1.Data;
 using GoogleCloudExtension.DataSources;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace GoogleCloudExtension.CloudExplorerSources.PubSub
@@ -154,7 +155,17 @@ namespace GoogleCloudExtension.CloudExplorerSources.PubSub
         /// </summary>
         public static string GetPathLeaf(string path)
         {
-            return path.Substring(1 + path.LastIndexOf("/", StringComparison.Ordinal));
+            return GetPathSections(path).LastOrDefault();
+        }
+
+        public static string GetTopicProject(string path)
+        {
+            return GetPathSections(path).Skip(1).FirstOrDefault();
+        }
+
+        private static string[] GetPathSections(string path)
+        {
+            return path.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
         }
 
         /// <summary>
--- a/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/NewInstallEvent.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/NewInstallEvent.cs
@@ -20,7 +20,7 @@ namespace GoogleCloudExtension.Analytics.Events
     /// </summary>
     internal static class NewInstallEvent
     {
-        internal const string NewInstallEventName = "newInstall";
+        private const string NewInstallEventName = "newInstall";
 
         public static AnalyticsEvent Create()
         {
--- a/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/NewInstallEvent.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/NewInstallEvent.cs
@@ -20,7 +20,7 @@ namespace GoogleCloudExtension.Analytics.Events
     /// </summary>
     internal static class NewInstallEvent
     {
-        private const string NewInstallEventName = "newInstall";
+        internal const string NewInstallEventName = "newInstall";
 
         public static AnalyticsEvent Create()
         {
--- a/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/UpgradeEvent.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/UpgradeEvent.cs
@@ -20,7 +20,7 @@ namespace GoogleCloudExtension.Analytics.Events
     /// </summary>
     internal static class UpgradeEvent
     {
-        internal const string UpgradeEventName = "upgrade";
+        private const string UpgradeEventName = "upgrade";
 
         public static AnalyticsEvent Create()
         {
--- a/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/UpgradeEvent.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/UpgradeEvent.cs
@@ -20,7 +20,7 @@ namespace GoogleCloudExtension.Analytics.Events
     /// </summary>
     internal static class UpgradeEvent
     {
-        private const string UpgradeEventName = "upgrade";
+        internal const string UpgradeEventName = "upgrade";
 
         public static AnalyticsEvent Create()
         {
--- a/GoogleCloudExtension/GoogleCloudExtension/Analytics/EventsReporterWrapper.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Analytics/EventsReporterWrapper.cs
@@ -30,14 +30,18 @@ namespace GoogleCloudExtension.Analytics
 
         private const string PropertyId = "UA-36037335-1";
 
-        private static Lazy<IEventsReporter> s_reporter = new Lazy<IEventsReporter>(CreateReporter);
+        // For testing.
+        internal static Lazy<IEventsReporter> ReporterLazy { private get; set; } = new Lazy<IEventsReporter>(CreateReporter);
+
+        private static IEventsReporter Reporter => ReporterLazy.Value;
+
 
         /// <summary>
         /// Used by unit test to prevent analytics from running.
         /// </summary>
         public static void DisableReporting()
         {
-            s_reporter = new Lazy<IEventsReporter>(() => null);
+            ReporterLazy = new Lazy<IEventsReporter>(() => null);
         }
 
         /// <summary>
@@ -64,24 +68,23 @@ namespace GoogleCloudExtension.Analytics
         /// </summary>
         public static void AnalyticsOptInStateChanged()
         {
-            s_reporter = new Lazy<IEventsReporter>(CreateReporter);
+            ReporterLazy = new Lazy<IEventsReporter>(CreateReporter);
         }
 
         /// <summary>
         /// Called to report an interesting event to analytics. If there's a queue of events it will be
         /// flushed as well.
         /// </summary>
-        /// <param name="eventData"></param>
-        public static void ReportEvent(AnalyticsEvent eventData)
-        {
-            s_reporter.Value?.ReportEvent(
-                source: ExtensionEventSource,
-                eventType: ExtensionEventType,
-                eventName: eventData.Name,
-                userLoggedIn: CredentialsStore.Default.CurrentAccount != null,
-                projectNumber: CredentialsStore.Default.CurrentProjectNumericId,
-                metadata: eventData.Metadata);
-        }
+        public static Action<AnalyticsEvent> ReportEvent { get; internal set; } = (AnalyticsEvent eventData) =>
+         {
+             Reporter?.ReportEvent(
+                 source: ExtensionEventSource,
+                 eventType: ExtensionEventType,
+                 eventName: eventData.Name,
+                 userLoggedIn: CredentialsStore.Default.CurrentAccount != null,
+                 projectNumber: CredentialsStore.Default.CurrentProjectNumericId,
+                 metadata: eventData.Metadata);
+         };
 
         private static IEventsReporter CreateReporter()
         {
--- a/GoogleCloudExtension/GoogleCloudExtension/Analytics/EventsReporterWrapper.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Analytics/EventsReporterWrapper.cs
@@ -30,18 +30,14 @@ namespace GoogleCloudExtension.Analytics
 
         private const string PropertyId = "UA-36037335-1";
 
-        // For testing.
-        internal static Lazy<IEventsReporter> ReporterLazy { private get; set; } = new Lazy<IEventsReporter>(CreateReporter);
-
-        private static IEventsReporter Reporter => ReporterLazy.Value;
-
+        private static Lazy<IEventsReporter> s_reporter = new Lazy<IEventsReporter>(CreateReporter);
 
         /// <summary>
         /// Used by unit test to prevent analytics from running.
         /// </summary>
         public static void DisableReporting()
         {
-            ReporterLazy = new Lazy<IEventsReporter>(() => null);
+            s_reporter = new Lazy<IEventsReporter>(() => null);
         }
 
         /// <summary>
@@ -68,23 +64,24 @@ namespace GoogleCloudExtension.Analytics
         /// </summary>
         public static void AnalyticsOptInStateChanged()
         {
-            ReporterLazy = new Lazy<IEventsReporter>(CreateReporter);
+            s_reporter = new Lazy<IEventsReporter>(CreateReporter);
         }
 
         /// <summary>
         /// Called to report an interesting event to analytics. If there's a queue of events it will be
         /// flushed as well.
         /// </summary>
-        public static Action<AnalyticsEvent> ReportEvent { get; internal set; } = (AnalyticsEvent eventData) =>
-         {
-             Reporter?.ReportEvent(
-                 source: ExtensionEventSource,
-                 eventType: ExtensionEventType,
-                 eventName: eventData.Name,
-                 userLoggedIn: CredentialsStore.Default.CurrentAccount != null,
-                 projectNumber: CredentialsStore.Default.CurrentProjectNumericId,
-                 metadata: eventData.Metadata);
-         };
+        /// <param name="eventData"></param>
+        public static void ReportEvent(AnalyticsEvent eventData)
+        {
+            s_reporter.Value?.ReportEvent(
+                source: ExtensionEventSource,
+                eventType: ExtensionEventType,
+                eventName: eventData.Name,
+                userLoggedIn: CredentialsStore.Default.CurrentAccount != null,
+                projectNumber: CredentialsStore.Default.CurrentProjectNumericId,
+                metadata: eventData.Metadata);
+        }
 
         private static IEventsReporter CreateReporter()
         {
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptions.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptions.cs
@@ -12,15 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using GoogleCloudExtension.Utils;
 using Microsoft.VisualStudio.Shell;
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.ComponentModel;
 using System.Windows;
 
 namespace GoogleCloudExtension.CloudExplorer.Options
 {
-    public class CloudExplorerOptions : UIElementDialogPage
+    [DesignerCategory("Code")]
+    public class CloudExplorerOptions : UIElementDialogPage, ICloudExplorerOptions
     {
         private static readonly IReadOnlyList<string> s_defaultPubSubTopicFilters = new[]
         {
@@ -39,12 +41,8 @@ namespace GoogleCloudExtension.CloudExplorer.Options
         /// </summary>
         public IEnumerable<string> PubSubTopicFilters
         {
-            get { return _child.ViewModel.PubSubTopicFilters.Select(s => s.Regex); }
-            set
-            {
-                _child.ViewModel.PubSubTopicFilters = value
-                    .Select(s => new CloudExplorerOptionsPageViewModel.PubSubTopicRegex(s)).ToList();
-            }
+            get { return _child.ViewModel.PubSubTopicFilters.Values(); }
+            set { _child.ViewModel.PubSubTopicFilters = value.ToEditableModels(); }
         }
 
         /// <inheritdoc />
@@ -53,15 +51,11 @@ namespace GoogleCloudExtension.CloudExplorer.Options
         /// <inheritdoc />
         protected override UIElement Child => _child;
 
-        public event EventHandler SavingSettingToStorage;
+        public event EventHandler SavingSettings;
 
-        public CloudExplorerOptions() : this(null)
+        public CloudExplorerOptions()
         {
-        }
-
-        internal CloudExplorerOptions(CloudExplorerOptionsPage child)
-        {
-            _child = child ?? new CloudExplorerOptionsPage(this);
+            _child = new CloudExplorerOptionsPage(this);
             AutomationObject = new SerializableCloudExplorerOptions(this);
         }
 
@@ -82,7 +76,7 @@ namespace GoogleCloudExtension.CloudExplorer.Options
         /// <inheritdoc />
         public override void SaveSettingsToStorage()
         {
-            SavingSettingToStorage?.Invoke(this, EventArgs.Empty);
+            SavingSettings?.Invoke(this, EventArgs.Empty);
             base.SaveSettingsToStorage();
         }
     }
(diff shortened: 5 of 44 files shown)


GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/NewInstallEvent.cs
```diff
@@ -20,7 +20,7 @@ namespace GoogleCloudExtension.Analytics.Events
     /// </summary>
     internal static class NewInstallEvent
     {
-        private const string NewInstallEventName = "newInstall";
+        internal const string NewInstallEventName = "newInstall";
```
ivannaranjo: This change appears unrelated to the fix.

GoogleCloudExtension/GoogleCloudExtension/Analytics/Events/UpgradeEvent.cs
```diff
@@ -20,7 +20,7 @@ namespace GoogleCloudExtension.Analytics.Events
     /// </summary>
     internal static class UpgradeEvent
     {
-        private const string UpgradeEventName = "upgrade";
+        internal const string UpgradeEventName = "upgrade";
```
ivannaranjo: Same here, unrelated to the fix.

GoogleCloudExtension/GoogleCloudExtension/Analytics/EventsReporterWrapper.cs
```diff
@@ -30,14 +30,18 @@ internal static class EventsReporterWrapper
...
+        private static IEventsReporter Reporter => ReporterLazy.Value;
+
 
         /// <summary>
         /// Used by unit test to prevent analytics from running.
         /// </summary>
         public static void DisableReporting()
         {
-            s_reporter = new Lazy<IEventsReporter>(() => null);
+            ReporterLazy = new Lazy<IEventsReporter>(() => null);
```
ivannaranjo: These changes are unrelated to the new options page, please split into their own PR for ease of review.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPage.xaml
```diff
@@ -0,0 +1,51 @@
...
+            
+            <DataGrid.Columns>
+                <DataGridTextColumn
+                    Header="{x:Static ext:Resources.CloudExplorerOptionsPubSubTopicsFilterColumHeader}"
+                    IsReadOnly="False"
+                    Binding="{Binding Value, UpdateSourceTrigger=PropertyChanged}"
+                    Width="*"/>
+            </DataGrid.Columns>
+        </DataGrid>
+        <Button
```
ivannaranjo: Can we use the standard styles for the `Button` so it looks like other buttons in the extension?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPage.xaml.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// Creates and initializes a new <see cref="CloudExplorerOptionsPage"/>.
+        /// </summary>
+        /// <param name="parentModel"></param>
+        public CloudExplorerOptionsPage(ICloudExplorerOptions parentModel)
+        {
+            ViewModel = new CloudExplorerOptionsPageViewModel(parentModel.ResetSettings);
+            DataContext = ViewModel;
+            parentModel.SavingSettings += UpdateBindingSources;
+            AddHandler(UIElementDialogPage.DialogKeyPendingEvent, new RoutedEventHandler(OnDialogKeyPending));
+            InitializeComponent();
```
ivannaranjo: `InitializeComponent()` should be the first thing.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPage.xaml.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// <summary>
+        /// Handles the <see cref="UIElementDialogPage.DialogKeyPendingEvent"/>,
+        /// allowing this WPF element to catch enter and escape keys during editing mode.
+        /// </summary>
+        /// <param name="sender">Unused.</param>
+        /// <param name="args">
+        /// Updates the <see cref="RoutedEventArgs.Handled"/> property.
+        /// </param>
+        private void OnDialogKeyPending(object sender, RoutedEventArgs args)
```
ivannaranjo: Shouldn't `enter` and `esc` be handled by the `DataGrid` or any other input controls to allow the user to commit, and cancel out, of the editing operations?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPageViewModel.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using GoogleCloudExtension.Utils;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+
+namespace GoogleCloudExtension.CloudExplorer.Options
+{
+    public class CloudExplorerOptionsPageViewModel : ViewModelBase
```
ivannaranjo: No class comments or public method comments.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPageViewModel.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        private ObservableCollection<EditableModel<string>> _pubSubTopicFilters;
+
+        public IEnumerable<EditableModel<string>> PubSubTopicFilters
```
ivannaranjo: Why not just make the property of type `ObservableCollection`? You can manipulate that class and the changes should be reflected in the UI. No need to create a new fresh copy of the collection every time you set it.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/ICloudExplorerOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace GoogleCloudExtension.CloudExplorer.Options
+{
+    /// <summary>
+    /// The options that affect the Cloud Explorer window.
+    /// </summary>
+    public interface ICloudExplorerOptions
```
ivannaranjo: I guess that this interface will contain more settings should we add them?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/SerializableCloudExplorerOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace GoogleCloudExtension.CloudExplorer.Options
+{
+    public class SerializableCloudExplorerOptions
```
ivannaranjo: No class comments.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/SerializableCloudExplorerOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// A json string serialized version of the PubSubTopicFilters.
+        /// Used for saving and loading to the settings storage.
+        /// </summary>
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Visible)]
+        // ReSharper disable once UnusedMember.Global
+        public string PubSubTopicFiltersJsonString
+        {
+            get
+            {
+                return _parentOptions.PubSubTopicFilters == null ?
```
ivannaranjo: I typically use `JsonConverter` to do the transition between json <-> object when I can use all of the defaults. I find it more readable.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -2942,4 +2942,16 @@
   <data name="WizardTemplateChooserTargetAspFrameworkLabel" xml:space="preserve">
     <value>Target _ASP.NET Framework:</value>
   </data>
+  <data name="CloudExplorerOptionsPubSubTopicsLabel" xml:space="preserve">
+    <value>Pub/Sub Topics to hide:</value>
+  </data>
+  <data name="CloudExplorerOptionsPubSubTopicsLabelHelp" xml:space="preserve">
+    <value>If any of the patterns match a topic's full name (e.g. projects/project-id/topics/topic-name), that topic will not be shown in the Cloud Explorer.</value>
```
ivannaranjo: This reads like a double negative of sorts. How about something like _Regular expressions for topics to exclude from the cloud explorer_ or something like that?

GoogleCloudExtension/GoogleCloudExtension/Utils/EditableModel.cs
```diff
@@ -0,0 +1,122 @@
...
+{
+
+    /// <summary>
+    /// An <see cref="IEditableObject"/> for a single object.
+    /// </summary>
+    /// <typeparam name="T">The type this EditableModel contains.</typeparam>
+    public class EditableModel<T> : Model, IEditableObject
+    {
+        private T _value;
+        private T _uneditedRegex;
```
ivannaranjo: The field `_uneditedRegex` name is problably and hold over from when this class was more specific. Should use a more generic name now.

GoogleCloudExtension/GoogleCloudExtension/Utils/EditableModel.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <param name="input">The object to wrap.</param>
+        /// <returns>A new <see cref="EditableModel{T}"/> wrapping the input.</returns>
+        public static EditableModel<T> Of<T>(T input) => new EditableModel<T>(input);
+
+        /// <summary>
+        /// Wraps the elements of an <see cref="IEnumerable{T}"/> in <see cref="EditableModel{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type to wrap.</typeparam>
+        /// <param name="input">The objects to wrap.</param>
+        public static IEnumerable<EditableModel<T>> ToEditableModels<T>(this IEnumerable<T> input) =>
```
ivannaranjo: I think that in this method you want to be eager and return a list with the models all created. Otherwise you risk recreating them unexpectedly.

GoogleCloudExtension/GoogleCloudExtension/Utils/EditableModel.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Linq;
+
+namespace GoogleCloudExtension.Utils
+{
+
+    /// <summary>
+    /// An <see cref="IEditableObject"/> for a single object.
+    /// </summary>
+    /// <typeparam name="T">The type this EditableModel contains.</typeparam>
+    public class EditableModel<T> : Model, IEditableObject
```
ivannaranjo: While I appreciate the generic class, this object is of very limited use. Only useable when the model is a single value such as a string. The `IEditableObject` interface can also be used for general models, with more than one property. For that there's really no generic way to write a class, generating code would be better.

I say this because you might find that we will not be using this generic much.

GoogleCloudExtension/GoogleCloudExtension/packages.config
```diff
@@ -16,6 +16,7 @@
   <package id="Google.Apis.ServiceManagement.v1" version="1.29.1.981" targetFramework="net452" />
   <package id="Google.Apis.SQLAdmin.v1beta4" version="1.29.1.949" targetFramework="net452" />
   <package id="Google.Apis.Storage.v1" version="1.29.1.988" targetFramework="net452" />
+  <package id="JetBrains.Annotations" version="11.1.0" targetFramework="net452" />
```
ivannaranjo: What is this new package for? What is the license?

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPage.xaml.cs
ILMTitan: Why? It doesn't seem to matter.

Fixed.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPage.xaml.cs
ILMTitan: The problem is the hosting non-WPF `UIElementDialogPage` captures non-alphanumeric keyboard input for its own purposes (e.g. closing or canceling the the whole options dialog). The routed event `UIElementDialogPage.DialogKeyPendingEvent` allows the hosted WPF control to override that behavior, by setting the event to `Handled = true`. With that done, the standard WPF events are triggered normally (which the `DataGrid` does handle).

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPageViewModel.cs
ILMTitan: Fixed.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPageViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/ICloudExplorerOptions.cs
ILMTitan: Yes, it would.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/SerializableCloudExplorerOptions.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/SerializableCloudExplorerOptions.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/Resources.resx
ILMTitan: Changed to 

*Topics with a full name (e.g. projects/project-id/topics/topic-name) that matches one of these regex patters will be hidden in the Cloud Explorer.*

GoogleCloudExtension/GoogleCloudExtension/Utils/EditableModel.cs
ILMTitan: Fixed.

GoogleCloudExtension/GoogleCloudExtension/Utils/EditableModel.cs
ILMTitan: Only if a caller iterates the enumerable twice, which is an error on their part.

GoogleCloudExtension/GoogleCloudExtension/packages.config
ILMTitan: It adds some attributes that assist with Resharper tooling. The `[UsedImplicitly]` attribute on the `EditableModel()` constructor comes from this package, and prevents Resharper from complaining that the constructor not used.

The license on the package is the [MIT license](https://raw.githubusercontent.com/JetBrains/ExternalAnnotations/master/LICENSE.md).
See the package on [NuGet](https://www.nuget.org/packages/JetBrains.Annotations).


Looking at it again, I don't need it. Removing the package.

GoogleCloudExtension/GoogleCloudExtension/Utils/EditableModel.cs
ivannaranjo: It is never an error to enumerate enumerables, that's what they are for.

GoogleCloudExtension/GoogleCloudExtension/Utils/EditableModel.cs
ivannaranjo: Yep, but we can undo it or de emphasize it.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPage.xaml.cs
ivannaranjo: I see.

ivannaranjo:
I think that we need to rethink this and possibly add this as an option for the cloud explorer itself and have it by project, etc... I do not think that we can have a global list of exclusions that can apply to every single project the user works with. In short, I think that the global options are not the right place to put this options.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPage.xaml
ILMTitan: Because I am importing *DefaultStylesResource.xaml*, rather than *ThemingResources.xaml*, this `Button` by default has the standard style defined in *TemingResources.xaml*.

ILMTitan:
I added some context menus to make it easier to access the options from the cloud explorer. I also created Issue #864 to track making a per-GCP Project version of these options.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -154,7 +155,17 @@ public class PubsubDataSource : DataSourceBase<PubsubService>, IPubsubDataSource
         /// </summary>
         public static string GetPathLeaf(string path)
         {
-            return path.Substring(1 + path.LastIndexOf("/", StringComparison.Ordinal));
+            return GetPathSections(path).LastOrDefault();
+        }
+
+        public static string GetTopicProject(string path)
```
ivannaranjo: Nit, method comment.

ivannaranjo: Specially important to add examples of the transformation that this method produces.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
```diff
@@ -19,6 +19,7 @@
 using GoogleCloudExtension.DataSources;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace GoogleCloudExtension.CloudExplorerSources.PubSub
```
ivannaranjo: Consider at some point having a full resource parser. The names used by the API follow a very particular format that is only _path like_ enough to allow these manipulations to succeed. This could be a future enhancement.

GoogleCloudExtension/GoogleCloudExtension/Analytics/EventsReporterWrapper.cs
```diff
@@ -75,16 +75,16 @@ public static void AnalyticsOptInStateChanged()
         /// Called to report an interesting event to analytics. If there's a queue of events it will be
         /// flushed as well.
         /// </summary>
-        public static Action<AnalyticsEvent> ReportEvent { get; internal set; } = (AnalyticsEvent eventData) =>
```
ivannaranjo: This change doesn't seem that should be part of the change for the cloud explorer options page.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptions.cs
```diff
@@ -45,7 +51,11 @@ public class CloudExplorerOptions : UIElementDialogPage, ICloudExplorerOptions
         public IEnumerable<string> PubSubTopicFilters
```
ivannaranjo: Since you are using an `ObservableCollection<T>` you should make the property that type. And then add/remove from the collection. Creating a new instance when you set the value breaks the observable model and makes the binding expression more expensive for no gain.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPageViewModel.cs
```diff
@@ -14,26 +14,21 @@
...
 {
+    /// <summary>
+    /// View model for the <see cref="CloudExplorerOptionsPage"/>.
+    /// </summary>
     public class CloudExplorerOptionsPageViewModel : ViewModelBase
     {
         private ObservableCollection<EditableModel<string>> _pubSubTopicFilters;
 
-        public IEnumerable<EditableModel<string>> PubSubTopicFilters
+        public ObservableCollection<EditableModel<string>> PubSubTopicFilters
```
ivannaranjo: A property of type `ObservableCollection<T>` should typically be a readonly property, since the change notifications are coming from manipulating the collection. Therefore you should just assign the value to the property during the construction of the object.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPageViewModel.cs
```diff
@@ -14,26 +14,21 @@
...
         {
             get { return _pubSubTopicFilters; }
-            set
-            {
-                SetValueAndRaise(
-                    ref _pubSubTopicFilters,
-                    new ObservableCollection<EditableModel<string>>(
-                        value ?? Enumerable.Empty<EditableModel<string>>()));
-            }
+            set { SetValueAndRaise(ref _pubSubTopicFilters, value); }
```
ivannaranjo: Since you are using the `ObservableCollection<T>` this call to `SetValueAndRaise` is really unnecessary.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -148,6 +158,19 @@ public override void InvalidateProjectOrAccount()
             _dataSource = new Lazy<IPubsubDataSource>(_dataSourceFactory);
         }
 
+        /// <summary>
+        /// Opens an external browser to the given url.
+        /// </summary>
+        /// <param name="url"></param>
+        public void OpenBrowser(string url)
```
ivannaranjo: If you are just opening a browser on a URL then you can just use `Process.Start(url)`, which will do the same thing and much simpler, including the opening the default browser registered by the user. We use this all over the extension.

GoogleCloudExtension/GoogleCloudExtension/GoogleCloudExtensionPackage.cs
```diff
@@ -257,9 +259,17 @@ protected override void Initialize()
 
         #region User Settings
 
-        public AnalyticsOptionsPage AnalyticsSettings => GetDialogPage<AnalyticsOptionsPage>();
+        public AnalyticsOptions AnalyticsSettings => GetDialogPage<AnalyticsOptions>();
 
-        public T GetDialogPage<T>() where T : DialogPage => (T) GetDialogPage(typeof(T));
+        public T GetDialogPage<T>() where T : DialogPage
```
ivannaranjo: Method comments.

GoogleCloudExtension/GoogleCloudExtension/Analytics/EventsReporterWrapper.cs
ILMTitan: I'm not sure where it is getting the code on the left here. The code on the right is what exists in both branches, master issue#562.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptions.cs
ILMTitan: I am not sure what you think is being broken.
I am also not sure why you think clearing and then adding back to the `ObservableCollection` is less expensive. Just setting the property to a new collection will trigger a single event (O(1)), which triggers the UI to update (a O(n) operation). Mutating the collection, however, triggers many events. First, an event when the collection is cleared. Then because it doesn't have an `AddRange`, each item would have to be added individually. That means a new event on each add (O(n)). Each of these events could trigger a UI update (O(n)), for a total cost of O(n<sup>2</sup>).

GoogleCloudExtension/GoogleCloudExtension/CloudExplorer/Options/CloudExplorerOptionsPageViewModel.cs
ILMTitan: While it could be a readonly property, I'm not sure where you are getting that it should be, or even typically is. Doing so would avoid the need for a property changed notification mechanism, but given the [Docs for `ObservableCollection`](https://msdn.microsoft.com/en-us/library/ms668604(v=vs.110).aspx#Anchor_8) has a note about using the `INotifyPropertyChanged` interface, it seems pretty standard to use that mechanism as well.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ILMTitan: Doing this allows me to mock and register the `IVsWebBrowsingService` in unit tests. It allows me to avoid having an ugly `internal Action<string> StartProcess = Process.Start` variable, and is more idiomatic to Visual Studio.

GoogleCloudExtension/GoogleCloudExtension/GoogleCloudExtensionPackage.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.DataSources/PubSubDataSource.cs
ILMTitan: Done.

ILMTitan:
PTAL.

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
ivannaranjo: Please don't, use the same way of opening browsers as the rest of the code. If you want to mock the way the browser is opened do it the existing way, it is not ugly, it is a mock.

ILMTitan: Changed.

ILMTitan:
Done.
PTAL.

Issue #787, opened 2017-08-23
Please fill the local path automatically with something sensible.


SurferJeffAtGoogle:
![image](https://user-images.githubusercontent.com/11778530/29639427-970022de-880f-11e7-8ff4-5c708ffff3e8.png)

ivannaranjo:
Can you take a look at this issue? Does this make sense to fix?

ILMTitan:
This should probably be initialized to `string initialLocalPath = Path.Join(DTE.Properties("Environment", "ProjectsAndSolution")["ProjectsLocation"].Value, repositoryName);`

Deren-Liao:
This is to set git root path.  As I looked at other git plug in,  the best would be to use previous git root location.  Or  use a default git root .

ILMTitan:
It is the git root path, but it is also most likely a Visual Studio Solution or Project folder. Therefore, we should default to a new folder under the location specified under Tools > Options > Projects and Solutions > Locations > Projects location, which is the default location when creating a new project from a template as well. The above code gets that option value.

Deren-Liao:
Normally, people sync git code to different folder other than VS solution default folder. At least that's my case. I'll default it to %userprofile%\csr  folder.

PR #859, opened 2017-12-22
Set a default clone local path

Deren-Liao:
Fix issue #787

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
@@ -39,12 +39,14 @@ namespace GoogleCloudExtension.CloudSourceRepositories
     public class CsrCloneWindowViewModel : ValidatingViewModelBase
     {
         internal static Func<string, IApiManager> s_getApiManagerFunc = ApiManager.GetApiManager;
+        internal static readonly string s_defaultLocalPath =
+            Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "csr");
         private static readonly List<string> s_requiredApis =
             new List<string> { KnownApis.CloudSourceRepositoryApiName };
 
         private readonly Action _closeOwnerFunc;
         private readonly HashSet<string> _newReposList = new HashSet<string>();
-        private string _localPath;
+        private string _localPath = s_defaultLocalPath;
         private Repo _latestCreatedRepo;
         private Repo _selectedRepo;
         private IEnumerable<Project> _projects;
@@ -213,6 +215,11 @@ namespace GoogleCloudExtension.CloudSourceRepositories
 
         private async Task CloneAsync()
         {
+            if (IsDefaultLocation(LocalPath) && !Directory.Exists(s_defaultLocalPath))
+            {
+                Directory.CreateDirectory(s_defaultLocalPath);
+            }
+
             // If OkCommand is enabled, SelectedRepository and LocalPath is valid
             string destPath = Path.Combine(LocalPath.Trim(), SelectedRepository.GetRepoName());
 
@@ -301,7 +308,7 @@ namespace GoogleCloudExtension.CloudSourceRepositories
                     nameof(Resources.ValdiationNotEmptyMessage), fieldName);
                 yield break;
             }
-            if (!Directory.Exists(localPath))
+            if (!IsDefaultLocation(localPath) && !Directory.Exists(localPath))
             {
                 yield return StringValidationResult.FromResource(nameof(Resources.CsrClonePathNotExistMessage));
                 yield break;
@@ -329,5 +336,8 @@ namespace GoogleCloudExtension.CloudSourceRepositories
                     () => RepositoriesAsync.StartListRepoTaskAsync(_selectedProject.ProjectId));
             }
         }
+
+        private bool IsDefaultLocation(string localPath) => 
+            String.Equals(localPath.Trim(), s_defaultLocalPath, StringComparison.OrdinalIgnoreCase);
     }
 }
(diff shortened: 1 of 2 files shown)


GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -39,12 +39,14 @@ namespace GoogleCloudExtension.CloudSourceRepositories
     public class CsrCloneWindowViewModel : ValidatingViewModelBase
     {
         internal static Func<string, IApiManager> s_getApiManagerFunc = ApiManager.GetApiManager;
+        internal static readonly string s_defaultLocalPath =
+            Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "csr");
```
ivannaranjo: I would suggest using the full name of the product, something like _Cloud Source Repositories_ so the user has a clearer idea of what the directory is. Also instead of using the `UserProfile` I would suggest using the `Documents` directory as the parent directory, just like Visual Studio does for the projects. The user's profile directory is not easily browseable from the Windows explorer.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -39,12 +39,14 @@ namespace GoogleCloudExtension.CloudSourceRepositories
...
         private static readonly List<string> s_requiredApis =
             new List<string> { KnownApis.CloudSourceRepositoryApiName };
 
         private readonly Action _closeOwnerFunc;
         private readonly HashSet<string> _newReposList = new HashSet<string>();
-        private string _localPath;
+        private string _localPath = s_defaultLocalPath;
```
ivannaranjo: Looking at this I don't think you need the `s_defaultLocalPath` actually, calculating the default path is not going to be that expensive that you need to save the `static` variable. Can simplify the code a bit.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -213,6 +215,11 @@ private void RepositoriesAsyncPropertyChanged(object sender, PropertyChangedEven
 
         private async Task CloneAsync()
         {
+            if (IsDefaultLocation(LocalPath) && !Directory.Exists(s_defaultLocalPath))
```
ivannaranjo: Why is it important that the `LocalPath` is the the default path here? Shouldn't you create the path regardless?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -329,5 +336,8 @@ private void OpenCreateRepoDialog()
                     () => RepositoriesAsync.StartListRepoTaskAsync(_selectedProject.ProjectId));
             }
         }
+
+        private bool IsDefaultLocation(string localPath) => 
+            String.Equals(localPath.Trim(), s_defaultLocalPath, StringComparison.OrdinalIgnoreCase);
```
ivannaranjo: You can't trim spaces in the local path, spaces are significant in the path.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Normally, we only create the sub folder .  For example, the root is c:\my-path\git   then we create the sub folder for the repository ,  like c:\my-path\git\csr-test-1 .    We instead do verification to test if the root folder exists, if not, an prompt is shown.   There are two reasons I don't create root folder  because  root folder can be a long path, or any path at external storage etc, creating such a path is likely to fail or takes too long time etc.   But for default folder,  we know it's safe and is likely to succeed to create such a folder if it does not exist.

Plus, for default folder, we don't want to show an error "the path does not exist".

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: the s_defaultLocalPath is used in multiple places .

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Done

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
Deren-Liao: Done

Deren-Liao:
Please take a look again

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
```diff
@@ -39,8 +39,8 @@ namespace GoogleCloudExtension.CloudSourceRepositories
     public class CsrCloneWindowViewModel : ValidatingViewModelBase
     {
         internal static Func<string, IApiManager> s_getApiManagerFunc = ApiManager.GetApiManager;
-        internal static readonly string s_defaultLocalPath =
-            Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "csr");
+        internal static string s_defaultLocalPath = Path.Combine(
```
ivannaranjo: Why remove the `readonly` attribute?

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrCloneWindowViewModel.cs
ivannaranjo: I don't agree, you should create the directory hierarchy if the user requests it, but that should probably be a separate bug to this.

Deren-Liao:
Thanks for doing the review

Issue #874, opened 2018-02-13
Text in the Stackdriver Logs date picker says "get log"


iantalarico:
Should say "get logs"

PR #885, opened 2018-02-13
Fix for #874, change text on Stackdriver Logs date picker.

amanda-tarafa:
Change text from "Get log" to "Get logs"


ILMTitan:
Fixes #874

Issue #895, opened 2018-03-06
NullReferenceException when there's no open solution and thus no StartupProject.


amanda-tarafa:
This happens when trying to access the main Google Cloud menu with no open solution.
This error was introduced by changes done by me on #891 when merging ProjectHelper and IParsedProject.
Will fix it and add unit tests to catch this kind of thing better.

amanda-tarafa:
Bug fixed with #896 .
Keeping this one open to add tests.

amanda-tarafa:
Adding the original labels, and closing. Created #905 to add tests and refactor.

PR #896, opened 2018-03-06
Fixing #895. Just adding a null check before accessing ProjectHelper.ParsedProject

amanda-tarafa:
Quick PR so as to avoid this bug interfering with everyone else's work. Will add tests at a later PR.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/GenerateConfigurationCommand/GenerateConfigurationContextMenuCommand.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/GenerateConfigurationCommand/GenerateConfigurationContextMenuCommand.cs
@@ -160,7 +160,7 @@ namespace GoogleCloudExtension.GenerateConfigurationCommand
             }
 
             // Ensure that the menu entry is only available for ASP.NET Core projects.
-            var selectedProject = SolutionHelper.CurrentSolution.SelectedProject.ParsedProject;
+            var selectedProject = SolutionHelper.CurrentSolution.SelectedProject?.ParsedProject;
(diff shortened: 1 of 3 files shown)


GoogleCloudExtension/GoogleCloudExtension/GenerateConfigurationCommand/GenerateConfigurationContextMenuCommand.cs
```diff
@@ -160,7 +160,7 @@ private void OnBeforeQueryStatus(object sender, EventArgs e)
             }
 
             // Ensure that the menu entry is only available for ASP.NET Core projects.
-            var selectedProject = SolutionHelper.CurrentSolution.SelectedProject.ParsedProject;
+            var selectedProject = SolutionHelper.CurrentSolution.SelectedProject?.ParsedProject;
```
iantalarico: Can ```CurrentSolution``` ever be null?

amanda-tarafa: No, I checked by running VS, even if there's no solution open DTE returns some sort of empty solution.
Also, before my previous PR the code was exactly as is now execpt for the call to ParsedProject:
var selectedProject = SolutionHelper.CurrentSolution.SelectedProject;
and current solution being null would have been a problem there if it was possible.
I can add the null check just to be on the super safe side.

iantalarico: No this should be fine.  We can double check this all with the tests when we get them written.

iantalarico:
Thanks for the quick fix!

PR #908, opened 2018-03-14
Remove the "Show only Windows instances" button

### LINKED ISSUE #882: "Show only Windows instances" button on top of the extension is confusing
The "Show only Windows instances" button on the top of the extension is confusing.  I think this was from when we had less resources.  We should reconsider this.

iantalarico:
This is confusing now that the extension has more resources and it only affects one of the resources listed.

Fixes #882


iantalarico:
Closing because CLA confusion (was planning to squash and merge)

Issue #906, opened 2018-03-12
Finding flaky tests on 2018-q1-master


amanda-tarafa:
The build has been failing on and off for a few days now. We probably have some flaky tests. The tests don't fail apparently, the build just hangs. Investigating.

amanda-tarafa:
After a couple of days of testing on my fork here are my conclusions:

Testing is flaky because we effectively have some `async void` tests, which means that we effectively have some asynchronous tests. These sometimes make vstest crash.

Tests reference in #898 are some of those effectively `async void` tests. Even though they are not defined as `async void`, they are testing 
`LogsViewerViewModel.SubmitAdvanceFilterCommand.Execute`
and
`LogsViewerViewModel.SimpleTextSearchCommand.Execute`
which in turn are calling
`ErrorHandlerUtils.HandleAsyncExceptions(ReloadAsync)`
which is defined as
`public static async void HandleAsyncExceptions(Func<Task> task)`
i.e., `async void`.
This effectively makes the calls to `LogsViewerViewModel.SubmitAdvanceFilterCommand.Execute` and `LogsViewerViewModel.SimpleTextSearchCommand.Execute`. And probably the fact that these two tests need two identical calls to `Execute` is just prove that these calls are asynchronous, the second call to `Execute` is only improving the chances of the first one having finished thus allowing the `Asserts` to pass. We probably have several other tests like these, given that several Commands are calling `ErrorHandlerUtils.HandleAsyncExceptions(ReloadAsync)`, and maybe other `async void` methods.

`async void` tests [are not considered good tests](https://msdn.microsoft.com/en-us/magazine/dn818493.aspx) since they are effectively fire and forget (we don't have a Task to keep track of the execution). And as far as I can tell by going through [vstest docs](https://github.com/Microsoft/vstest-docs) and [vstest code](https://github.com/Microsoft/vstest), it does not support `async void` test.

What I've done:
In #907 I've added code to print the vstest diagnostics file whenever vstest crashes. This will hopefully allow us to see why it crashes exactly every time. Curiously, just by generating the diagnostics file, flakiness has stopped, just because by making vstest write diagnostics info we are probably giving task enough time to finish without tests conflicting with one another.

Solutions:
First we need to identify all of our `async void` tests.
And then remove them.
[...]

PR #910, opened 2018-03-19
Improving diagnostics info we get, specially when vstest crashes.

amanda-tarafa:
Working towards #906.

- Configuring tests to run in isolation, we get more diagnostics info that way.

- Pushing dump files (and pdbs and dlls) as artifacts when vstest crashes for better diagnostics info.


amanda-tarafa:
Yes, it almost always crashes locally on my laptop, not on my VM though. I already know what is making it crash, I don't know exactly why that happens although I have some suspicions and have ran some scenarios that validate those suspicions. I'm writing a doc with all my findings and the evidence for those findings, possible next steps, and at least a couple other instances of flakiness I've come across. Will share it as soon as I got it.

iantalarico:
Awesome, thanks!

Issue #917, opened 2018-03-22
Test dependency and mocking of global resources.


amanda-tarafa:
Review usage, test dependency and mocking of global resources in general. This is one of the ways in which we could introduce flakiness more inadvertently, and given that several of the global resources are unmanaged, these problems would be hard to diagnose, they don’t make tests fail, but vstest.console.exe crash.
Spawns from #906.

amanda-tarafa:
For dependencies on Package:

1. Consider creating a `TestBase` class that always mocks `GoogleCloudExtensionPackage.Instance` without setting up any methods but with strict behavior so that it is clear what each test is dependent on.
2. Have helper methods on `TestBase` to set up or mock the `GoogleCloudExtensionPackage` methods that are needed, so that tests can explicitly setup what they need.

Take a look at #929 .

ILMTitan:
Closed in #934

Issue #916, opened 2018-03-22
Undercover async void tests methods.


amanda-tarafa:
In general async void tests should be reviewed and fixed. Even though we might not be noticing any related problems right now, the more tests we add, the more chances there’ll be for one test stepping over another. Some async void methods called from tests can be easily changed to async Task. Some others, as per @jonskeet suggestion, can be turned into async Task for use from within tests, and wrap around and async void to keep compatibility with the current code. Some issues we might find when doing this:
Some of the async void methods that are being called from tests are nested two or three levels into non async methods, so a lot of refactoring would be needed.
UI commands wouldn’t be testable through their Execute method, since that is currently void, but we could also make the Execute of a command a thin wrapper around async Task methods and test those.
All of this would need a lot of refactoring and unit testing before the refactoring to make sure that nothing is broken.
This one spawns from #906

amanda-tarafa:
As part of solving this consider adding a `Task` or an `AsyncProperty` to `ProtectedAsyncCommand` and wrapping all `async` commands with that class instead of with `ProtectedCommand`.
Revisit #929 and its comments to improve with `ProtectedAsyncCommand`.

Issue #914, opened 2018-03-22
PubSubSourceRootViewModelTests don't pass on their own.


amanda-tarafa:
Tests on [GoogleCloudExtensionUnitTests.CloudExplorerSources.PubSub.PubSubSourceRootViewModelTests](https://github.com/GoogleCloudPlatform/google-cloud-visualstudio/blob/8c320c471ce8bf48ed28a759ce474bfffee0974e/GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudExplorerSources/PubSub/PubSubSourceRootViewModelTests.cs#L39) pass when run with the rest of tests but fail if run by themselves. Definitely some other tests are helping these ones pass.

This one stems from #906

amanda-tarafa:
Fixed in #929

PR #929, opened 2018-04-11
Fixing flakyness in PubSubSourceRootViewModelTests.

amanda-tarafa:
Fixing #914 by fixing some of #917 and #916 .

These tests started to randomly fail on my laptop while working on another PR. Given that we already knew they were flaky (#914) I decided to fix them first before continuing the other PR.

This can still be improved, and when tackling #917 and #916 a more reusable solution can be achieved, but probably this PR will be a good starting point for that.


amanda-tarafa:
PTAL

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudExplorerSources/PubSub/PubSubSourceRootViewModelTests.cs
```diff
@@ -343,7 +358,7 @@ public void TestNewTopicCommandError()
                 return true;
             };
 
-            _objectUnderTest.OnNewTopicCommand();
+            await _objectUnderTest.OnNewTopicCommandAsync();
```
ILMTitan: I think I was wrong to test against an internal method to begin with. We should test against the public API of the class.

    _objectUnderTest.ContextMenu.ItemsSource.OfType<MenuItem>()
        .Single(mi => mi.Header.Equals(Resources.CloudExplorerPubSubNewTopicMenuHeader)).Command.Execute(null);
    await _objectUnderTest.OnNewTopicCommandExecution.Task;

It is ugly for now, but that just means we should consider refactoring SourceRootViewModelBase.

A compromise would be to add the `MenuItem` or `Command` itself as a field of the `PubSubSourceRootViewModel` and test using that.

GoogleCloudExtension/GoogleCloudExtension/IGoogleCloudExtensionPackage.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+using System;
+using System.ComponentModel.Design;
+
+namespace GoogleCloudExtension
+{
+    /// <summary>
+    /// Interface extracted from <see cref="GoogleCloudExtensionPackage"/> mainly for testing purposes.
+    /// </summary>
+    public interface IGoogleCloudExtensionPackage : IVsPackage, Microsoft.VisualStudio.OLE.Interop.IServiceProvider, IOleCommandTarget, IVsPersistSolutionOpts, IServiceContainer, System.IServiceProvider, IVsUserSettings, IVsUserSettingsMigration, IVsUserSettingsQuery, IVsToolWindowFactory, IVsToolboxItemProvider
```
ILMTitan: I was going to suggest creating a series of interfaces to match the class hierarchy of `Package`, but all these interfaces are directly applied to `Package`, so this is fine.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudExplorerSources/PubSub/PubSubSourceRootViewModelTests.cs
```diff
@@ -101,6 +93,15 @@ public void BeforeEach()
             _contextMock.Setup(c => c.CurrentProject.ProjectId).Returns(MockProjectId);
             _factoryMock = new Mock<Func<IPubsubDataSource>>();
             _factoryMock.Setup(f => f()).Returns(() => _dataSourceMock.Object);
+
+            _gcePackageMock = new Mock<IGoogleCloudExtensionPackage>(MockBehavior.Strict);
+            _cloudExplorerOptionsMock = new Mock<CloudExplorerOptions>(MockBehavior.Strict) { CallBase = false };
```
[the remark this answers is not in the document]
amanda-tarafa: Removed

GoogleCloudExtension/GoogleCloudExtension/CloudExplorerSources/PubSub/PubSubSourceRootViewModel.cs
```diff
@@ -231,10 +230,18 @@ internal void OnOpenCloudConsoleCommand()
             OpenBrowser(url);
         }
 
+        /// <summary>
+        /// Wraper on <see cref="OnNewTopicCommandAsync"/> for command execution.
+        /// </summary>
+        private async void OnNewTopicCommand()
```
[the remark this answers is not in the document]
amanda-tarafa: Yes, I agree with you. 
I was actually thinking, given that we have a ProtectedAsyncCommand class, that we should put the AsyncProperty in that class, representing the execution of the command wrapped, so we can uniformly do this with all of our async commands. I won't do that now but will leave a note on #916 to that effect.

I will leave this one as is for now, and will revisit when working on #916.

GoogleCloudExtension/GoogleCloudExtensionUnitTests/CloudExplorerSources/PubSub/PubSubSourceRootViewModelTests.cs
amanda-tarafa: Yes, I agree with you.

But your proposed changed depends on implementing your previous suggestion about how to wait for an async command (see also my comments there). When I take on #916 I will revisit and change to what you are proposing or very similar.

As for how to get to the MenuItem, yes it's ugly, but tbh I think it might be the right way to go as everything stands  except for refactoring as you say. My second best choice would be to add the MenuItem(s) (there are three of them) directly to the PubSubSourceRootViewModel but this would only be for testing purposes.

Leaving this as is for now until #916 is tackled. How to get to the MenuItem might be outside of the scope of #916 so might go with the 'ugly' proposed solution at that point.

GoogleCloudExtension/GoogleCloudExtension/IGoogleCloudExtensionPackage.cs
amanda-tarafa: As a side note to this, there are several methods directly defined in Package and some others directly defined in our GoogleCloudExtensionPackage which are the ones that we should define in IGoogleCloudExtensionPackage interface. I just added the ones I needed for this PR but we can add the others as needed.
I'm adding some notes in #917 the might be helpful.

amanda-tarafa:
Thanks @ILMTitan .

PR #931, opened 2018-04-11
Update Publish choice dialog tool tips text.

### LINKED ISSUE #867: When publishing to GCP, the tool should provide brief details on how different publish targets differ
When publishing to GCP, you can publish to GKE, AppEngine, etc. There should be (at least) tooltip for each button representing publish target that briefly explains why one should choose that option.

I created very basic webapp in C# that exposed 4 APIs via 1 controller, tested it locally, then tried to deploy it to GCP. When you click on “Publish to Google Cloud…”, you get 3 choices:
App Engine Flex
Container Engine
Compute Engine (disabled)

There is no assistance about these options so if you are not familiar with GCP, you may have trouble choosing one.

ILMTitan:
Enable the tool tips of disabled buttons.

Fixes #867

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Resources.resx
+++ b/GoogleCloudExtension/GoogleCloudExtension/Resources.resx
@@ -826,7 +826,7 @@
     <comment>The name of the GCE choice.</comment>
   </data>
   <data name="PublishDialogChoiceStepGceToolTip" xml:space="preserve">
-    <value>Deploy your .NET Framework application to a Windows virtual machine in Google Compute Engine.</value>
+    <value>Deploy your Classic ASP.NET application to a Windows virtual machine in Google Compute Engine.</value>
     <comment>The tooltip for the GCE button.</comment>
   </data>
   <data name="PublishDialogChoiceStepMessage" xml:space="preserve">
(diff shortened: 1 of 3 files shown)


GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -818,15 +818,15 @@
...
+    <value>App Engine is the preferred choice for quickly setting up your .NET Core application.</value>
     <comment>The tooltip for the App Engine Flex button.</comment>
   </data>
   <data name="PublishDialogChoiceStepGceName" xml:space="preserve">
     <value>_Compute Engine</value>
     <comment>The name of the GCE choice.</comment>
   </data>
   <data name="PublishDialogChoiceStepGceToolTip" xml:space="preserve">
-    <value>Deploy to Compute Engine</value>
+    <value>Deploy your .NET Framework application to a Windows virtual machine in Google Compute Engine.</value>
```
amanda-tarafa: Asp .NET, since a lot of the questions we are getting are around why we can't publish Core in GCP?

iantalarico: +1 to this. I think it may be explicitly mentioning in all these cases what we support to avoid confusion.

ILMTitan: I am concerned people will still confuse ASP.NET with ASP.NET Core. Perhaps I should say "Classic ASP.NET application"?

iantalarico: I think Classic ASP.NET works

PR #958, opened 2018-05-11
Don't fail silently when windows credentials go wrong.

### LINKED ISSUE #732: Add notification for corrupt Windows credentials files.
We should notify the user that there's a corruption problem with the Windows credentials files, for Q2 we will silently delete bad files.

ILMTitan:
Inform the user of issues so they have a chance to fix them.
Fixes #732

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
@@ -15,9 +15,9 @@
 using Google.Apis.Compute.v1.Data;
 using GoogleCloudExtension.DataSources;
 using GoogleCloudExtension.GCloud;
+using GoogleCloudExtension.Utils;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
@@ -146,22 +146,28 @@ namespace GoogleCloudExtension.Accounts
 
                 return new WindowsInstanceCredentials { User = userName, Password = Encoding.UTF8.GetString(passwordBytes) };
             }
-            catch (CryptographicException)
+            catch (CryptographicException cryptographicException)
             {
-                Debug.WriteLine($"Failed to decrypt credentials from: {path}");
+                UserPromptUtils.ErrorPrompt(
+                    string.Format(Resources.WindowsCredentialsStore_DecryptErrorMessage, path),
+                    Resources.WindowsCredentialsStore_DecryptionErrorTitle, cryptographicException.ToString());
                 try
                 {
                     File.Delete(path);
                 }
-                catch (IOException)
+                catch (IOException ioException)
                 {
-                    Debug.WriteLine($"Failed cleaning corrupted credentials {path}");
+                    UserPromptUtils.ErrorPrompt(
+                        string.Format(Resources.WindowsCredentialsStore_DeletingCorruptedErrorMessage, path),
+                        Resources.WindowsCredentialsStore_DeletingCorruptedErrorTitle, ioException.ToString());
                 }
                 return null;
             }
-            catch (IOException)
+            catch (IOException e)
             {
-                Debug.WriteLine($"Failed to load credentials from: {path}");
+                UserPromptUtils.ErrorPrompt(
+                    string.Format(Resources.WindowsCredentialsStore_CredentialFileLoadErrorMessage, path),
+                    Resources.WindowsCredentialsStore_CredentialFileLoadErrorTitle, e.ToString());
                 return null;
             }
         }
--- a/GoogleCloudExtension/GoogleCloudExtension/Resources.Designer.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Resources.Designer.cs
@@ -19,7 +19,7 @@ namespace GoogleCloudExtension {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
     [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
     public class Resources {
@@ -5191,8 +5191,8 @@ namespace GoogleCloudExtension {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Could not find a 1.0.0-preview version of .Net Core SDK.
-        ///Either intall the Visual Studio 2015 DotNetCore tooling, or upgrade to Visual Studio 2017..
+        ///   Looks up a localized string similar to Could not find a 1.0.0-preview version of .NET Core SDK.
+        ///Either intall the .NET Core tools for Visual Studio 2015, or upgrade to Visual Studio 2017..
         /// </summary>
         public static string NoDotNetCorePreviewFoundError {
             get {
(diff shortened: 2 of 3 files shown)


GoogleCloudExtension/GoogleCloudExtension.GCloud.UnitTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,20 @@
...
+[assembly: AssemblyProduct("GoogleCloudExtension.GCloud.UnitTests")]
+[assembly: AssemblyCopyright("Copyright ©  2018")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+[assembly: ComVisible(false)]
+
+[assembly: Guid("608977ee-0606-45f6-b68c-4861842b6fe3")]
+
+// [assembly: AssemblyVersion("1.0.*")]
```
iantalarico: Remove?

GoogleCloudExtension/GoogleCloudExtension.GCloud.UnitTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,20 @@
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+[assembly: AssemblyTitle("GoogleCloudExtension.GCloud.UnitTests")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("GoogleCloudExtension.GCloud.UnitTests")]
+[assembly: AssemblyCopyright("Copyright ©  2018")]
```
iantalarico: Should this have Google somewhere in it?

GoogleCloudExtension/GoogleCloudExtension.GCloud.UnitTests/WindowsInstanceCredentialsTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        private static object[][] NotEqualObjects { get; } =
+        {
+            new object[] {null},
+            new object[] {"UserNameAndPassword"},
+            new object[] {new WindowsInstanceCredentials("DifferentUser", "Password")},
+            new object[] {new WindowsInstanceCredentials("User", "DifferentPassword")}
+        };
+
+        [TestMethod]
+        public void TestJsonSerialization()
```
iantalarico: Should this and the below be one test?  

I don't think we care so much about the interim but the result.  We can encode and decode a `WindowsInstanceCredentials` object

GoogleCloudExtension/GoogleCloudExtension.GCloud/WindowsInstanceCredentials.cs
```diff
@@ -22,9 +22,46 @@ namespace GoogleCloudExtension.GCloud
     public sealed class WindowsInstanceCredentials
     {
         [JsonProperty("username")]
-        public string User { get; set; }
+        public string User { get; }
 
         [JsonProperty("password")]
-        public string Password { get; set; }
+        public string Password { get; }
```
iantalarico: 👍 to these changes

GoogleCloudExtension/GoogleCloudExtension.GCloud/WindowsInstanceCredentials.cs
```diff
@@ -22,9 +22,46 @@ namespace GoogleCloudExtension.GCloud
...
+
+        [JsonConstructor]
+        public WindowsInstanceCredentials(string user, string password)
+        {
+            User = user;
```
iantalarico: Do we need to check for null or empty string?

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -40,42 +41,54 @@ internal class WindowsCredentialsStore
...
         /// In memory cache of the credentials for the current credentials (account and project pair).
         /// </summary>
-        private readonly Dictionary<string, IEnumerable<WindowsInstanceCredentials>> _credentialsForInstance = new Dictionary<string, IEnumerable<WindowsInstanceCredentials>>();
+        private readonly Dictionary<string, IEnumerable<WindowsInstanceCredentials>> _credentialsForInstance =
+            new Dictionary<string, IEnumerable<WindowsInstanceCredentials>>();
 
         public static WindowsCredentialsStore Default => s_defaultStore.Value;
 
+        /// Mockable static methods for testing.
+        internal Func<byte[], byte[], DataProtectionScope, byte[]> Unprotect { private get; set; } =
```
iantalarico: Do we need this way?  It seems like we could fake a file instead?

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
```diff
@@ -140,54 +153,60 @@ private WindowsInstanceCredentials LoadEncryptedCredentials(string path)
...
+            catch (CryptographicException cryptographicException)
             {
-                Debug.WriteLine($"Failed to decrypt credentials from: {path}");
+                UserPromptUtils.ErrorPrompt(
+                    string.Format(Resources.WindowsCredentialsStoreDecryptErrorMessage, path),
+                    Resources.WindowsCredentialsStoreDecryptionErrorTitle, cryptographicException.ToString());
                 try
                 {
-                    File.Delete(path);
+                    DeleteFile(path);
```
iantalarico: Should we give the user an option to not delete the file?

GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
```diff
@@ -29,6 +29,12 @@ internal static class UserPromptUtils
         private static readonly Lazy<ImageSource> s_warningIcon = new Lazy<ImageSource>(() => ResourceUtils.LoadImage(WarningIconPath));
         private static readonly Lazy<ImageSource> s_errorIcon = new Lazy<ImageSource>(() => ResourceUtils.LoadImage(ErrorIconPath));
 
+        // Mockable static method for testing.
+        internal static Func<UserPromptWindow.Options, bool> PromptUserOverride { private get; set; } = null;
```
iantalarico: Same with these

GoogleCloudExtension/GoogleCloudExtensionUnitTests/Accounts/WindowsCredentialsStoreTest.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using Project = Google.Apis.CloudResourceManager.v1.Data.Project;
+
+namespace GoogleCloudExtensionUnitTests.Accounts
+{
+    /// <summary>
+    /// Summary description for WindowsCredentialsStoreTest
```
iantalarico: Should this comment be updated?

GoogleCloudExtension/GoogleCloudExtension.GCloud.UnitTests/WindowsInstanceCredentialsTests.cs
ILMTitan: I want to make sure the class can still decode files already stored on user computers after changes to the class.

GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
ILMTitan: I don't see a better way to do this here without a larger refactoring to use non-static methods.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ILMTitan: I got rid of the Unprotect/Protect mocks. That was a case of being too mock-happy. I'm not sure using the actual file system would be easier than the file system mock, but perhaps https://github.com/tathamoddie/System.IO.Abstractions would be.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension.GCloud/WindowsInstanceCredentials.cs
ILMTitan: Updated to check user against null. There are cases where password is null.

ILMTitan:
PTAL

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
iantalarico: I don't mean an fake (like a mock) of the file system.  I just mean let the test create a file.

It just seems like a code specifically for testing is getting put in.

ILMTitan: I prefer to mock the file system for a few reasons:

1. It is easier to simulate failures: just tell the mock setup `.Throws<IOException>()`.
2. You don't have to clean up mocks like you would actual files.
3. Directory.Exists and Directory.EnumerateFiles are easier to setup as mocks than as actual files.
4. You don't have to inject special test file paths with mocks.
    * To use actual files in this case, I would have to make `s_credentialsStoreRoot` editable so the tests would hit a "safe" location, rather than a real one.

Drawbacks are:

1. You have to put in code specifically for testing.
    * So long as this doesn't hurt code readability, I don't think this is a big problem.
2. Tests can have many mock setups, making them difficult to read.
    * The most complicated test here has 5 mock setups, approaching difficult to read territory, but I'm not sure setting up actual files would be much more readable.

GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
iantalarico: I think it's something we should think about in the long run.  I think in general when we start doing things like this we need to look at our overall design as something may be wrong.

GoogleCloudExtension/GoogleCloudExtension/Accounts/WindowsCredentialsStore.cs
iantalarico: That sounds reasonable as I mentioned below I think adding things like this may mean we want to take a careful look at our design in the future.

iantalarico: In general I think the less we mock the better.  Mocking adds more room for error (but as you mention, does make testing errors harder in some cases)

GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
```diff
@@ -103,6 +103,26 @@ public static void ErrorPrompt(string message, string title, string errorDetails
                 });
         }
 
+        /// <summary>
+        /// Shows an error message dialog to the user, with an Ok and a Cancel button.
+        /// </summary>
+        /// <param name="message">The message for the dialog.</param>
+        /// <param name="title">The title for the dialog.</param>
+        /// <param name="errorDetails">The error details for the dialog, optional.</param>
+        public static bool ErrorActionPrompt(string message, string title, string errorDetails = null)
```
iantalarico: nit: add note about return value?

GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
```diff
@@ -103,6 +103,26 @@ public static void ErrorPrompt(string message, string title, string errorDetails
...
+        {
+            return PromptUser(
```
iantalarico: suggestion: 
...= > PromptUser(...

GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/Utils/UserPromptUtils.cs
ILMTitan: Done.

PR #959, opened 2018-05-15
Ensure the default deployment name is a valid name.

### LINKED ISSUE #798: Generate appropriate name for GKE service.
We should generate an appropriate name for a GKE deployment and service based on the project's name. If the user specifies a custom name, we should remember that as well.

ILMTitan:
Fixes #798

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtensionUnitTests/Utils/GcpPublishStepsUtilsTests.cs
+++ b/GoogleCloudExtension/GoogleCloudExtensionUnitTests/Utils/GcpPublishStepsUtilsTests.cs
@@ -25,7 +25,7 @@ namespace GoogleCloudExtensionUnitTests.Utils
         [DataRow(null)]
         [DataRow("")]
         [DataRow("already-kebob-case")]
-        [DataRow("end-with-dash")]
+        [DataRow("end-with-dash-")]
         public void TestToValid_NameUnchanged(string unchangingName)
         {
             string result = GcpPublishStepsUtils.ToValidName(unchangingName);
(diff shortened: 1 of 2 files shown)


GoogleCloudExtension/GoogleCloudExtensionUnitTests/Utils/GcpPublishStepsUtilsTests.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace GoogleCloudExtensionUnitTests.Utils
+{
+    [TestClass]
+    public class GcpPublishStepsUtilsTests
+    {
+        [TestMethod]
+        [DataRow(null)]
+        [DataRow("")]
+        [DataRow("already-kebob-case")]
+        [DataRow("end-with-dash")]
```
redborian: Good use of DataRows 👍

GoogleCloudExtension/GoogleCloudExtensionUnitTests/Utils/StringUtilTests.cs
```diff
@@ -58,5 +58,33 @@ public void LastNonSpaceIndexTests()
...
+
+        [TestMethod]
+        [DataRow("ALLUPPERCASE", "alluppercase")]
+        [DataRow("UpperCamelCase", "upper-camel-case")]
+        [DataRow("lowerCamelCase", "lower-camel-case")]
+        [DataRow("Upper-Kebob-Case", "upper-kebob-case")]
+        [DataRow("20number2", "20-number-2")]
+        [DataRow("UpperCamelCaseWith2Numbers100", "upper-camel-case-with-2-numbers-100")]
+        [DataRow("CamelCase&Symbols!", "camel-case&symbols!")]
+        public void TestToKebobCase_Changes(string argument, string expectedResult)
```
redborian: Need to add a test for the extreme case of name's length > 100

ILMTitan: Because there is neither a method comment/contract nor implementation about very long arguments, I don't think the test is strictly necessary, but I did add it.

PR #960, opened 2018-05-15
Enable refreshing instances from within the GCE publish dialog.

### LINKED ISSUE #938: On the GCE publish dialog, add a link to refresh instances.
When publishing to GCE, the user either selects an existing instance or might want to create a new one. For this, the user goes to the Web Cloud Console, which is OK, but we don't have a straightforward mechanism for the user to refresh the dialog with the instances they have created, they need to close the dialog an open it again or reselect the project from the project selector, which is not intuitive.

image

ILMTitan:
Add GceStepViewModel.RefreshInstancesCommand with tests.
Add refresh instances button to GceStepContent.
Remove Next button and related members, as it was never enabled.
Fixes #938

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Resources.resx
+++ b/GoogleCloudExtension/GoogleCloudExtension/Resources.resx
@@ -845,10 +845,6 @@
     <value>_VM instance:</value>
     <comment>The message shown to select an instance.</comment>
   </data>
-  <data name="PublishDialogNextButtonCaption" xml:space="preserve">
-    <value>_Next</value>
-    <comment>The caption for the next button.</comment>
-  </data>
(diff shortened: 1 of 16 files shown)


GoogleCloudExtension/GoogleCloudExtension/Resources.resx
```diff
@@ -845,10 +845,6 @@
     <value>_VM instance:</value>
     <comment>The message shown to select an instance.</comment>
   </data>
-  <data name="PublishDialogNextButtonCaption" xml:space="preserve">
```
redborian: It is good to clean up the Resources.resx 👍

ILMTitan: We actually have a script that will fail the build when there are unused resources.

PR #973, opened 2018-05-30
Refactor publish dialog.

ILMTitan:
Further refactor the Publish dialogs before adding to them.


GoogleCloudExtension/GoogleCloudExtension.Utils/Async/AsyncProperty.cs
```diff
@@ -22,6 +22,11 @@ namespace GoogleCloudExtension.Utils.Async
     public class AsyncProperty : AsyncPropertyBase<Task>
     {
         public AsyncProperty(Task task) : base(task) { }
+
+        /// <summary>
+        /// Creates an <see cref="AsyncProperty"/> with an already completed task.
+        /// </summary>
+        public AsyncProperty() : base(Task.CompletedTask) { }
```
redborian: I feel that the default should be a null task. If we want a completed task, we can just pass it in.

ILMTitan: It is a reasonable thought, but I disagree. Passing null to the base creates an `AsyncProperty` in a rather unnatural state, one where `IsPending` and `IsCompleted` are both `false`. All (1) uses of this constructor need an `AsyncProperty` where `IsCompleted` is `true`, and I would expect that to continue going forward.

redborian: I was looking at the usages of "new AsyncProperty(null)" and found that it is used in PublishDialogStepBase. It felt weird to manually initialize an AsyncProperty with null. Just a minor issue. Sgtm 👍

PR #979, opened 2018-06-07
Don't crash for user file corruption.

ILMTitan:
These changes keep the User Account dialog window from crashing when there are invalid credentials files stored on the computer, and enable recovery from a state where there is an invalid file for a user.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/ManageAccounts/UserAccountViewModel.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/ManageAccounts/UserAccountViewModel.cs
@@ -30,11 +30,11 @@ namespace GoogleCloudExtension.ManageAccounts
 
         public string AccountName { get; }
 
-        public UserAccount UserAccount { get; }
+        public IUserAccount UserAccount { get; }
 
         public bool IsCurrentAccount => CredentialsStore.Default.CurrentAccount?.AccountName == UserAccount.AccountName;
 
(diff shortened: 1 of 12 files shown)


GoogleCloudExtension/GoogleCloudExtension/ManageAccounts/UserAccountViewModel.cs
```diff
@@ -27,18 +30,28 @@ public class UserAccountViewModel : Model
...
+            Task<Person> personTask;
+            try
+            {
+                var dataSourceFactory = GoogleCloudExtensionPackage.Instance.GetService<IDataSourceFactory>();
+                IGPlusDataSource dataSource = dataSourceFactory.CreatePlusDataSource(userAccount.GetGoogleCredential());
+                personTask = dataSource.GetProfileAsync();
+            }
+            catch (Exception)
+            {
+                personTask = Task.FromResult<Person>(null);
```
redborian: Should we be logging this somewhere?

ILMTitan: It feels like there should be, but I don't really know of one. There is [`IVsActivityLog`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsactivitylog?view=visualstudiosdk-2017), but that is supposed to be used for "big" events. This too small for it.

PR #1006, opened 2018-06-26
Add link to "Manage Firewall" dialog.

### LINKED ISSUE #183: Add link to "Manage Firewall" dialog to point to the firewall settings for the project.
Add link in the dialog to point to the firewall settings for the project.

ILMTitan:
Adds a link the the firewall rules page of the cloud console on the "Manage Firewall" dialog.
Add a bunch of unit tests.
Fixes #183


GoogleCloudExtension/GoogleCloudExtension/Services/BrowserService.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.ComponentModel.Composition;
+using System.Diagnostics;
+
+namespace GoogleCloudExtension.Services
+{
+    /// <summary>
+    /// Service for opening a browser window.
+    /// </summary>
+    [Export(typeof(IBrowserService))]
+    public class BrowserService : IBrowserService
```
redborian: There should be other places where we would open a browser. Can you update them too?

GoogleCloudExtension/GoogleCloudExtension/FirewallManagement/PortManagerViewModel.cs
```diff
@@ -81,45 +100,38 @@ public class PortManagerViewModel : ViewModelBase
...
-                if (entry.IsEnabled)
-                {
-                    portsToEnable.Add(firewallPort);
-                }
-                else
-                {
-                    portsToDisable.Add(firewallPort);
-                }
-            }
+            string url = string.Format(ConsoleFirewallsUrlFormat, _credentialsStore.Value.CurrentProjectId);
```
redborian: URL encoding the get parameters?

GoogleCloudExtension/GoogleCloudExtension/Services/BrowserService.cs
ILMTitan: I would rather not do that in this PR.

GoogleCloudExtension/GoogleCloudExtension/FirewallManagement/PortManagerViewModel.cs
ILMTitan: Because the parameter is a project id, that should not be necessary. Project IDs are limited to URL safe characters.

Issue #932, opened 2018-04-12
Failed to establish Remote PowerShell connection


iantalarico:
I am unable to attach a debugger.

Using:
- VS Professional 2017 v15.6.6
- Google Cloud Tools for VS: 1.2.100

Steps:
- Create a GCP windows VM instance
- Open VS and then the GCP extension.
- Find your project and windows VM instance
- Try to "Attach debugger"

This will fail with "Failed to establish Remote PowerShell connection"

ILMTitan:
I need to double check from home, but this looks like an issue with Google Corporate network policies disallowing the creation of PSSessions. I'm updating the error messages to make this more clear.

iantalarico:
@ILMTitan Now that you mention this I think I remember this being an issue.

ILMTitan:
Confirmed that my home computer was able to attach the debugger (after a fixing #1011). Closing.

iantalarico:
Should we make some note, reminder or comment somewhere so if I forget about this again we know why?

ILMTitan:
We should be able to search for this issue, even after it is closed.

PR #1009, opened 2018-06-26
Better error on powershell failure.

ILMTitan:
Add better messaging for an error in AttachDebuggerStep InstallStartRemoteTool.
Display information on the first error, rather than continuing with a null `$session`.
Show the line of powershell that failed as well as the error record description.

This was brought to light by issue #932.


GoogleCloudExtension/GoogleCloudExtension.PowerShellUtils/RemoteToolInstaller.cs
```diff
@@ -75,29 +74,21 @@ public class RemoteToolInstaller
...
-            var setupScript = RemotePowerShellUtils.GetEmbeddedFile(InstallerPsFilePath);
+            string setupScript = RemotePowerShellUtils.GetEmbeddedFile(InstallerPsFilePath);
 
             powerShell.AddVariable(CredentialVariablename, _credential);
             powerShell.AddVariable(RemoteToolSourcePath, _debuggerToolLocalPath);
-            powerShell.AddScript(@"$sessionOptions = New-PSSessionOption –SkipCACheck –SkipCNCheck –SkipRevocationCheck");
-            powerShell.AddScript($@"$session = New-PSSession {_computerName} -UseSSL -Credential $credential -SessionOption $sessionOptions");
+            powerShell.AddScript("$ErrorActionPreference = \"Stop\"");
+            powerShell.AddScript("$sessionOptions = New-PSSessionOption –SkipCACheck –SkipCNCheck –SkipRevocationCheck");
+            powerShell.AddScript($"$session = New-PSSession {_computerName} -UseSSL -Credential $credential -SessionOption $sessionOptions");
```
quoctruong: Should we add `ErrorAction Stop` to the `New-PSSession` cmdlet instead of setting the PowerShell default behavior?

ILMTitan: I considered that, but I think setting the default behavior is better. It ensures we see any error, not just the ones we expect, and also keeps the New-PSSession command from being extremely long.

PR #1015, opened 2018-06-28
Faster appveyor rmdir

ILMTitan:
Speed up appveyor by moving the `rmdir c:\QT` into a background powershell job.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,13 @@ image: Visual Studio 2017
 
 configuration: Release
 
+
+init:
+  # This change was recommended by the AppVeyor team to save space.
+  - rmdir C:\cygwin /s /q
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -8,6 +8,13 @@ image: Visual Studio 2017
 
 configuration: Release
 
+
+init:
+  # This change was recommended by the AppVeyor team to save space.
+  - rmdir C:\cygwin /s /q
+  # This was taking up to nine minutes. Make it a job to run it behind the scenes and let the build start immediatly.
+  - ps: Start-Job {Remove-Item "C:\QT" -Recurse -Force} | Out-Null
```
quoctruong: Will the current build interfere with this folder?

ILMTitan: Prior to this PR, the folder was deleted beforehand, so I don't see how it could interfere with the build.

Issue #429, opened 2017-02-15
Add checkbox to GKE publish dialog to update deployment with the same name if exists.


ivannaranjo:
Having a checkbox in the GKE publish dialog will allow the user to pre-choose whether to update the existing deployment with the same name when starting the publish process. The current process can get confusing because it takes a while to detect whether an existing deployment with that name already exists. If the user pre-approved this then we can go ahead with the publishing process.

If the user doesn't pre approve then we need to ask the user anyway, or fail if an existing deployment with that name exists.

PR #1017, opened 2018-07-02
Refactor KubectlContext to inherit from GCloudContext.

ILMTitan:
Move functions from KubectlWrapper and GcloudWrapper to their context.
Add unit tests.
Changes in preparation for fix for #429.


GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudValidationResult.cs
```diff
@@ -39,23 +39,37 @@ public class GCloudValidationResult
...
         {
             IsCloudSdkInstalled = isCloudSdkInstalled;
             IsCloudSdkUpdated = isCloudSdkUpdated;
             IsRequiredComponentInstalled = isRequiredComponentInstalled;
-            CloudSdkVersion = cloudSdkVersion;
         }
+
+        public static GCloudValidationResult NotInstalled { get; } = new GCloudValidationResult(false, false, false);
+
+        public static GCloudValidationResult GetNotUpdated(Version version) =>
```
quoctruong: The naming between `GetNotUpdated` and `Valid` properties can be a bit confusing because both of them are valid but `Valid` is up-to-date whereas the other one is not. Perhaps you can name one as `Updated` and one as `Outdated`?

GoogleCloudExtension/GoogleCloudExtension/GCloud/KubectlContext.cs
```diff
@@ -0,0 +1,251 @@
...
+        /// </summary>
+        /// <param name="deployment">The deployment for which to create and expose the service.</param>
+        /// <param name="makePublic">True if the service should be made public, false otherwise.</param>
+        /// <param name="outputAction">The output callback to be called with output from the command.</param>
+        /// <returns>True if the operation succeeded false otherwise.</returns>
+        public Task<bool> ExposeServiceAsync(string deployment, bool makePublic, Action<string> outputAction)
+        {
+
+            string type = makePublic ? "--type=LoadBalancer" : "--type=ClusterIP";
+            string command = $"expose deployment {deployment} --port=80 --target-port=8080 {type}";
```
quoctruong: This comment is not related to this PR but should we give the users the ability to select the ports besides 8080 and 80 too?

GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudValidationResult.cs
ILMTitan: `GetNotUpdated` is not a valid state. `IsValid` depends on `IsCloudSdkUpdated`, which is false for `GetNotUpdated`. Perhaps I should rename that property to `MinimumVersionMet`? I'm not happy with that name either.

GoogleCloudExtension/GoogleCloudExtension/GCloud/KubectlContext.cs
ILMTitan: Probably, but I am not yet confident enough to know when users would want to do that.

GoogleCloudExtension/GoogleCloudExtension.GCloud/GCloudValidationResult.cs
quoctruong: How about `Valid` and `Obsolete`?

ILMTitan: Much better. Done.

ILMTitan:
@quoctruong PTAL

PR #1022, opened 2018-07-09
Make Deployment Name on the Publish Gke dialog a editable combo box.

### LINKED ISSUE #429: Add checkbox to GKE publish dialog to update deployment with the same name if exists.

ILMTitan:
Add a text block to indicate if we are updating or creating a GKE deployment.
Change the GkeDeployment static class to GkeDeploymentService.
Move all deployment logic to GkeDeploymentService.
During Gke Publish, skip the existing deployment check (already done as part of the UI).

Fixes #429


ILMTitan:
If required, I can split this into two PRs. One commits up to 50ebec1, and one for commits 15a60a1 on.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+                using (StatusbarService.Freeze())
+                using (StatusbarService.ShowDeployAnimation())
+                using (ProgressBarHelper progress =
+                    StatusbarService.ShowProgressBar(Resources.GkePublishDeploymentStatusMessage))
+                using (ShellUtils.SetShellUIBusy())
+                {
+                    DateTime deploymentStartTime = DateTime.Now;
+                    result = new Result();
+
+                    string imageTag = await BuildImageAsync(project, options, progress, result);
```
quoctruong: Perhaps the `BuildImageAsync` should return a `string` and `PublishImageToGkeAsync` should return a `Result` object.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+
+            // Expose the service if requested and it is not already exposed.
+            if (options.ExposeService)
+            {
+                await UpdateOrExposeServiceAsync(options, result);
+            }
+            else
+            {
+                IList<GkeService> services = await options.KubectlContext.GetServicesAsync();
+                GkeService service = services?.FirstOrDefault(x => x.Metadata.Name == options.DeploymentName);
```
quoctruong: Should this be `x?.Metadata?.Name == options.DeploymentName`?

GoogleCloudExtension/GoogleCloudExtension/PublishDialog/Steps/Gke/GkeStepViewModel.cs
```diff
@@ -206,28 +239,49 @@ public GkeStepViewModel(IPublishDialog publishDialog) : base(publishDialog)
...
+            IKubectlContext context = await _kubectlContext;
+            return await context.GetDeploymentsAsync();
+        }
+
+        private void UpdateSelectedDeployment()
+        {
+            if (SelectedDeployment?.Metadata.Name != DeploymentName)
+            {
+                SelectedDeployment = ExistingDeployments?.Value?.FirstOrDefault(
+                    d => d?.Metadata?.Name != null && d.Metadata.Name == DeploymentName);
```
quoctruong: Can this just be `d => d?.Metadata?.Name == DeploymentName`?

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+                result.Failed = true;
+                return;
+            }
+
+            result.ClusterServiceIpAddress =
+                await options.KubectlContext.GetServiceClusterIpAsync(options.DeploymentName);
+
+            if (options.ExposePublicService)
+            {
+                result.PublicServiceIpAddress = await WaitForServicePublicIpAddressAsync(options);
```
quoctruong: Is there a chance that this function fails (when it returns null)? If so, should we check for that?

GoogleCloudExtension/GoogleCloudExtension/Deployment/AppEngineFlexDeployment.cs
```diff
@@ -191,7 +191,7 @@ public DeploymentOptions(string service, string version, bool promote, bool open
             {
                 // Wait for the bundle creation operation to finish, updating progress as it goes.
                 Task<bool> createAppBundleTask = NetCoreAppUtils.CreateAppBundleAsync(project, stageDirectory, toolsPathProvider, outputAction, options.Configuration);
-                if (!await ProgressHelper.UpdateProgress(createAppBundleTask, progress, from: 0.1, to: 0.3))
+                if (!await progress.UpdateProgress(createAppBundleTask, from: 0.1, to: 0.3))
```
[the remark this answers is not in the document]
ILMTitan: Yes, we pass an outputAction to `CreateAppBundleAsync` that prints the console output of the build command to the Google Cloud Tools pane of the output window.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
ILMTitan: I don't think that is necessary. Services shouldn't contain null entries, and Metadata and Name are required fields.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+        private void OutputResultData(
+            IParsedProject project,
+            Options options,
+            Result result)
+        {
+            if (result.ServiceExposed)
+            {
+                if (result.PublicServiceIpAddress != null)
+                {
+                    GcpOutputWindow.OutputLine(
```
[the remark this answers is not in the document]
ILMTitan: I prefer to avoid using that type of convenience property from within an MEF service, preferring instead to (lazily) import dependencies with an `[ImportingConstructor]`. In this case, `GcpOutputWindow` is not the class, but a property for the lazily imported `IGcpOutputWindow` service.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
ILMTitan: We check `result.PublicServiceIpAddress` for null in `OutputResultData(...)` and output a special timeout message if it should have succeeded (when `options.ExposePublicService` is true).

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+            _statusbarService = statusbarService;
+            _browserService = browserService;
+            _toolsPathProvider = VsVersionUtils.ToolsPathProvider;
+        }
+
+
+        /// <summary>
+        /// The options that define an app's deployment. All options are required.
+        /// </summary>
+        public class Options
```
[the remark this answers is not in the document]
ILMTitan: Because it is a nested class, I don't think it needs to be. It will be called `Options` only in `GkeDeploymentService`. Everywhere else it will be called `GkeDeploymentService.Options`.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+            /// <summary>
+            /// The number of replicas to create during the deployment.
+            /// </summary>
+            public int Replicas { get; }
+        }
+
+        /// <summary>
+        /// This class contains the result of a GKE deployment.
+        /// </summary>
+        private class Result
```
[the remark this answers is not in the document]
ILMTitan: Same logic as above. If it needs to be more specific, it would have to be `GkeDeploymentResult`, as there are several unrelated deployments in the project.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+            /// Whether to expose a Kubernetes service based on the deployment created. This will be an HTTP service.
+            /// </summary>
+            public bool ExposeService { get; }
+
+            /// <summary>
+            /// Whether the service to be exposed should be public or not.
+            /// </summary>
+            public bool ExposePublicService { get; }
+
+            public bool OpenWebsite { get; }
```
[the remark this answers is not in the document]
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -0,0 +1,478 @@
...
+                    bool serviceDeleted = await options.KubectlContext.DeleteServiceAsync(
+                        options.DeploymentName,
+                        GcpOutputWindow.OutputLine);
+                    result.Failed |= !serviceDeleted;
+                }
+            }
+        }
+
+
+        private async Task<string> BuildImageAsync(
```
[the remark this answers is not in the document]
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/PublishDialog/Steps/Gke/GkeStepViewModel.cs
ILMTitan: Done.

GoogleCloudExtension/GoogleCloudExtension/PublishDialog/Steps/Gke/GkeStepViewModel.cs
```diff
@@ -105,9 +108,37 @@ public Cluster SelectedCluster
...
+        /// <summary>
+        /// The list of deployments that already exist on the cluster.
+        /// </summary>
+        public AsyncProperty<IList<GkeDeployment>> ExistingDeployments
+        {
+            get => _existingDeployments;
+            set
+            {
+                SetValueAndRaise(ref _existingDeployments, value);
+                ExistingDeployments.PropertyChanged += (sender, args) => UpdateSelectedDeployment();
```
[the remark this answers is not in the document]
ILMTitan: I don't see how we care. It won't cause a memory leak, and extra calls to the method won't be an issue; the method only updates everything to the correct state.

ILMTitan:
PTAL

GoogleCloudExtension/GoogleCloudExtension/PublishDialog/Steps/Gke/GkeStepViewModel.cs
```diff
@@ -345,165 +398,27 @@ protected override void SaveProjectProperties()
...
-                    else
-                    {
-                        GcpOutputWindow.Default.OutputLine(
-                            string.Format(
-                                Resources.GkePublishServiceClusterIpMessage, options.DeploymentName,
-                                result.ClusterServiceIpAddress));
-                    }
-                }
-            }
+            PublishDialog.TrackTask(_kubectlContext);
```
quoctruong: Is there any chance that the `_kubectlContext` is not set before `PublishAsync` is called?

ILMTitan: No. `PublishAsync` is only called by `PublishCommandAsync`. `PublishCommandAsync.CanExecuteCommand` can only be true if `SelectedCluster` is set, which is also when `_kubectlContext` is set.

PR #1026, opened 2018-07-11
Gke Deployment Unit Tests

ILMTitan:
Add unit tests for GkeDeploymentService and NetCoreAppUtils.


GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -145,25 +144,45 @@ private class Result
...
 
             /// <summary>
             /// The IP address within the cluster for the service. This property can only be null if there
             /// was an error deploying the app.
             /// </summary>
-            public string ClusterServiceIpAddress { get; set; }
+            public string ServiceClusterIpAddress { get; }
 
             /// <summary>
             /// Is true if the a service was exposed publicly.
```
quoctruong: This comment is a bit misleading. `ServiceExposed` can be true for `ClusterIP` case (in `GetClusterServiceResult`) but I don't think the service is exposed publicly in this case.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -237,62 +259,58 @@ private class Result
...
+            progress.Report(0.8);
+            if (!success)
             {
-                return result;
+                return Result.FailedResult;
             }
 
-            // Expose the service if requested and it is not already exposed.
-            if (options.ExposeService)
+            GkeService existingService = await options.KubectlContext.GetServiceAsync(options.DeploymentName);
```
quoctruong: I think an alternative way of writing this would be for `DeletingServiceAsync`  to return `Result.SuccessResult` if either `existingService` is null or `options.KubectlContext.DeleteServiceAsync` returns `true`, and to return `Result.FailedResult` otherwise. This function then becomes:

```
if (!options.ExposeService) {
  return await DeleteExistingServiceAsync(existingService, options);
}

if (existingService == null) {
  return await ExposeNewServiceAsync(options.DeploymentName, options);
}

return await UpdateExistingServiceAsync(existingService, options);
```

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
```diff
@@ -317,68 +335,64 @@ private class Result
...
         }
 
-        private async Task UpdateOrExposeServiceAsync(
-            Options options,
-            Result result)
+        private async Task<Result> UpdateExistingServiceAsync(GkeService existingService, Options options)
         {
-            IList<GkeService> services = await options.KubectlContext.GetServicesAsync();
-            GkeService service = services?.FirstOrDefault(x => x.Metadata.Name == options.DeploymentName);
             string requestedType = options.ExposePublicService ?
```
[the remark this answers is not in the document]
ILMTitan: If we enable configuration from a file, we will support them. For now, we won't support them to keep the UI simple.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
ILMTitan: Fixed.

GoogleCloudExtension/GoogleCloudExtension/Deployment/GkeDeploymentService.cs
ILMTitan: I do like that better.

ILMTitan:
PTAL

PR #1029, opened 2018-07-17
Fix flakey PowerShell test.

### LINKED ISSUE #1025: Flakey PowerShell test.
namespace `GoogleCloudExtension.PowerShellUtils.Tests`
class `RemotePowerShellUtilsTests`
method `TestInvokeAsync_ThrowsTaskCanceledExceptionWhenTokenCanceled`

This test method is flakey. There is probably still a race condition in the tested method.

ILMTitan:
Don't register `powershell.Stop` with the CancellationToken until powershell is actually running.

Fixes #1025

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension.PowerShellUtils.Tests/RemotePowerShellUtilsTests.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension.PowerShellUtils.Tests/RemotePowerShellUtilsTests.cs
@@ -33,8 +33,6 @@ namespace GoogleCloudExtension.PowerShellUtils.Tests
         private const string DefaultUser = "DefaultUser";
         private const string DefaultPassword = "DefaultPassword";
 
-        public TestContext TestContext { get; set; }
-
         [TestMethod]
         public void TestGetEmbeddedFile_ThrowsForUnknownFile()
         {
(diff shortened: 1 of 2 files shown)


GoogleCloudExtension/GoogleCloudExtension.PowerShellUtils.Tests/RemotePowerShellUtilsTests.cs
```diff
@@ -94,6 +92,8 @@ public void TestCreatePSCredential_CreatesCredentialWithGivenPassword()
         {
             using (PowerShell powerShell = PowerShell.Create())
             {
+                powerShell.AddScript("Start-Sleep -Seconds 30");
```
iantalarico: Why do we need to add the sleep here if we are waiting to register the BeginStop below?

ILMTitan: I was getting false positives. Powershell can not start if there are no commands. This test (TestInvokeAsync_DoesNotStartPowershellWhenTokenAlreadyCanceled) is supposed to test that the first line, `cancelToken.ThrowIfCancelRequested()`, works. Without the script, the test was succeeding even when I commented that line out.

iantalarico: sounds reasonable.  Maybe make this comment in the file for the next person that comes along?

ILMTitan: Good point. Done.

PR #1030, opened 2018-07-18
Downgrade NewtonSoft.Json to 10.0.2

ILMTitan:


GoogleCloudExtension/GoogleCloudExtension/packages.config
```diff
@@ -44,7 +44,7 @@
   <package id="Microsoft.VisualStudio.Utilities" version="15.7.27703" targetFramework="net46" />
   <package id="Microsoft.VisualStudio.Validation" version="15.3.53" targetFramework="net46" />
   <package id="Microsoft.VSSDK.BuildTools" version="15.7.109" targetFramework="net452" developmentDependency="true" />
-  <package id="Newtonsoft.Json" version="11.0.2" targetFramework="net452" />
+  <package id="Newtonsoft.Json" version="10.0.2" targetFramework="net46" />
```
iantalarico: why the change from 452 to 46?

GoogleCloudExtension/GoogleCloudExtension.Utils/packages.config
```diff
@@ -1,4 +1,20 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
+<!--
+    Copyright 2016 Google Inc. All Rights Reserved.
```
iantalarico: nit: all new copyrights are 2016 should be 2018 and for all above

GoogleCloudExtension/GoogleCloudExtension/packages.config
ILMTitan: VS 2017 runs on .NET Framework 4.6, so we are now targeting that framework with the projects.

GoogleCloudExtension/GoogleCloudExtension.Utils/packages.config
ILMTitan: 2016 is the year they were created.

Issue #1013, opened 2018-06-28
Stop using Bower in the templates.


ILMTitan:
When installing bower, this message comes up:

> npm WARN deprecated bower@1.8.4: We don't recommend using Bower for new projects. Please consider Yarn and Webpack or Parcel. You can read how to migrate legacy project here: https://bower.io/blog/2017/how-to-migrate-away-from-bower/

We change to a different package manager to manage our javascript packages in our MVC templates.

PR #1033, opened 2018-07-19
Update templates to use npm rather than bower.

ILMTitan:


ILMTitan:
Visual Studio has npm installed as an external tool, but does not put it on the path. I added an MSBuild warning for when npm is not found. Users can then either install npm, or do a manual package update from within Visual Studio (which will cause MSBuild to skip the NpmInstall target).

ILMTitan:
Fixes #1013

PR #1045, opened 2018-07-31
Fix deploy to GCE

ILMTitan:
* Remove app_offline.htm. It was not playing well with IIS Express (not sure why).
* Add `<PublishWithAspNetCoreTargetManifest>false</...>` to 2.x Project files. Fixes [this deployment error](https://github.com/dotnet/coreclr/issues/13542).
* Fix showing the output for the GCE deploy. Running the build was hiding it again. Move the show after the build.
* Retry a failed deploy build once. The deploy will sometimes not be able to delete a dll the first time, especially when switching deployed projects.


GoogleCloudExtension/GoogleCloudExtension/Deployment/WindowsVmDeployment.cs
```diff
@@ -102,10 +107,20 @@ public class WindowsVmDeployment : IWindowsVmDeployment
...
-                    string.Join(" ", parameters),
+                    msBuildParameters,
                     GcpOutputWindow.OutputLine);
+
+                if (result)
+                {
+                    return true;
+                }
+
+                // An inital failure is common, retry.
```
iantalarico: Should we surface this to the user?  Any reason not to retry more than once?

GoogleCloudExtension/GoogleCloudExtension/Deployment/WindowsVmDeployment.cs
```diff
@@ -102,10 +107,20 @@ public class WindowsVmDeployment : IWindowsVmDeployment
...
             {
-                return await ProcessService.RunCommandAsync(
+                string msBuildParameters = string.Join(" ", parameters);
+                bool result = await ProcessService.RunCommandAsync(
                     msbuildPath,
-                    string.Join(" ", parameters),
+                    msBuildParameters,
                     GcpOutputWindow.OutputLine);
+
+                if (result)
```
iantalarico: Do we need to look more closely at this type of error and why it failed before we retry?

GoogleCloudExtension/GoogleCloudExtension/Deployment/WindowsVmDeployment.cs
ILMTitan: The user will be able to see the build output occurring twice, including the initial errors. It seems to be a timing issue with shutting the application down. Trying again seems to work.

GoogleCloudExtension/GoogleCloudExtension/Deployment/WindowsVmDeployment.cs
ILMTitan: 1. I can't think of an error that would be made worse by retrying the deploy
2. Figuring out the error type would involve parsing the stdout, which I don't want to do.

ILMTitan:
PTAL

PR #1049, opened 2018-08-02
Update options to allow hiding the gcp menu bar control.

ILMTitan:
Add an option to the settings dialog page to hide the GCP Menu Bar control.
Change the display name of the setting to "General".
Enable i8n of the settings pages names.


GoogleCloudExtension/GoogleCloudExtension/Analytics/AnalyticsOptInDialog/AnalyticsOptInWindowViewModel.cs
```diff
@@ -12,16 +12,18 @@
...
 using GoogleCloudExtension.Utils;
+using System;
 
 namespace GoogleCloudExtension.Analytics.AnalyticsOptInDialog
 {
     /// <summary>
     /// View model for user control AnalyticsOptInWindowContent.xaml.
     /// </summary>
-    public class AnalyticsOptInWindowViewModel : ViewModelBase
+    public class AnalyticsOptInWindowViewModel : ViewModelBase, IViewModelBase<bool>
```
quoctruong: Doesn't `ViewModelBase` also implements `IViewModelBase`? Can we have a generic `ViewModelBase<T>` that implements `IViewModelBase<T>` so this can just be `AnalyticsOptInWindowViewModel : ViewModelBase<bool>`?

GoogleCloudExtension/GoogleCloudExtension/Options/AnalyticsOptions.cs
```diff
@@ -22,23 +22,37 @@
 namespace GoogleCloudExtension.Options
 {
     /// <summary>
-    /// This class represents the extension's analytics settings.
+    /// This class represents the extension's general options. It can not be renamed to keep backwards compatiblity.
     /// </summary>
     [DesignerCategory("Code")]
-    public class AnalyticsOptions : UIElementDialogPage
+    public class AnalyticsOptions : UIElementDialogPage, INotifyPropertyChanged
```
quoctruong: Should we change this class name now that it represents general options instead of just analytics?

ILMTitan: Unfortunately, we can't. The path to the options in the registry includes the name of the class storing those options, so changing the class name would reset everyone's options when they upgrade.

GoogleCloudExtension/GoogleCloudExtension/Analytics/AnalyticsOptInDialog/AnalyticsOptInWindowViewModel.cs
ILMTitan: Done.

PR #1076, opened 2018-12-07
Do not crash when there are multiple credentials files for a single user account.

### LINKED ISSUE #1065: Duplicate account files should not crash the extension.
See #1064.

Multiple user account json files with the same account name should not prevent the extension from launching. We should instead choose one arbitrarily, and delete the others.

ILMTitan:
Fixes #1065

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/Accounts/CredentialsStore.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/Accounts/CredentialsStore.cs
@@ -178,14 +178,13 @@ namespace GoogleCloudExtension.Accounts
         /// Deletes the <paramref name="account"/> from the store. The account must exist in the store
         /// or it will throw.
         /// </summary>
-        /// <param name="account">The accound to delete.</param>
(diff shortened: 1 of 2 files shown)


GoogleCloudExtension/GoogleCloudExtension/Accounts/CredentialsStore.cs
```diff
@@ -247,10 +246,19 @@ private string GetUserAccountPath(string accountName)
...
+                .Select(
+                    path => new StoredUserAccount
+                    {
+                        FileName = Path.GetFileName(path),
+                        UserAccount = LoadUserAccount(path)
+                    })
+                .GroupBy(a => a.UserAccount.AccountName)
+                .ToDictionary(
+                    g => g.Key,
+                    g => g.OrderBy(account => account.FileName, StringComparer.InvariantCulture).First());
```
quoctruong: nit: I think using an `Aggregate` or `Max` may be more efficient than sorting the list and get the first item.

PR #1077, opened 2018-12-07
Ignore two tests that are flaky on AppVeyor.

ILMTitan:
TestInvokeAsync_ThrowsTaskCanceledExceptionWhenTokenCanceled
and 
TestInvokeAsync_StopsPowerShellWhenTokenCanceled
are flakey when run on AppVeyor (but not locally).
Skip them on AppVeyor.


iantalarico:
Before we skip them do we know why they are flaky?  Is it a resource issue?

ILMTitan:
I do not know why they are flaky. I have spent several afternoons trying to unflake them, and even made some PRs that I thought would unflake them, but never to any success.

It seems like, for some reason, on AppVeyor, PowerShell sometimes does not stop after `PowerShell.Stop()` is called. Instead, it runs to completion (which times out the tests).

PR #1097, opened 2019-10-07
Fix crash when Visual Studio Team is opened

quoctruong:
We are trying to access an Instance of the GoogleCloudExtensionPackage when it is not yet initialized.

### DIFF
--- a/GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
+++ b/GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
@@ -144,11 +144,14 @@ namespace GoogleCloudExtension.CloudSourceRepositories
 
         async Task ISectionViewModel.InitializeAsync(ITeamExplorerUtils teamExplorerService)
         {
-            int retryAttempts = 30;
+            // When the user switches to TeamExplorer immediately after opening Visual Studio,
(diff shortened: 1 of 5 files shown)


GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -142,8 +142,21 @@ public void Refresh()
             }
         }
 
-        void ISectionViewModel.Initialize(ITeamExplorerUtils teamExplorerService)
+        async Task ISectionViewModel.InitializeAsync(ITeamExplorerUtils teamExplorerService)
         {
+            int retryAttempts = 30;
```
meteatamel: With 2000ms delay time, I think, we shouldn't have retryAttempts more than 3 or 5 times max? Can we also add a comment to explain why we do the waiting here.

GoogleCloudExtension/GoogleCloudExtension/CloudSourceRepositories/CsrSectionControlViewModel.cs
```diff
@@ -142,8 +142,21 @@ public void Refresh()
...
+            while (retryAttempts > 0 && GoogleCloudExtensionPackage.Instance == null)
+            {
+                await Task.Delay(2000);
+                retryAttempts -= 0;
```
meteatamel: This doesn't decrement retryAttempt, no?

